PR #234, opened 2026-01-29
feat(fuse): add image support for FUSE filesystem

aaronsb:
## Summary

- Route image files copied into FUSE to the dedicated `/ingest/image` API (ADR-057) instead of the generic `/ingest` path, preserving visual embeddings, Garage S3 storage, and proper `content_type` metadata
- Display image documents as dual filesystem entries: raw image bytes (`diagram.png`) + companion markdown (`diagram.png.md`) with YAML frontmatter, relative image link, and vision AI prose description
- Add `images/` subdirectory to query results for lazy-loaded image evidence from concept instances, with source deduplication and filename collision handling
- Add integration tests (tests 8-11) covering image ingestion, dual-file completion, binary content verification, and job cleanup

## Files changed

| File | Changes |
|------|---------|
| `fuse/kg_fuse/api_client.py` | `get_bytes()` for raw binary API responses |
| `fuse/kg_fuse/models.py` | `content_type` field, `images_dir` in DIR_TYPES, new entry types |
| `fuse/kg_fuse/formatters.py` | `format_image_prose()`, image evidence refs in `format_concept()` |
| `fuse/kg_fuse/filesystem.py` | Image detection, dual-file listing, read handlers, write routing, query images dir, 100MB immutable cache |
| `fuse/tests/test_fuse_integration.sh` | Tests 8-11 for image pipeline, `images/` dir in query test |
| `fuse/tests/fixtures/test_western_town.jpg` | Test image fixture |

## Test plan

- [ ] Run existing unit tests: `cd fuse && .venv/bin/pytest tests/ -v` (53 tests pass)
- [ ] Mount FUSE and run integration tests: `./tests/test_fuse_integration.sh`
- [ ] Manual: `cp image.png ~/Knowledge/ontology/{name}/` and verify dual-file appearance
- [ ] Manual: `file ~/Knowledge/ontology/{name}/documents/image.png` confirms image format
- [ ] Manual: `cat ~/Knowledge/ontology/{name}/documents/image.png.md` shows prose + link

### DIFF
--- a/fuse/kg_fuse/filesystem.py
+++ b/fuse/kg_fuse/filesystem.py
@@ -41,7 +41,8 @@ import pyfuse3
 
 from .api_client import KnowledgeGraphClient
 from .config import TagsConfig, JobsConfig
-from .formatters import format_concept, format_document, format_image_prose, format_job, render_meta_file
+from .formatters import format_concept, format_document, format_job, render_meta_file
+from .image_handler import ImageHandler
 from .job_tracker import JobTracker, TERMINAL_JOB_STATUSES
 from .models import InodeEntry, is_dir_type
 from .query_store import QueryStore
@@ -101,10 +102,15 @@ class KnowledgeGraphFS(pyfuse3.Operations):
         # Job tracking: lazy polling with automatic cleanup
         self._job_tracker = JobTracker()
 
-        # Image cache: images are immutable (content-addressed), cache aggressively
-        self._image_cache: dict[str, bytes] = {}  # cache_key -> raw image bytes
-        self._image_cache_total = 0  # Total bytes in cache
-        self._image_cache_max = 100 * 1024 * 1024  # 100MB budget
+        # Image handler: reads, caches, ingests, and manages image inodes
+        self._image_handler = ImageHandler(
+            api=self._api,
+            tags_config=self.tags_config,
+            job_tracker=self._job_tracker,
+            inodes=self._inodes,
+            allocate_inode=self._allocate_inode,
+            sanitize_filename=self._sanitize_filename,
+        )
 
     def _make_attr(self, inode: int, is_dir: bool = False, size: int = 0, writable: bool = False) -> pyfuse3.EntryAttributes:
         """Create file attributes."""
@@ -260,7 +266,7 @@ class KnowledgeGraphFS(pyfuse3.Operations):
 
             # Check for images directory
             if name_str == "images":
-                inode = self._get_or_create_images_dir_inode(ontology, parent_path, parent_inode)
+                inode = self._image_handler.get_or_create_images_dir_inode(ontology, parent_path, parent_inode)
                 return await self.getattr(inode, ctx)
 
             nested_path = f"{parent_path}/{name_str}" if parent_path else name_str
@@ -278,7 +284,10 @@ class KnowledgeGraphFS(pyfuse3.Operations):
 
         # images directory: list image evidence files
         elif parent_entry.entry_type == "images_dir":
-            entries = await self._list_query_images(parent_inode, parent_entry.ontology, parent_entry.query_path)
+            entries = await self._image_handler.list_query_images(
+                parent_inode, parent_entry.ontology, parent_entry.query_path,
+                self._dir_cache, self._cache_time, self._cache_ttl
+            )
             for inode, file_name in entries:
                 if file_name == name_str:
                     return await self.getattr(inode, ctx)
@@ -461,7 +470,10 @@ class KnowledgeGraphFS(pyfuse3.Operations):
 
         elif entry.entry_type == "images_dir":
             # images/: list image evidence from query concepts
-            entries = await self._list_query_images(fh, entry.ontology, entry.query_path)
+            entries = await self._image_handler.list_query_images(
+                fh, entry.ontology, entry.query_path,
+                self._dir_cache, self._cache_time, self._cache_ttl
+            )
 
         elif entry.entry_type == "meta_dir":
             # .meta/: list virtual config files
@@ -579,13 +591,13 @@ class KnowledgeGraphFS(pyfuse3.Operations):
 
                 if content_type == "image":
                     # Image: two entries — raw image bytes + companion .md
-                    img_inode = self._get_or_create_image_document_inode(
+                    img_inode = self._image_handler.get_or_create_image_document_inode(
                         filename, parent_inode, ontology, document_id
                     )
                     entries.append((img_inode, filename))
 
                     prose_name = f"{filename}.md"
-                    prose_inode = self._get_or_create_image_prose_inode(
+                    prose_inode = self._image_handler.get_or_create_image_prose_inode(
                         prose_name, parent_inode, ontology, document_id
                     )
                     entries.append((prose_inode, prose_name))
@@ -674,7 +686,7 @@ class KnowledgeGraphFS(pyfuse3.Operations):
             log.error(f"Failed to execute search for {ontology}/{query_path}: {e}")
 
         # Always include images/ directory (lazy-loaded on readdir)
-        images_dir_inode = self._get_or_create_images_dir_inode(ontology, query_path, parent_inode)
+        images_dir_inode = self._image_handler.get_or_create_images_dir_inode(ontology, query_path, parent_inode)
         entries.append((images_dir_inode, "images"))
 
         # Add child query directories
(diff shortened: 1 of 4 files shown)


aaronsb:
## Code Review: Image Support for FUSE Filesystem

**What this changes**: Adds full image lifecycle to the FUSE virtual filesystem -- ingestion routing to `/ingest/image` (ADR-057), dual-file representation in `documents/` (raw bytes + companion `.md`), query-scoped `images/` directory with lazy-loaded evidence, and an immutable image cache with a 100MB budget. Integration tests cover the end-to-end pipeline across four new test sections.

**PR size**: 609 lines across 6 files, 2 commits. Standard review scope.

---

### Findings

#### 1. Bug: `_read_image_bytes` fetches JSON but needs raw bytes

**Location**: `fuse/kg_fuse/filesystem.py:1106-1125`

**Problem**: `_read_image_bytes` calls `self._api.get()` (which returns parsed JSON via `response.json()`), then extracts a base64-encoded `content["image"]` string and decodes it. Meanwhile, `_read_image_evidence` calls the new `self._api.get_bytes()` to fetch raw bytes directly.

This works if the `/documents/{id}/content` endpoint returns base64-encoded image data inside a JSON envelope, which appears to be the case. However, this means every image document read requires the API to base64-encode the image (33% size inflation) and the FUSE driver to decode it back. That is a correctness-valid path but worth noting as a performance consideration -- the companion `_read_image_evidence` avoids this overhead by using `get_bytes()` against a presumably raw-bytes endpoint.

**Assessment**: Not a bug, but an asymmetry worth documenting. If the API supports a raw-bytes endpoint for document images (similar to `/sources/{id}/image`), switching `_read_image_bytes` to use `get_bytes()` would eliminate the base64 round-trip overhead.

#### 2. Design Smell: `document_id` field overloading

**Location**: `fuse/kg_fuse/models.py:33`, `fuse/kg_fuse/filesystem.py:1141,1489`

**Problem**: The `document_id` field on `InodeEntry` stores a `source_id` for `image_evidence` entries, as documented in the comment `# Store source_id in document_id field`. This is a semantic mismatch -- the field name says "document" but the value is a "source."

**Why it matters**: Future maintainers (or even the same author a few months from now) will need to remember this exception. The code already has comments marking the overload in three places, which is actually a sign that the abstraction wants to be explicit.

**Suggestion**: Adding a `source_id: Optional[str] = None` field to `InodeEntry` would cost one line in the dataclass and eliminate the semantic confusion. The `image_evidence` entries would then use `source_id` natively, and the three explanatory comments could be removed. This is a low-priority improvement -- the current approach works and is documented.

#### 3. Cache never evicts -- silent data loss on budget exhaustion

**Location**: `fuse/kg_fuse/filesystem.py:1154-1158`

**Problem**: `_cache_image` silently drops images that would exceed the 100MB budget:

```python
def _cache_image(self, key: str, data: bytes) -> None:
    if self._image_cache_total + len(data) < self._image_cache_max:
        self._image_cache[key] = data
        self._image_cache_total += len(data)
```

Once the budget fills up, all subsequent images are silently not cached. There is no eviction policy, no logging, and no way to recover cache space without unmounting. In a knowledge graph with many image documents, this could mean the first 100MB of images get cached and everything else gets re-fetched on every access.

**Why it matters**: For content-addressed immutable images (as documented), this is acceptable behavior -- the cache is purely a performance optimization. But the silent drop means operators have no visibility into cache pressure. Two improvements worth considering:

1. Log a debug message when an image is not cached due to budget exhaustion.
2. Consider an LRU eviction policy if image-heavy ontologies become common.

The current approach is reasonable for an initial implementation. The immutability invariant (content-addressed = same bytes forever) means the cache never has staleness bugs, which is a strong design property.

#### 4. `import base64` inside method body

**Location**: `fuse/kg_fuse/filesystem.py:1115`

**Problem**: `base64` is imported inside `_read_image_bytes()` rather than at module level. This is a standard library module, so there is no lazy-loading performance benefit -- it is available immediately.

**Assessment**: This is a very minor style issue. Moving `import base64` to the top-level imports (line 32 area) would be more conventional. The inline import works fine, it just reads as though there was a reason to defer the import when there is none.

#### 5. `image_document` reports `size=0` until first read

**Location**: `fuse/kg_fuse/filesystem.py:891`

**Problem**: Image document inodes are created with `size=0` and the actual size is only set after `_read_image_bytes` completes. This means `stat` / `ls -l` will show a 0-byte file until someone reads it. Some tools (thumbnail generators, file managers) may skip zero-length files or behave unexpectedly.

**Why it matters**: This is a known trade-off in virtual filesystems -- you either pay the cost of pre-fetching size metadata or report an approximation. The existing text `document` entries use `size=4096` as a placeholder, and `image_prose` entries do the same. Using `size=0` for images breaks this convention.

**Suggestion**: Consider using a non-zero placeholder (e.g., `4096` like other entries) or fetching size metadata from the document listing API if it provides a `content_length` field. The inconsistency between `image_document` (size=0) and `image_prose` (size=4096) is the primary concern.

#### 6. `_list_query_images` is O(N) API calls for N concepts

**Location**: `fuse/kg_fuse/filesystem.py:1495-1557`

**Problem**: For each concept in the parent query, `_list_query_images` makes an individual API call to `/query/concept/{id}` to discover image evidence. If a query returns 50 concepts, this is 50 sequential HTTP requests to populate the `images/` directory.

**Why it matters**: This is mitigated by directory caching (30-second TTL), so it only happens on first access and cache expiry. But it could cause noticeable latency on `ls images/` for image-rich queries. The lazy-loading design (images/ is always present, populated on readdir) is correct.

**Assessment**: This is acceptable for the initial implementation. If it becomes a latency problem, a batch endpoint (`POST /query/concepts/images` accepting multiple concept_ids) would be the right fix on the API side. The FUSE code structure already supports plugging in such an endpoint.

#### 7. FUSE dispatch chain is complete and correct

I traced every new entry type through all four dispatch points:

| Entry Type | `lookup` | `readdir` | `open` | `read` |
|---|---|---|---|---|
| `images_dir` | query parent handles it | handled in readdir | N/A (dir) | N/A (dir) |
| `image_document` | documents_dir parent | listed in _list_documents | in open allow-list | returns raw bytes |
| `image_prose` | documents_dir parent | listed in _list_documents | in open allow-list | returns formatted md |
| `image_evidence` | images_dir parent | listed in _list_query_images | in open allow-list | returns raw bytes |

All four new entry types are properly registered in:
- `open()` allow-list (line 1037-1038)
- `read()` dispatch (lines 1062-1071)
- `lookup()` parent handlers (lines 262-264 for images_dir, 280-284 for image_evidence)
- `readdir()` (lines 462-464 for images_dir)

No missing dispatch paths found.

#### 8. Filename collision handling is solid

**Location**: `fuse/kg_fuse/filesystem.py:1539-1545`

The `_list_query_images` method handles filename collisions with a counter suffix (`base-1.ext`, `base-2.ext`, etc.), and deduplicates by `source_id` via the `seen_sources` set. This is correct and handles the edge case where multiple concepts reference the same image (dedup) or different images have the same filename (collision resolution).

#### 9. Integration tests are well-structured but have a gap

**Location**: `fuse/tests/test_fuse_integration.sh:436-643`

The tests cover: image ingestion (test 8), dual-file completion (test 9), content verification with magic bytes and prose structure (test 10), and job cleanup (test 11). The JPEG magic byte check and PNG fallback is a nice touch.

**Gap**: There is no test that reads the raw image bytes from `documents/` and verifies the file size is non-zero and matches the original fixture. The magic bytes check only reads the first 3 bytes. Adding `wc -c` comparison between the original fixture and the FUSE-served file would catch truncation bugs.

**Minor**: Test 10 (lines 557-562) checks for PNG magic bytes but only reads 3 bytes via `xxd -l 3`, while the PNG signature is 8 bytes (89 50 4E 47 0D 0A 1A 0A). The substring check `${IMAGE_MAGIC:0:8}` operates on a 6-character hex string (3 bytes = 6 hex chars), so it can never match "89504e47" (8 hex chars). This branch is unreachable. Since the test fixture is JPEG, this does not affect correctness, but the PNG fallback is dead code.

### What looks good

- **ADR-057 compliance**: Image ingestion correctly routes to `/ingest/image` with the expected form fields (`ontology`, `auto_approve`, `source_type`). This maintains the separation between text and image ingestion pipelines.

- **Dual-file pattern**: The `image.png` + `image.png.md` convention is well thought out. Image viewers get raw bytes; Markdown readers (Obsidian, agents) get prose + metadata. The relative image link in the companion `.md` (``) makes the pair self-referencing.

- **Lazy loading**: The `images/` directory is always present in queries but only populated when accessed. This avoids API calls for users who never look at image evidence.

- **Immutable cache invariant**: Content-addressed images cannot change, so the cache has no invalidation bugs by construction. This is the right choice for this data model.

- **Write path routing**: The `_is_image_file()` check in `release()` cleanly separates the text and image ingestion paths without changing the existing text flow.

- **Job tracking for images**: Both the raw image and companion `.md` get `.ingesting` virtual files, so `ls documents/` shows consistent state during processing.

### filesystem.py at 1705 lines

The file was 1415 lines before this change and is now 1705 lines. That crosses the 500-line threshold significantly. However, looking at the structure, the file is a single FUSE operations class that inherently needs all these methods (lookup, readdir, open, read, write, create, release, etc.) in one place because pyfuse3 dispatches to method overrides on a single class.

A reasonable extraction would be an `ImageHandler` helper class encapsulating `_read_image_bytes`, `_read_image_prose`, `_read_image_evidence`, `_cache_image`, `_list_query_images`, `_ingest_image`, and the three `_get_or_create_image_*_inode` methods. This could be injected into `KnowledgeGraphFS.__init__` and delegated to from the dispatch methods. That would remove approximately 200 lines from `filesystem.py` and create a focused module.

This is not blocking -- the current structure is coherent and well-organized. But as more media types get added (PDFs, audio), the same pattern will repeat, and extracting now would set a precedent.

### Summary

**Assessment**: This is a well-implemented feature that integrates cleanly into the existing FUSE architecture. The dispatch chain is complete, edge cases are handled, and the design decisions (immutable cache, lazy loading, dual-file pattern) are sound. No blocking issues found.

**Non-blocking suggestions** (in priority order):
1. Add a `source_id` field to `InodeEntry` to eliminate the `document_id` overload
2. Use a non-zero placeholder size for `image_document` entries (consistency with other types)
3. Move `import base64` to top-level
4. Add a debug log when image cache budget is exhausted
5. Fix the dead PNG branch in the integration test
6. Consider extracting image-handling methods into a helper class as the file grows

---
*Review by Claude Opus 4.5*

PR #235, opened 2026-01-29
fix: store content_type and storage_key on DocumentMeta nodes

aaronsb:
## Summary

- `create_document_meta()` never stored `content_type` or `storage_key` on DocumentMeta nodes — only Source nodes had them
- This caused `GET /documents` to always return `content_type: "document"`, breaking FUSE image detection
- `GET /documents/{id}/content` couldn't find image binaries without `storage_key` on DocumentMeta

## Changes

- **`api/app/lib/age_client.py`** — Add `content_type` and `storage_key` parameters to `create_document_meta()`
- **`api/app/workers/ingestion_worker.py`** — Pass both from `job_data` at the call site
- **`schema/migrations/043_backfill_document_meta_content_type.sql`** — Backfill existing DocumentMeta nodes from their Source nodes
- **`fuse/kg_fuse/filesystem.py`** — Set `direct_io=True` on image file handles so kernel reads aren't truncated to `st_size`
- **`fuse/tests/test_fuse_integration.sh`** — Fix byte-count test (`cat|wc`), regex warning (`grep -qF`), extend job cleanup retries

## Test plan

- [x] Integration tests: 28/29 pass (1 flaky timing test for job cleanup)
- [x] `kg search query 'road'` returns "Dirt Road" concept with inline image evidence
- [x] FUSE serves full image bytes (1,008,066 bytes matches fixture)
- [x] Image companion `.md` has correct frontmatter and relative link
- [x] Migration backfills `content_type` and `storage_key` on existing DocumentMeta nodes

### DIFF
--- a/api/app/lib/age_client.py
+++ b/api/app/lib/age_client.py
@@ -2421,7 +2421,9 @@ class AGEClient:
         ingested_at: Optional[str] = None,
         source_ids: Optional[List[str]] = None,
         # ADR-081: Source document lifecycle
-        garage_key: Optional[str] = None
+        garage_key: Optional[str] = None,
+        content_type: str = "document",
+        storage_key: Optional[str] = None,
     ) -> Dict[str, Any]:
         """
         Create a DocumentMeta node and link it to Source nodes (ADR-051, ADR-081).
@@ -2443,6 +2445,8 @@ class AGEClient:
             ingested_at: ISO timestamp (optional, defaults to now())
             source_ids: List of source_ids to link via HAS_SOURCE relationship (optional)
             garage_key: Garage object key for source document (ADR-081)
+            content_type: "document" or "image" (default: "document")
+            storage_key: Garage object key for image binary (ADR-057, images only)
 
         Returns:
             Created DocumentMeta node properties
@@ -2478,6 +2482,10 @@ class AGEClient:
             "job_id": job_id
         }
 
+        # Content type (document or image)
+        if content_type != "document":
+            properties["content_type"] = content_type
+
         # Add optional provenance metadata (best-effort)
         if filename:
             properties["filename"] = filename
@@ -2490,6 +2498,9 @@ class AGEClient:
         # ADR-081: Link to source document in Garage
         if garage_key:
             properties["garage_key"] = garage_key
+        # ADR-057: Image binary location in Garage
+        if storage_key:
+            properties["storage_key"] = storage_key
 
         # Add timestamp (default to now if not provided)
         if ingested_at:
--- a/api/app/workers/ingestion_worker.py
+++ b/api/app/workers/ingestion_worker.py
@@ -395,7 +395,10 @@ def run_ingestion_worker(
                 hostname=job_data.get("source_hostname"),      # Hostname where ingested
                 source_ids=source_ids,
                 # ADR-081: Link to source document in Garage
-                garage_key=job_data.get("source_garage_key")
+                garage_key=job_data.get("source_garage_key"),
+                content_type=job_data.get("content_type", "document"),
+                # ADR-057: Image binary location in Garage
+                storage_key=job_data.get("storage_key"),
             )
             logger.info(f"✓ Created DocumentMeta node: {job_data['content_hash'][:16]}... ({stats.sources_created} sources)")
         except Exception as e:
--- a/fuse/kg_fuse/filesystem.py
+++ b/fuse/kg_fuse/filesystem.py
@@ -1024,7 +1024,14 @@ class KnowledgeGraphFS(pyfuse3.Operations):
         if entry.entry_type == "job_file" and (flags & os.O_WRONLY or flags & os.O_RDWR):
             raise pyfuse3.FUSEError(errno.EACCES)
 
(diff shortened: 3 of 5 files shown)


aaronsb:
## Code Review: fix/document-meta-content-type

**PR #235** -- 5 files, +144/-11 lines. Standard-tier review covering correctness, safety, edge cases, consistency, and regressions.

---

### 1. Correctness -- Does the fix address the root cause?

**Yes.** The root cause was clearly a data gap: `create_document_meta()` in `age_client.py` accepted neither `content_type` nor `storage_key` as parameters, so those properties were never written to `DocumentMeta` nodes even though the ingestion worker had them available in `job_data`. Meanwhile, the API routes in `documents.py` read `d.content_type` and `d.storage_key` directly from DocumentMeta nodes (lines 237, 498, 657), defaulting to `'document'` when missing. This made every image document report as `content_type="document"` and have no `storage_key`, breaking both FUSE image detection and `/documents/{id}/content` retrieval.

The fix closes the loop at both ends:
- **Forward path** (`age_client.py` + `ingestion_worker.py`): New DocumentMeta nodes now get `content_type` and `storage_key` properties at creation time.
- **Backward path** (migration 043): Existing DocumentMeta nodes get backfilled from their linked Source nodes.

The FUSE `direct_io` fix in `filesystem.py` is a correctly identified secondary symptom -- without it, the kernel page cache could truncate image reads to the placeholder `st_size` (4096) even after the underlying data became available.

### 2. Safety -- Cypher queries in the migration

**No injection risk.** All three Cypher blocks in migration 043 use hardcoded literal values (`'image'`), not interpolated user input. They are wrapped in `DO $migration$` blocks with `EXECUTE format(...)` using `$cypher$` dollar-quoted strings. No parameters are passed from external sources.

The migration's `WHERE ... IS NULL` guards make it idempotent -- safe to run multiple times. The verification step at the end is read-only.

One observation: the migration does **not** use an explicit `BEGIN`/`COMMIT` transaction wrapper like migration 042 does. Each `DO` block runs in its own implicit transaction, which is fine for correctness (the steps are independent), but it differs from the prior migration pattern. This is a minor style inconsistency, not a functional concern -- AGE Cypher blocks within `DO` blocks cannot be rolled back as a single unit anyway since AGE manages its own transactions internally.

### 3. Edge Cases

**content_type conditional storage** (`age_client.py:2486-2487`):

```python
if content_type != "document":
    properties["content_type"] = content_type
```

This means text documents will have **no** `content_type` property on their DocumentMeta nodes. This is intentional and safe because every API consumer already uses `meta.get('content_type') or 'document'` as a fallback (confirmed at `documents.py` lines 422, 516, 683). The sparse-property approach avoids cluttering the majority of nodes (text documents) with a redundant default, which is consistent with how Apache AGE graph properties typically work.

**storage_key when None** (`age_client.py:2502-2503`):

```python
if storage_key:
    properties["storage_key"] = storage_key
```

The truthiness check correctly skips `None` and empty strings. The downstream code in `documents.py:528` already guards with `if storage_key:` before attempting S3 retrieval. No risk of None-related failures.

**Migration edge case -- images with multiple Source nodes**: The comment on line 55 of the migration says "images have one source per doc," and the `SET d.storage_key = s.storage_key` pattern relies on this. If an image document somehow had multiple Source nodes with different `storage_key` values, AGE would apply the last matched one (non-deterministic). Given the image ingestion pipeline always creates exactly one Source per image, this is acceptable -- but worth noting as an assumption.

### 4. Consistency

The changes follow established patterns well:

- **Parameter naming**: `content_type` and `storage_key` match the exact property names already used on Source nodes and in API response models.
- **Default handling**: `content_type: str = "document"` in the function signature matches the same default in `ingestion.py:176` and the ingestion worker call site (`job_data.get("content_type", "document")`).
- **Migration structure**: Follows the existing pattern of numbered SQL files with header comments, `LOAD 'age'`, search path setup, and `schema_migrations` tracking at the end.
- **Worker pass-through**: The `ingestion_worker.py` call site pattern (`.get("key", default)`) is identical to how other optional fields like `source_type`, `file_path`, and `hostname` are passed.

### 5. Regressions or Issues

**No regressions identified.** Specific checks:

- The `direct_io=True` change in `filesystem.py` only affects `image_document` and `image_evidence` entry types. Text documents, concepts, meta files, and job files are unaffected. The `direct_io` flag causes slightly more overhead per read (bypasses kernel page cache), but this is correct for binary files whose size is not known at open time.

- The test script changes are purely defensive improvements:
  - `cat ... | wc -c` instead of `wc -c < ...` reads actual FUSE content rather than relying on `fstat` (which returns the placeholder size). This is the correct approach for FUSE filesystems where `st_size` may be stale.
  - `grep -qF` (fixed string) instead of `grep -q` with regex metacharacters (`!`, `[`, `]`) avoids shell/regex interpretation issues. Good fix.
  - The retry loop (5 attempts with `sleep 2`) for job cleanup is more robust than the previous single-attempt approach, addressing the noted flaky test.

### Assessment

This is a clean, well-targeted fix. The root cause analysis is correct, the fix addresses both the forward path (new documents) and backward path (existing data), and the changes follow existing code conventions. The migration is safe and idempotent. No blocking issues.

PR #240, opened 2026-01-30
feat: ADR-200 Phase 3a — ontology breathing control surface

aaronsb:
## Summary

Build the manual controls that the breathing worker will need, so we can drive the car before handing it to the autonomous agent.

- **Scoring service** (`ontology_scorer.py`): mass (Michaelis-Menten), coherence (pairwise cosine), exposure (epoch delta + adjacency), protection (sigmoid composite)
- **AGE client** (+12 methods): read controls (stats, degree ranking, affinity, cached scores) + write controls (reassign sources, dissolve ontology, update scores, batch SCOPED_BY)
- **API** (+7 endpoints): GET/POST scores, POST score-all, GET candidates, GET affinity, POST reassign, POST dissolve
- **CLI** (+7 subcommands): `scores`, `score`, `score-all`, `candidates`, `affinity`, `reassign`, `dissolve`
- **MCP** (+7 actions): matching CLI commands on the `ontology` tool
- **Pydantic models** (+11): OntologyStats, OntologyScores, ConceptDegreeRanking, AffinityResult, ReassignRequest/Response, DissolveRequest/Response

Key primitive: `reassign_sources()` — moves sources between ontologies (batched, frozen-aware). This is the missing control that enables both promotion and demotion.

## Test plan

**Automated (115 passed):**
- [x] Unit: OntologyScorer — mass, coherence, exposure, protection, cosine similarity (29 tests)
- [x] Unit: AGE client — stats, ranking, affinity, scores, reassign, dissolve (15 test classes)
- [x] Route: all 7 new endpoints with 200/403/404 paths (13 tests)

**Manual testing against live graph:**
- [ ] `kg ontology score-all` — verify all ontologies get scored
- [ ] `kg ontology scores` — verify cached scores display
- [ ] `kg ontology candidates <name>` — verify ranked concepts appear
- [ ] `kg ontology affinity <name>` — verify cross-ontology overlap data
- [ ] MCP: `ontology` tool with `scores` action — verify JSON response

## Context

- **ADR:** ADR-200 Breathing Ontologies
- **Depends on:** Phases 1-2 (PR #237, #238, #239 — all merged)
- **Next:** Phase 3b (proposals/automation) uses these controls

### DIFF
--- a/api/app/lib/age_client.py
+++ b/api/app/lib/age_client.py
@@ -2406,7 +2406,7 @@ class AGEClient:
         # System relationship types - internal use only, not user-facing vocabulary
         SYSTEM_TYPES = {
             'APPEARS', 'EVIDENCED_BY', 'FROM_SOURCE', 'IN_CATEGORY',
-            'LOAD', 'SET', 'APPEARS'  # LOAD/SET may appear from SQL parsing artifacts
+            'SCOPED_BY', 'LOAD', 'SET',  # LOAD/SET may appear from SQL parsing artifacts
         }
 
         try:
--- a/api/app/routes/ontology.py
+++ b/api/app/routes/ontology.py
@@ -268,8 +268,9 @@ async def get_ontology_info(
         # ADR-200: Check existence via graph node first, then sources
         has_sources = False
         exists_check = client._execute_cypher(
-            f"MATCH (s:Source {{document: '{ontology_name}'}}) RETURN count(s) > 0 as ontology_exists",
-            fetch_one=True
+            "MATCH (s:Source {document: $name}) RETURN count(s) > 0 as ontology_exists",
+            params={"name": ontology_name},
+            fetch_one=True,
         )
         if exists_check and exists_check['ontology_exists']:
             has_sources = True


aaronsb:
## Code Review: ADR-200 Phase 3a -- Ontology Breathing Control Surface

**Scope**: ~3,100 lines added across 17 files. Thorough review -- architecture, algorithms, security, consistency, test coverage.

**Overall assessment**: Well-structured feature addition with clean separation between scoring (pure computation), data access (AGE client), and API surface (routes/CLI/MCP). The plan-driven approach is evident and the implementation tracks closely to the enumerated plan. However, there is one **blocking** issue and several substantive findings.

---

### BLOCKING: Relationship Name Mismatch -- `APPEARS_IN` vs `APPEARS`

**Location**: `/api/app/lib/age_client.py` -- all new Phase 3a Cypher queries (lines 1682-1932)

**Problem**: Every new Phase 3a Cypher query in the AGE client uses `[:APPEARS_IN]` as the relationship type between Concept and Source nodes. The rest of the codebase -- including the routes file in this same PR, the existing AGE client methods, and every other service -- uses `[:APPEARS]`.

Evidence from grep results:

- **`[:APPEARS]`** is used in: `routes/ontology.py`, `routes/queries.py`, `routes/documents.py`, `services/query_service.py`, `services/embedding_worker.py`, `launchers/projection.py`, and the existing `age_client.py` methods (line 563: `MERGE (c)-[:APPEARS]->(s)`)
- **`[:APPEARS_IN]`** is used ONLY in the new Phase 3a methods added in this PR (lines 1682-1932), plus `services/confidence_analyzer.py` (line 162) and one line in `routes/documents.py` (line 652)

The canonical relationship is created in `age_client.py:563` as `[:APPEARS]`. The new Phase 3a queries will silently return zero results for all stats, degree rankings, and affinity calculations because they are querying a relationship type that does not exist on most edges.

**Impact**: All scoring algorithms will compute on empty data. Mass will be 0 (no concepts/relationships found). Coherence will be 0 (no embeddings found). Candidates and affinity will return empty lists. The entire breathing control surface will appear functional but produce meaningless results.

**Fix**: Replace `APPEARS_IN` with `APPEARS` in all new Cypher queries in the Phase 3a section of `age_client.py`. There are 10 occurrences across `get_ontology_stats`, `get_concept_degree_ranking`, `get_cross_ontology_affinity`, and `get_ontology_concept_embeddings`.

---

### Issue: Coherence Docstring Contradicts Implementation

**Location**: `/api/app/lib/ontology_scorer.py:79-123`

**Problem**: The docstring (line 83-84) says "coherence = 1 - diversity" and "Low coherence = concepts are diverse." But the implementation at line 122 computes `avg_similarity` directly -- it does NOT compute `1 - mean_similarity`. The inline comment at line 120 even corrects itself: "Coherence = mean similarity (not 1-diversity, since we want high coherence = high similarity)".

The code is correct as written (high coherence = high pairwise similarity = tight domain). The docstring is a leftover from an earlier design that was revised during implementation.

**Suggestion**: Update the class-level docstring at line 9 ("Coherence: 1 - mean pairwise cosine similarity (Gini-Simpson)") and the method docstring to match the actual implementation. The current text will confuse anyone reading the module header without checking the code.

---

### Issue: `score_all_ontologies` and Root Endpoint Can Be Slow

**Location**: `/api/app/lib/ontology_scorer.py:272-293` and `/api/app/main.py:375-385`

**Problem -- Scorer**: `score_all_ontologies()` calls `score_ontology()` sequentially for each ontology. Each `score_ontology()` makes 6+ Cypher queries (stats has 5, plus coherence embeddings, exposure node lookups, affinity, and the score update). For N ontologies, that is roughly 6N+ database round trips with no timeout or circuit breaker. The `POST /ontology/scores` endpoint that calls this could become very slow as the graph grows.

**Problem -- Root**: The root `/` info endpoint now instantiates a full `AGEClient()` on every health check to read the epoch counter. This adds connection pool overhead to what should be the lightest endpoint in the system.

**Suggestions**:
- For the scorer: Not blocking for Phase 3a (manual use), but worth noting for Phase 3b automation. Consider adding a timeout parameter or making the all-scoring endpoint async with a job.
- For the root endpoint: Consider caching the epoch value (even a simple TTL cache of 30-60 seconds) rather than creating a fresh AGE client per request. Or read it from the same connection pool the app already uses, if one is shared.

---

### Issue: Missing Input Validation on `limit` Parameters

**Location**: `/api/app/routes/ontology.py:960-961` and `:1003`

**Problem**: The `limit` query parameter for `candidates` and `affinity` endpoints defaults to 20 and 10 respectively, but has no upper bound. A client could pass `limit=100000` and generate an expensive unbounded Cypher query.

**Suggestion**: Add `ge=1, le=100` (or similar reasonable cap) to the `QueryParam` definitions, consistent with how other paginated endpoints in the codebase work.

---

### Issue: String-Based Error Routing in Route Layer

**Location**: `/api/app/routes/ontology.py:1067-1074` and `:1118-1125`

**Problem**: The reassign and dissolve endpoints determine HTTP status codes by doing substring matches on error messages returned from the AGE client: `if "not found" in error` and `if "frozen" in error`. This is fragile -- if someone rewrites an error message (e.g., "not found" becomes "does not exist"), the routing breaks silently and falls through to a 500.

**Why it matters**: The AGE client methods already know whether the error is a "not found" vs "forbidden" condition. That information is lost by flattening it into a string.

**Suggestion**: Consider returning a structured error from the AGE client, e.g., `{"error_type": "not_found", "message": "..."}` or using specific exception types. The current pattern works but is a maintenance trap. Since the AGE client methods are new in this PR, this would be a low-cost improvement.

---

### Issue: Reassign Atomicity

**Location**: `/api/app/lib/age_client.py` -- `reassign_sources()` (new method)

**Problem**: The reassign operation performs 3 separate Cypher queries per batch (update document property, delete old SCOPED_BY, create new SCOPED_BY). If the operation fails between steps, you could end up with sources that have their `document` property updated but still have the old SCOPED_BY edge (or no edge at all). The batch loop compounds this -- partial batches could succeed while later ones fail.

The method does report partial success (`sources_reassigned` counts what was done), which is good. But there is no rollback.

**Why it matters**: The dissolve operation depends on reassign. A partial reassign failure during dissolve could leave the ontology in an inconsistent state -- some sources moved, some not, and then the ontology node gets deleted anyway (line in `dissolve_ontology`: `node_deleted = self.delete_ontology_node(name)` runs even after partial reassign failure... actually, looking more carefully, it does return early on reassign failure, so the node is not deleted. That is correct).

**Suggestion**: This is an acceptable tradeoff for Phase 3a (manual operation with human oversight). For Phase 3b automation, consider wrapping the 3 queries in a single transaction or at minimum a compensating cleanup step. Worth documenting as a known limitation.

---

### Positive Observations

**Good separation of concerns**: The `OntologyScorer` is a pure computation class that takes an AGE client dependency. It does not import HTTP or route concerns. This makes it testable and reusable for the Phase 3b worker.

**Lifecycle enforcement is consistent**: Frozen ontology rejection on reassign (AGE client level), dissolve rejection for pinned/frozen (AGE client level). The enforcement lives where the mutations happen, not just at the route layer.

**Graceful degradation**: `get_ontology_stats` returns partial stats on error rather than None. `score_all_ontologies` continues past individual failures. These are production-friendly patterns.

**Test coverage is solid**: 15 AGE client test classes, 7 scorer test classes, 7 route test classes. Edge cases covered include: zero vectors, orthogonal embeddings, frozen enforcement, nonexistent ontologies, race conditions (ensure_ontology_exists), and partial failures in bulk scoring. The tests for the scorer verify mathematical properties (mass never exceeds 1.0, protection can go negative) rather than just asserting on magic numbers.

**Plan traceability**: The implementation matches the enumerated plan almost exactly. Each chunk is identifiable. The plan's verification steps are testable.

---

### File Size Notes

- `age_client.py` is now 3,619 lines. This was already large before this PR (+639 lines). Worth considering whether the ontology-specific methods could eventually live in a dedicated `ontology_client.py` that wraps or extends AGEClient. Not blocking.
- `routes/ontology.py` is 1,143 lines. Approaching the threshold. The Phase 3a endpoints follow the same pattern as existing ones, so this is organic growth, not complexity accumulation.
- `mcp-server.ts` is 2,544 lines. The ontology tool's action enum is now 13 values long. This is becoming a "god tool" from the MCP perspective. Not blocking for this PR but worth considering for future phases.

---

### Summary

| Category | Status |
|----------|--------|
| SOLID compliance | Good -- scorer is SRP, AGE client groups by domain, routes are thin |
| Security | Lifecycle enforcement correct; missing limit bounds on query params |
| Error handling | Functional but fragile string matching for HTTP status routing |
| Algorithm correctness | **BLOCKED** -- APPEARS_IN vs APPEARS mismatch means all queries return empty |
| API consistency | Matches existing patterns well |
| Test coverage | Thorough -- unit + route level, edge cases covered |

**Verdict**: Fix the `APPEARS_IN` / `APPEARS` relationship name mismatch before merge. The coherence docstring and limit validation are secondary but should be addressed.

---
*AI-assisted review via Claude (Opus 4.5)*

Issue #245, opened 2026-01-30
Ontology files endpoint should return source IDs


aaronsb:
## Context

When exercising the Phase 3a reassign workflow, we discovered that neither `GET /ontology/{name}/files` nor `GET /documents?ontology=name` returns source IDs. The reassign endpoint (`POST /ontology/{name}/reassign`) requires source IDs.

During manual testing, we had to drop to raw SQL via `operator.sh query` to discover source IDs:
```sql
SET search_path = ag_catalog, public;
SELECT * FROM cypher('knowledge_graph',
  $$MATCH (s:Source {document: 'distributed-systems'}) RETURN s.source_id LIMIT 12$$
) as (source_id agtype)
```

The cypher query endpoint (`/query/cypher`) returns graph-shaped results (nodes/relationships) that don't expose scalar values like source_id cleanly.

## Gaps Found

1. **Files endpoint** — `GET /ontology/{name}/files` returns file_path, chunk_count, concept_count but no source_ids
2. **Documents endpoint** — `GET /documents?ontology=name` returns document_id but source_ids is null
3. **No CLI path** — `kg ontology files <name>` shows files but no IDs usable with `kg ontology reassign`

## Proposal

Add `source_ids: string[]` to the files endpoint response. Each file may have multiple source chunks:

```json
{
  "file_path": "/tmp/example.txt",
  "chunk_count": 2,
  "concept_count": 12,
  "source_ids": ["sha256:abc_chunk1", "sha256:abc_chunk2"]
}
```

This makes the files → reassign workflow self-contained through both CLI and MCP:
1. `kg ontology files <name>` → see files with source IDs
2. `kg ontology reassign <name> --to <target> --source-ids <ids...>` → move them

## Related

- ADR-200 Phase 3a (reassign sources)
- Discovered during manual exercise of breathing controls

aaronsb:
Fixed — source_ids are now included in the GET /ontology/{name}/files response and the OntologyFileInfo model.

PR #246, opened 2026-01-31
ADR-200 Phase 3b: Ontology breathing worker

aaronsb:
## Summary

Implements Phase 3b of ADR-200 (Breathing Ontologies) — the breathing worker that evaluates ontology health and generates promotion/demotion proposals.

### What's included

**Pre-requisites (Steps 0-2):**
- Source ID discoverability in ontology files endpoint (#245)
- Epoch tagging on concept creation (`created_at_epoch`, `last_seen_epoch`, `seen_count`)
- Exercise report operations log for Phase 3a

**Breathing infrastructure (Step 3):**
- `breathing_evaluator.py` — LLM evaluation for promotion/demotion decisions (OpenAI, Anthropic, Ollama)
- `breathing_manager.py` — Orchestration: score all → identify candidates → LLM evaluate → store proposals
- `breathing_worker.py` + `breathing_launcher.py` — Job system integration with dual-trigger (cron + post-ingestion)
- Weighted top-K centroid recomputation in `ontology_scorer.py`
- Migration 046: `kg_api.breathing_proposals` table
- Migration 047: `kg_api.breathing_options` config table, `last_breathing_epoch` counter, cron schedule
- Pydantic models for proposals and cycle results
- API routes: list/get/review proposals, trigger breathing cycle
- CLI commands: `kg ontology proposals`, `kg ontology proposal <id>`, `kg ontology breathe`
- MCP actions: `proposals`, `proposal_review`, `breathing_cycle`

**Design decisions:**
- Proposals only — no automatic execution (human-in-the-loop, Phase 4 adds execution)
- Database-driven config via `kg_api.breathing_options` (no hardcoded thresholds)
- Epoch delta gating prevents redundant runs from cron + ingestion + manual triggers
- `last_breathing_epoch` lives in `graph_metrics` alongside all other epoch counters

## Test plan

- [ ] 14 unit tests for breathing evaluator (LLM dispatch, JSON parsing, fallbacks)
- [ ] 17 unit tests for breathing manager (candidates, thresholds, lifecycle exclusion, dry-run)
- [ ] 9 route tests for proposal CRUD (list/filter/get/approve/reject/conflict/validation)
- [ ] `kg ontology breathe --dry-run` returns candidates without generating proposals
- [ ] `kg ontology breathe` generates proposals with LLM reasoning
- [ ] `kg ontology proposals` lists pending proposals
- [ ] `kg ontology proposal approve/reject <id>` updates proposal status
- [ ] Verify breathing_options table values are read by launcher
- [ ] Verify epoch consistency — no raw SQL for epoch reads in routes

### DIFF
--- a/api/app/launchers/breathing.py
+++ b/api/app/launchers/breathing.py
@@ -58,21 +58,25 @@ class BreathingLauncher(JobLauncher):
     - Post-ingestion hook (ingestion_worker, after epoch increment)
     - Manual API call (POST /ontology/breathing-cycle)
 
-    The epoch delta check in check_conditions() gates all triggers,
-    so multiple trigger sources cannot cause redundant runs.
+    check_conditions() uses an atomic UPDATE to both verify the epoch
+    delta AND claim the epoch window in one statement. This prevents
+    concurrent triggers from both passing the check.
 
     Parameters are read from kg_api.breathing_options at launch time.
     """
 
     def __init__(self, job_queue, max_retries: int = 5):
         super().__init__(job_queue, max_retries=max_retries)
+        self._cached_options = None
+        self._cached_epoch = None
 
     def check_conditions(self) -> bool:
         """
-        Check if breathing is enabled and enough epochs have passed.
+        Atomically check epoch delta and claim the window.
 
-        Reads configuration from kg_api.breathing_options and epoch state
-        from public.graph_metrics (same table as all other epoch counters).
+        Uses UPDATE ... WHERE to combine the condition check and epoch
+        stamp into a single statement, preventing TOCTOU races when
+        multiple triggers fire concurrently.
         """
         client = AGEClient()
         try:
@@ -85,27 +89,33 @@ class BreathingLauncher(JobLauncher):
                     return False
 
                 current_epoch = client.get_current_epoch()
+                epoch_interval = options["epoch_interval"]
 
+                # Atomic check-and-claim: UPDATE only succeeds if delta is sufficient
                 with conn.cursor() as cur:
                     cur.execute(
-                        "SELECT counter FROM public.graph_metrics "
-                        "WHERE metric_name = 'last_breathing_epoch'"
+                        "UPDATE public.graph_metrics "
+                        "SET counter = %s, updated_at = CURRENT_TIMESTAMP "
+                        "WHERE metric_name = 'last_breathing_epoch' "
+                        "AND %s - counter >= %s "
+                        "RETURNING counter",
+                        (current_epoch, current_epoch, epoch_interval)
                     )
                     row = cur.fetchone()
-                    last_epoch = row[0] if row else 0
+                conn.commit()
 
-                epoch_interval = options["epoch_interval"]
-                delta = current_epoch - last_epoch
-
-                if delta >= epoch_interval:
+                if row:
+                    self._cached_options = options
+                    self._cached_epoch = current_epoch
                     logger.info(
-                        f"BreathingLauncher: epoch delta {delta} >= {epoch_interval}, "
-                        f"triggering cycle (current={current_epoch}, last={last_epoch})"
+                        f"BreathingLauncher: claimed epoch {current_epoch} "
+                        f"(interval={epoch_interval}), triggering cycle"
                     )
                     return True
 
                 logger.debug(
-                    f"BreathingLauncher: epoch delta {delta} < {epoch_interval}, skipping"
+                    f"BreathingLauncher: epoch delta < {epoch_interval} "
+                    f"or already claimed, skipping"
                 )
                 return False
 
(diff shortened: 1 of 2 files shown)


aaronsb:
## Review: Fix Commits for Code Review Findings #1-#8

Reviewing the three fix commits addressing prior code review findings on this PR. Combined diff: 6 files, +164 / -161 lines.

---

### Finding #2: TOCTOU Race in Launcher -- Atomic UPDATE...RETURNING

**Files**: `api/app/launchers/breathing.py` (lines 94-105, 128-145)

**Assessment**: Correct and well-executed.

The original two-phase read-then-write pattern is replaced by a single `UPDATE ... WHERE metric_name = 'last_breathing_epoch' AND %s - counter >= %s RETURNING counter`. This is a textbook atomic check-and-claim. Two concurrent callers cannot both succeed because only one UPDATE will match the WHERE clause and return a row; the other sees `row = None`.

The caching design (`_cached_options`, `_cached_epoch`) that feeds `prepare_job_data()` is a smart consequence of this change. Since the epoch is now claimed inside `check_conditions()`, re-reading it in `prepare_job_data()` would be wasteful and could even read a newer epoch than what was claimed. Using cached values maintains consistency between the two methods.

One thing worth noting: the `_cached_options` and `_cached_epoch` instance variables are not cleared after `prepare_job_data()` runs, so they persist across the launcher's lifetime. In the current single-shot `launch()` flow from `base.py`, this is fine -- `check_conditions()` always runs first and resets them. But if the launcher instance were ever reused across cycles where conditions were *not* met (returning False), stale cached values would persist. The fallback `self._cached_options or DEFAULTS` in `prepare_job_data()` provides a safety net, so this is not a blocking concern, just something to be aware of if the launcher lifecycle changes in future phases.

---

### Finding #4: Sync LLM in Async Function -- asyncio.to_thread

**Files**: `api/app/lib/breathing_evaluator.py` (lines 210-280)

**Assessment**: Correct and clean separation.

The synchronous provider I/O (OpenAI, Anthropic, Ollama HTTP calls) is extracted into `_call_llm_sync()` and then called via `await asyncio.to_thread(...)` from the existing `async def _call_llm()`. This unblocks the event loop during what could be multi-second LLM inference calls.

The decomposition is well done: `_call_llm_sync` handles provider dispatch and returns raw content; `_call_llm` handles the async boundary, JSON parsing, markdown extraction, and fallback. Responsibilities are cleanly separated.

One consideration: `asyncio.to_thread` uses the default thread pool executor. If multiple breathing evaluations ran concurrently (currently they don't -- they're sequential in `run_breathing_cycle`), thread pool exhaustion could occur under high load. Not a concern for Phase 3b's sequential execution, but worth a note if Phase 4 parallelizes evaluations.

---

### Finding #3: Empty top_neighbors -- Populated via Cypher Query

**Files**: `api/app/services/breathing_manager.py` (lines 288-301)

**Assessment**: Correct. The placeholder `top_neighbors=[]` is replaced with a real Cypher neighbor query:

```cypher
MATCH (c:Concept {concept_id: $cid})-[r]-(n:Concept)
RETURN DISTINCT n.label as label, count(r) as rel_count
ORDER BY rel_count DESC
LIMIT 10
```

This gives the LLM meaningful context for distinguishing "nucleus" concepts (coherent neighbors) from "crossroads" concepts (diverse, unrelated neighbors). The `try/except` fallback to an empty list is appropriate defensive coding.

Minor observation: the `DISTINCT` on `n.label` combined with `count(r)` will de-duplicate labels but the count may not reflect the true relationship count if a neighbor appears through multiple paths. In practice this is unlikely to matter for the LLM's qualitative assessment.

---

### Finding #6: Triple-Duplicated Row Mapping -- Extracted _row_to_proposal

**Files**: `api/app/routes/ontology.py` (lines 234-259)

**Assessment**: Clean DRY extraction. Three identical 15-line inline mappings replaced by a single `_row_to_proposal()` function with clear documentation of the expected column order.

The docstring documenting the column order contract is important here -- positional SQL result mapping is inherently fragile. If the SELECT column list ever changes, only one place needs updating now instead of three. The column order documentation makes that contract explicit, which is the right approach.

---

### Finding #1: Review Status Server-Side Guard

**Files**: `api/app/routes/ontology.py` (lines 394-396)

**Assessment**: Correct belt-and-suspenders approach. The Pydantic model already validates via `Field(..., pattern="^(approved|rejected)$")`, so the server-side check at line 395 is strictly redundant under normal API usage. However, it protects against two scenarios: (a) a future refactor that loosens the Pydantic constraint, and (b) direct internal calls to `review_proposal` that bypass FastAPI's request validation. The comment at line 394 ("Belt-and-suspenders: Pydantic validates via regex, but guard here too") documents the intent clearly.

---

### Finding #7: Expiration Enforcement on Proposal Listing

**Files**: `api/app/routes/ontology.py` (lines 282-287)

**Assessment**: Correct. When no explicit `status` filter is provided, the query now adds `(expires_at IS NULL OR expires_at > NOW())` to hide expired proposals from default views. When a user explicitly requests `status=expired` (or any other status), the expiration filter is not applied, allowing them to see the full history.

The `IS NULL` guard handles proposals that predate the `expires_at` column or were created without a default expiry.

One gap to flag: the `expires_at` column has no index. The `idx_breathing_proposals_status` index covers `status` queries, but the `expires_at > NOW()` predicate in the default (no-status) path does a sequential scan against `expires_at`. For Phase 3b volumes this is fine (proposals are generated in small batches of `max_proposals`, capped at 20). If proposal volume grows significantly in Phase 4, consider adding an index on `expires_at`.

---

### Finding #5: Missing Migration 046 Recording

**Files**: `schema/migrations/046_breathing_proposals.sql` (lines 33-35)

**Assessment**: Correct. The `INSERT INTO public.schema_migrations (version, name) VALUES (46, 'breathing_proposals') ON CONFLICT (version) DO NOTHING` follows the exact pattern used in all other migrations (045, 044, 043, etc.). The `ON CONFLICT DO NOTHING` makes it idempotent for re-runs.

---

### Finding #8: Worker NameError Bug -- Guarded age_client

**Files**: `api/app/workers/breathing_worker.py` (lines 57-100)

**Assessment**: Correct fix for what was a real crash bug. Previously, `age_client = AGEClient()` was assigned before the try block, but `scorer`, `ai_provider`, and `manager` were also assigned before the try block. If `AGEClient()` itself raised (e.g., database down), the `finally: age_client.close()` would have hit a `NameError` since `age_client` was never assigned.

The fix initializes `age_client = None` before the try block and guards the close with `if age_client:`. All dependent setup (`scorer`, `ai_provider`, `manager`) is now inside the try block where failures are properly handled. The error return dict on line 93-96 correctly propagates the exception message to the job queue.

---

### Summary

All eight findings are addressed correctly. The fixes are targeted and do not introduce regressions. In particular:

- **No new code paths left untested in a dangerous way** -- the atomic UPDATE is safer than what it replaced, and `_call_llm_sync` is a pure extraction of existing code.
- **No new coupling** -- the caching pattern in the launcher keeps state within the single `launch()` call flow.
- **No security concerns** -- the server-side status guard adds defense-in-depth.

**Suggestions for future work** (non-blocking):
- Consider clearing `_cached_options` / `_cached_epoch` at the end of `prepare_job_data()` to prevent stale state if the launcher lifecycle evolves.
- If proposal volume grows in Phase 4, consider adding an index on `expires_at` for the default listing query.
- The positional SQL mapping in `_row_to_proposal` works well for now; if more columns are added over time, a `dict` cursor (`RealDictCursor`) would eliminate the column-order fragility entirely.

Issue #242, opened 2026-01-30
Parameterize f-string Cypher queries in routes


aaronsb:
## Context

During code review of PR #240, a pre-existing Cypher injection pattern was identified in the route layer. Several queries interpolate user-supplied values directly into openCypher strings via f-strings instead of using parameterized queries.

The Phase 3a code in `age_client.py` correctly uses parameterized queries throughout. One instance in `ontology.py` was fixed in PR #240 (commit 07a54569), but the pattern is broader.

## Scope

**Fixed (PR #240):**
- `api/app/routes/ontology.py:271` — ontology info existence check

**Remaining:**
- `api/app/routes/queries.py` — ~15 f-string Cypher queries interpolating `concept_id` values
- `api/app/routes/database.py:57` — interpolates node label into MATCH

Example of the pattern:
```python
# Current (injectable):
f"MATCH (c:Concept {{concept_id: '{concept_id}'}}) RETURN ..."

# Should be:
"MATCH (c:Concept {concept_id: $concept_id}) RETURN ...",
params={"concept_id": concept_id}
```

## Risk

The `concept_id` values in `queries.py` come from prior graph query results (not direct user input), so the practical risk is low. But parameterized queries are the correct pattern — they're safer, easier to read, and avoid the `{{` escaping noise.

## Approach

Convert each f-string Cypher call to use the `params` dict pattern that `_execute_cypher()` already supports. Mechanical refactor, no logic changes.

## Related

- PR #240 (fixed one instance)
- ADR-048 (query safety & GraphQueryFacade)

Issue #236, opened 2026-01-30
graph tool: edge creation fails with 422 when category parameter is specified


aaronsb:
## Summary

When creating edges via the `graph` tool (both single `create` and `queue` actions), specifying the `category` parameter causes a `422 Unprocessable Entity` error from the API, even when using values that are listed as valid in the tool's JSON schema.

## Steps to Reproduce

1. Create two concepts successfully
2. Attempt to create an edge between them with an explicit `category`:

```json
{
  "action": "create",
  "entity": "edge",
  "from_concept_id": "c_xxx",
  "to_concept_id": "c_yyy",
  "relationship_type": "COMPOSED_OF",
  "category": "conceptual"
}
```

**Result:** `422 Unprocessable Entity`

3. Retry the same edge without `category`:

```json
{
  "action": "create",
  "entity": "edge",
  "from_concept_id": "c_xxx",
  "to_concept_id": "c_yyy",
  "relationship_type": "COMPOSED_OF"
}
```

**Result:** Success — edge created with `category: "logical_truth"` (the default)

## Expected Behavior

The `category` parameter should accept any value from its declared enum: `logical_truth`, `empirical_claim`, `value_judgment`, `conceptual`, `temporal_sequence`, `causal`, `other`.

## Actual Behavior

Any explicit `category` value (tested with `"conceptual"`) produces a 422. Only omitting the parameter (which defaults to `"logical_truth"`) works.

## Impact

- All edge categories are forced to `logical_truth` since it's the only usable value
- This limits the semantic richness of the graph — you can't distinguish empirical claims from conceptual relationships from causal links
- The `queue` action also fails on first error when any operation includes `category`, halting the entire batch

## Environment

- MCP server: `knowledge-graph` 
- Client: Claude Code CLI
- Observed during manual graph construction via `graph` tool

Issue #181, opened 2025-12-15
Admin role missing ontology delete permission


aaronsb:
## Summary

The default `admin` role doesn't have `delete` permission on `ontologies` resource, but `platform_admin` does. This creates a confusing situation where the admin user can't delete ontologies through the CLI without manual permission grants.

## Steps to Reproduce

1. Fresh `./operator.sh init`
2. Login as admin: `kg login -u admin`
3. Try to delete an ontology: `kg ontology delete "Test" --force`
4. Get: `Permission denied: delete on ontologies`

## Expected Behavior

The `admin` role should include `delete` on `ontologies` since admins typically manage ontology lifecycle.

## Current Workaround

Manually grant the permission via SQL:
```sql
INSERT INTO kg_auth.role_permissions (role_name, resource_type, action, scope_type, granted)
VALUES ('admin', 'ontologies', 'delete', 'global', true);
```

## Fix

Add this permission to the default role setup in the auth migration or seed script.

## Related

- ADR-028 RBAC implementation
- Default role permissions in `schema/migrations/` or seed scripts

Issue #177, opened 2025-12-14
Migrate callers from GarageClient wrapper to modular services


aaronsb:
## Context

PR #176 refactored the monolithic `GarageClient` into modular services (ADR-080). A backward-compatible wrapper preserves existing imports, but callers should eventually migrate to the focused services.

## Current State

9 import sites use the old pattern:
```python
from api.api.lib.garage_client import get_garage_client
garage = get_garage_client()
garage.upload_image(...)
```

## Target State

```python
from api.api.lib.garage import get_image_storage
images = get_image_storage()
images.upload(...)
```

## Benefits

- **Interface Segregation**: Callers only see methods they need
- **Clearer intent**: `get_image_storage()` vs generic `get_garage_client()`
- **Easier testing**: Mock only the service you're using

## Files to Migrate

- [ ] `api/api/routes/ontology.py` (line 315)
- [ ] `api/api/routes/sources.py` (line 13)
- [ ] `api/api/workers/ingestion_worker.py` (line 108)
- [ ] `api/api/workers/projection_worker.py` (lines 187, 218, 242, 265)
- [ ] `tests/unit/lib/test_garage_projection.py` (line 64)

## Notes

- Low priority - backward compat wrapper works fine
- Can be done incrementally, one file at a time
- Consider after ADR-081 implementation is complete

## References

- PR #176: Garage Service Architecture refactor
- ADR-080: Garage Service Architecture
- Code review Issue #6 (Interface Segregation observation)

PR #247, opened 2026-01-31
Fix edge category 422, admin permissions, GarageClient migration, Cypher injection

aaronsb:
## Summary

Batch of fixes addressing 4 open issues:

- **#236**: MCP server's edge category enum was completely mismatched with the API's Pydantic enum (ADR-022). Only `logical_truth` and `causal` overlapped — every other category caused a 422. Fixed in `mcp-server.ts` and `graph-operations.ts`.
- **#181**: Admin role was missing `ontologies:delete` permission. Added to migration 028.
- **#177**: Migrated all 5 remaining `get_garage_client()` callers to modular services (`get_image_storage`, `get_projection_storage`). No callers of the backward-compat wrapper remain. Updated projection tests to use `ProjectionStorageService` directly.
- **#242**: Parameterized all f-string Cypher queries in `queries.py` — replaced concept_id interpolation with `$cid` params, validated epistemic status filter against allowlist.

## Test plan

- [x] MCP edge creation with all 7 category values — verified via live test
- [x] 94 directly affected tests passing
- [x] 720 total tests passing (pre-existing failures in synonym_detector/polarity_axis unchanged)
- [ ] Login as admin, `kg ontology delete <name> --force` — should succeed
- [ ] Verify search/concept detail endpoints still work with parameterized queries

### DIFF
--- a/api/app/lib/pathfinding_facade.py
+++ b/api/app/lib/pathfinding_facade.py
@@ -41,8 +41,19 @@ Usage:
 
 from typing import List, Dict, Optional, Any, Set, Tuple
 import logging
+import re
 
 logger = logging.getLogger(__name__)
+
+# Valid Cypher relationship type: uppercase letters, digits, underscores
+_VALID_REL_TYPE_RE = re.compile(r'^[A-Z][A-Z0-9_]*$')
+
+
+def _validate_rel_types(types: List[str]) -> None:
+    """Validate relationship type names for safe Cypher interpolation."""
+    for t in types:
+        if not _VALID_REL_TYPE_RE.match(t):
+            raise ValueError(f"Invalid relationship type name: {t!r}")
 
 
 class PathfindingFacade:
@@ -290,6 +301,7 @@ class PathfindingFacade:
         # Build relationship type filter
         rel_filter = ""
         if allowed_rel_types and len(allowed_rel_types) > 0:
+            _validate_rel_types(allowed_rel_types)
             type_list = ", ".join([f"'{t}'" for t in allowed_rel_types])
             rel_filter = f"AND type(r) IN [{type_list}]"
 
@@ -416,24 +428,24 @@ class PathfindingFacade:
 
     def _get_concept_data(self, concept_id: str) -> Optional[Dict[str, Any]]:
         """Fetch concept node data."""
-        query = f"""
-            MATCH (c:Concept {{concept_id: '{concept_id}'}})
+        query = """
+            MATCH (c:Concept {concept_id: $cid})
             RETURN c.concept_id as concept_id, c.label as label, c.description as description
         """
         try:
-            results = self.db._execute_cypher(query, fetch_one=True)
+            results = self.db._execute_cypher(query, params={"cid": concept_id}, fetch_one=True)
             return results if results else None
         except Exception:
             return None
 
     def _concept_exists(self, concept_id: str) -> bool:
         """Check if concept exists."""
-        query = f"""
-            MATCH (c:Concept {{concept_id: '{concept_id}'}})
+        query = """
+            MATCH (c:Concept {concept_id: $cid})
             RETURN count(c) as cnt
         """
         try:
-            result = self.db._execute_cypher(query, fetch_one=True)
+            result = self.db._execute_cypher(query, params={"cid": concept_id}, fetch_one=True)
             return result and result.get('cnt', 0) > 0
         except Exception:
             return False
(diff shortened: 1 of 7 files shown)


aaronsb:
## Code Review: Commit a483a525 -- Parameterize Cypher queries to prevent injection

**Scope**: `api/app/routes/vocabulary.py`, `api/app/routes/ontology.py`, `api/app/routes/documents.py`, `api/app/services/query_service.py`, `api/app/lib/pathfinding_facade.py`

**Verdict**: Solid security hardening. No correctness regressions found. Two items worth noting below -- one medium concern and one observation for future work.

---

### 1. Correctness: All parameterizations look correct

**ontology.py** -- All 8 `ontology_name` interpolations correctly replaced with `$name` param + `params={"name": ontology_name}`. The `fetch_one=True` flags are preserved where they were before. No semantic changes to the queries themselves.

**vocabulary.py** -- Both `match_vocab_types` calls now use `$status` / `$name` params. I confirmed `query_facade.py:match_vocab_types()` already accepts `params: Optional[Dict] = None` and passes it through to `_execute_cypher`, so this is correctly wired.

**documents.py** -- The branch-query approach (if/else instead of f-string WHERE concatenation) preserves identical query semantics for both branches. The `count_query` correctly receives `ontology_pattern` only in the filtered branch (line 697).

**pathfinding_facade.py** -- `_get_concept_data` and `_concept_exists` are cleanly parameterized with `$cid`. The `_find_additional_paths` method correctly uses `$from_id` / `$to_id` params, and the `result_limit` integer is computed before interpolation (not user-controlled).

**query_service.py** -- Validation added at all three call sites: `build_related_concepts_query`, `build_shortest_path_query`. Good.

### 2. Regex guard `^[A-Z][A-Z0-9_]*$` -- Sufficient for its purpose

The regex requires uppercase ASCII letter as first character, then only uppercase letters, digits, and underscores. This is a correct allowlist for Cypher relationship type identifiers. It prevents any injection of Cypher syntax characters (quotes, brackets, pipes within a single type name, etc.). The `|` joining of types happens *after* individual validation, so each type is guaranteed safe before assembly.

Minor note: the regex is declared identically in both `query_service.py` and `pathfinding_facade.py`. Not a problem for correctness, but if this pattern evolves (e.g., to also allow lowercase types from some future schema), having two copies means two places to update. Consider extracting to a shared utility at some point.

### 3. Medium concern: `_get_neighbors_batch` still interpolates `node_ids` without parameterization

**Location**: `/home/aaron/Projects/ai/knowledge-graph-system/api/app/lib/pathfinding_facade.py`, lines 323-324

```python
ids_str = ", ".join([f"'{id}'" for id in node_ids])
query_substituted = query.replace("$ids", f"[{ids_str}]")
```

This was not introduced by this commit (it is pre-existing), and the `node_ids` here are concept IDs from the internal BFS frontier -- they originate from *previous query results*, not directly from user input. So the practical risk is low. However, a comment in the code says "AGE doesn't handle list parameters well", which explains the workaround. If AGE list param support has improved, parameterizing this would complete the hardening of pathfinding_facade.py. Flagging for awareness, not as a blocker.

### 4. Observation: Regex injection in `ontology_pattern`

**Location**: `/home/aaron/Projects/ai/knowledge-graph-system/api/app/routes/documents.py`, lines 226 and 652

```python
params["ontology_pattern"] = f"(?i).*{ontology_filter}.*"
```

The `ontology_filter`/`ontology` value is a user-supplied query parameter that gets embedded into a Cypher regex pattern. While the value is *parameterized* (passed as `$ontology_pattern`), the *regex itself* is constructed from unsanitized user input. A user could supply `ontology=.*` or other regex metacharacters that alter the matching behavior. This is not a Cypher injection (the parameterization prevents that), but it is a regex injection that could cause unexpected matching or ReDoS on pathological patterns. The same pattern exists in `sources.py` (line 87, not part of this commit). This is pre-existing and not in scope for this PR, but worth a follow-up issue.

### 5. No regressions detected

- All `fetch_one=True` arguments are preserved where they existed before.
- All `params` dicts use consistent key names matching their `$param` references.
- The documents.py branch queries produce byte-identical Cypher for each branch compared to the original dynamic query.
- The `WITH` keyword casing is consistent throughout (Cypher is case-insensitive for keywords anyway, but the style is maintained).

### Summary

| Area | Status |
|------|--------|
| Parameterization correctness | Pass -- all substitutions match original semantics |
| Completeness within scope | Pass -- all targeted f-string interpolations converted |
| Regex guard sufficiency | Pass -- strict uppercase ASCII allowlist |
| Regressions | None found |
| Pre-existing concern | `_get_neighbors_batch` node_id interpolation (low risk, internal data) |
| Pre-existing concern | Regex injection via `ontology_pattern` (not a Cypher injection, but regex metachar pass-through) |

Good to merge.

PR #248, opened 2026-01-31
feat(ontology): derive inter-ontology edges (ADR-200 Phase 5)

aaronsb:
## Summary

- Derive OVERLAPS, SPECIALIZES, and GENERALIZES edges between ontologies based on shared concept overlap during breathing cycles
- New `ontology edges` CLI command and MCP `edges` action to query derived relationships
- MCP formatters for both `edges` and `affinity` actions (structured markdown instead of raw JSON)

## What's new

**API** (`api/app/`):
- `AGEClient` ontology edge CRUD: upsert, query, delete stale edges
- `OntologyScorer` computes overlap/specialization scores between ontology pairs
- `BreathingManager` calls edge derivation after each breathing cycle
- REST routes: `GET /ontology/{name}/edges`
- Schema migration 048 adds `ontology_edges` relational table

**CLI/MCP** (`cli/src/`):
- `kg ontology edges <name>` command
- `formatOntologyEdges()` and `formatOntologyAffinity()` MCP formatters
- API client methods for edge queries

**Tests** (`tests/`):
- Unit tests for AGEClient ontology edge methods
- Unit tests for OntologyScorer
- Unit tests for BreathingManager edge derivation integration
- Route tests for the edges endpoint

## Test plan

- [ ] Run `pytest tests/unit/lib/test_age_client_ontology.py`
- [ ] Run `pytest tests/unit/lib/test_ontology_scorer.py`
- [ ] Run `pytest tests/unit/services/test_breathing_manager.py`
- [ ] Run `pytest tests/api/test_ontology_routes.py`
- [ ] Verify `kg ontology edges <name>` returns formatted output
- [ ] Verify MCP `ontology` tool with `action: edges` returns markdown
- [ ] Verify MCP `ontology` tool with `action: affinity` returns markdown

### DIFF
--- a/api/app/lib/age_client.py
+++ b/api/app/lib/age_client.py
@@ -2309,6 +2309,7 @@ class AGEClient:
     # =========================================================================
 
     VALID_ONTOLOGY_EDGE_TYPES = ("OVERLAPS", "SPECIALIZES", "GENERALIZES")
+    VALID_ONTOLOGY_EDGE_SOURCES = ("breathing_worker", "manual")
 
     def upsert_ontology_edge(
         self,
@@ -2341,6 +2342,11 @@ class AGEClient:
             raise ValueError(
                 f"Invalid ontology edge type: {edge_type!r}. "
                 f"Must be one of {self.VALID_ONTOLOGY_EDGE_TYPES}"
+            )
+        if source not in self.VALID_ONTOLOGY_EDGE_SOURCES:
+            raise ValueError(
+                f"Invalid ontology edge source: {source!r}. "
+                f"Must be one of {self.VALID_ONTOLOGY_EDGE_SOURCES}"
             )
 
         query = f"""
@@ -2511,6 +2517,45 @@ class AGEClient:
                 logger.error(f"Failed to delete all derived {etype} edges: {e}")
 
         return total_deleted
+
+    def delete_ontology_edge(
+        self, from_name: str, to_name: str, edge_type: str
+    ) -> int:
+        """
+        Delete a specific ontology-to-ontology edge (any source).
+
+        Args:
+            from_name: Source ontology name
+            to_name: Target ontology name
+            edge_type: OVERLAPS, SPECIALIZES, or GENERALIZES
+
+        Returns:
+            Number of edges deleted (0 or 1)
+        """
+        if edge_type not in self.VALID_ONTOLOGY_EDGE_TYPES:
+            raise ValueError(
+                f"Invalid ontology edge type: {edge_type!r}. "
+                f"Must be one of {self.VALID_ONTOLOGY_EDGE_TYPES}"
+            )
+
+        query = f"""
+        MATCH (a:Ontology {{name: $from_name}})-[r:{edge_type}]->(b:Ontology {{name: $to_name}})
+        DELETE r
+        RETURN count(r) as deleted
+        """
+        try:
+            result = self._execute_cypher(
+                query,
+                params={"from_name": from_name, "to_name": to_name},
+                fetch_one=True,
+            )
+            return int(str(result.get("deleted", 0))) if result else 0
+        except Exception as e:
+            logger.error(
+                f"Failed to delete {edge_type} edge "
+                f"{from_name} -> {to_name}: {e}"
+            )
+            return 0
 
     # =========================================================================
     # Vocabulary Management Methods (ADR-032)
--- a/api/app/routes/ontology.py
+++ b/api/app/routes/ontology.py
@@ -1453,19 +1453,8 @@ async def delete_ontology_edge(
 
     client = get_age_client()
     try:
-        # Delete the specific edge
-        query = f"""
-        MATCH (a:Ontology {{name: $from_name}})-[r:{edge_type}]->(b:Ontology {{name: $to_name}})
-        DELETE r
-        RETURN count(r) as deleted
-        """
-        result = client._execute_cypher(
-            query,
-            params={"from_name": ontology_name, "to_name": to_ontology},
-            fetch_one=True,
-        )
+        deleted = client.delete_ontology_edge(ontology_name, to_ontology, edge_type)
 
-        deleted = result.get("deleted", 0) if result else 0
         if deleted == 0:
             raise HTTPException(
                 status_code=404,
--- a/api/app/services/breathing_manager.py
+++ b/api/app/services/breathing_manager.py
@@ -52,8 +52,8 @@ class BreathingManager:
         3. Derive ontology-to-ontology edges (Phase 5)
         4. Identify demotion candidates (low protection)
         5. Identify promotion candidates (high-degree concepts)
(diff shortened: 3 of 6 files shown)


aaronsb:
## Code Review: feat(ontology): derive inter-ontology edges (ADR-200 Phase 5)

**Scope**: 24 files, +1651/-33 lines across API (Python), CLI/MCP (TypeScript), schema, and tests.

Overall this is a well-structured feature addition that follows existing codebase patterns closely. The architecture is clean -- edge derivation logic lives in `OntologyScorer`, CRUD in `AGEClient`, orchestration in `BreathingManager`, and the REST/CLI/MCP layers are thin wrappers. Tests are thorough. There are a few issues worth addressing before merge, detailed below by severity.

---

### SECURITY: Cypher f-string interpolation of `edge_type` in route layer

**Location**: `api/app/routes/ontology.py:1457-1461`

**Problem**: The `delete_ontology_edge` route constructs a Cypher query using f-string interpolation of `edge_type` from the URL path parameter:

```python
query = f"""
MATCH (a:Ontology {{name: $from_name}})-[r:{edge_type}]->(b:Ontology {{name: $to_name}})
DELETE r
RETURN count(r) as deleted
"""
```

The allowlist check at line 1446-1452 **does** guard this -- `edge_type` must be one of `("OVERLAPS", "SPECIALIZES", "GENERALIZES")` before the f-string is reached. So there is no actual injection vulnerability here.

However, there is an **encapsulation concern**: this route constructs raw Cypher and calls `client._execute_cypher()` directly (a private method), bypassing the `AGEClient` API. This is the only place in the new code that does this. All other Cypher operations properly go through `AGEClient` methods. This also means the route layer depends on the internal implementation of `AGEClient`, which is a Dependency Inversion concern.

**Suggestion**: Move this into an `AGEClient.delete_ontology_edge()` method that takes `from_name`, `to_name`, and `edge_type` (with the same allowlist validation the other methods use). The route should call that method. This keeps Cypher construction confined to `AGEClient` where all the other query logic lives, consistent with the pattern established by `upsert_ontology_edge()` and `delete_derived_ontology_edges()`.

**Severity**: Medium. The code is safe today, but the pattern inconsistency creates a maintenance risk -- if someone copies this pattern elsewhere without the allowlist guard, it becomes an injection vector.

---

### DESIGN: Step numbering collision in breathing_manager.py

**Location**: `api/app/services/breathing_manager.py:99` and `api/app/services/breathing_manager.py:128`

**Problem**: After renumbering, there are two steps labeled "5":
- Line 99: `# 5. Identify promotion candidates`  
- Line 128: `# 5. Evaluate and store proposals`

The comment at line 128 was not updated when the steps were renumbered from 4/5/6 to 5/6/7.

**Severity**: Low -- cosmetic only, but could confuse future readers.

---

### DESIGN: `derive_ontology_edges` runs during dry_run

**Location**: `api/app/services/breathing_manager.py:78-91`

**Problem**: The `derive_edges` step (step 3) runs **before** the dry_run check at line 105. This means `dry_run=True` still creates/deletes real edges in the graph. The dry_run flag is documented as: "If True, identify candidates but don't call LLM or store."

This might be intentional -- scoring and centroids (steps 1-2) also run during dry_run. But edge creation/deletion is a **write operation** with side effects (deleting all existing derived edges then recreating them), which feels different from the read-only scoring operations.

**Suggestion**: Consider either:
1. Making edge derivation conditional on `not dry_run` (matching the proposal generation behavior), or
2. Documenting explicitly that dry_run still performs edge derivation (so the count fields in the dry_run response are accurate).

Either approach is valid, but the current behavior should be a conscious choice rather than an oversight.

**Severity**: Medium -- could surprise users who expect dry_run to be side-effect-free.

---

### DESIGN: `source` parameter not validated in `upsert_ontology_edge`

**Location**: `api/app/lib/age_client.py:2321` (the `source` parameter)

**Problem**: The method validates `edge_type` against `VALID_ONTOLOGY_EDGE_TYPES` but accepts any string for `source`. The migration and docstrings establish a contract of `'breathing_worker' | 'manual'`, but this is not enforced. If someone passes `source="manual"` when they mean `source="breathing_worker"`, those edges would persist across breathing cycles since the delete methods filter on `r.source = 'breathing_worker'`.

**Suggestion**: Add a `VALID_EDGE_SOURCES = ("breathing_worker", "manual")` constant and validate at the top of the method, similar to the edge_type check.

**Severity**: Low -- both callers currently pass correct values, but defensive validation matches the existing pattern.

---

### ARCHITECTURE: File size note (informational, not blocking)

- `api/app/lib/age_client.py`: 3861 lines (+208 in this PR)
- `api/app/routes/ontology.py`: 1589 lines (+180 in this PR)

Both files are well past the 500-line threshold. The new code follows existing organization patterns (section headers, consistent method signatures), so the additions are reasonable. But this is worth flagging as ongoing technical debt -- `age_client.py` in particular is accumulating a lot of responsibility. A future refactoring to split it by domain (concept operations, ontology operations, vocabulary operations, etc.) would improve maintainability.

This is not a concern for this PR specifically -- just noting the trajectory.

---

### POSITIVE OBSERVATIONS

Several things this PR does well:

1. **Security pattern**: The `VALID_ONTOLOGY_EDGE_TYPES` allowlist in `AGEClient` is the correct pattern for dynamic Cypher relationship types. Edge type is validated before any f-string interpolation. All node property values use `$parameter` binding. This follows the security fix from commit a483a525.

2. **Non-fatal edge derivation**: The `try/except` in `breathing_manager.py:80-91` that catches edge derivation failures without aborting the breathing cycle is a good resilience pattern. The corresponding test (`test_edge_derivation_failure_is_non_fatal`) verifies this.

3. **Full refresh strategy**: Deleting all derived edges then recreating them (rather than incremental updates) is simpler and eliminates stale-edge accumulation. Good tradeoff for what should be a small number of ontology-level edges.

4. **Test coverage**: 648 lines of tests across 4 files. All new AGEClient methods have unit tests including error paths. The breathing manager integration tests cover the enable/disable flag, threshold passthrough, and failure resilience. Route tests cover happy path, 404, 400, and 422 cases.

5. **MCP formatters**: Moving from raw JSON to structured markdown for `affinity` and adding a new `edges` formatter is a good pattern. The TypeScript interfaces in the formatter file duplicate the types from `types/index.ts` -- this is minor but could be cleaned up with a shared import.

6. **Schema migration**: Clean, with verification checks and `ON CONFLICT DO NOTHING` for idempotency.

---

### SUMMARY

| Category | Finding | Severity |
|----------|---------|----------|
| Encapsulation | Route layer constructs Cypher directly instead of using AGEClient method | Medium |
| Behavior | `dry_run` still performs edge write operations | Medium |
| Comment | Step numbering collision (two "5"s) | Low |
| Validation | `source` parameter not validated against known values | Low |
| Architecture | File sizes growing (informational) | Info |

The medium items are worth addressing before merge. The low items can be addressed in this PR or tracked as follow-up.

---
*AI-assisted review via Claude (Opus 4.5)*

PR #263, opened 2026-02-01
feat: graph-accel core engine — Phase 1 of ADR-201

aaronsb:
## Summary

- Rust workspace (`graph-accel/`) with pure traversal engine and benchmark harness — Phase 1 of the in-memory graph acceleration extension ([ADR-201](docs/architecture/database-schema/ADR-201-in-memory-graph-acceleration-extension.md))
- `graph-accel-core`: adjacency list, BFS neighborhood, shortest path. 13 unit tests. No Postgres dependencies.
- `graph-accel-bench`: 6 synthetic topology generators (L-system, scale-free, small-world, Erdős-Rényi, barbell, DLA) with traversal benchmarks at configurable scale
- Validated at 5M nodes / 50M edges — bounded-depth queries in milliseconds, full-graph traversal in seconds
- ADR-201 updated with benchmark data, projected speedup tables, load trigger model, memory usage analysis

### Benchmark highlights (5M nodes, 50M edges, single-threaded)

| Topology | Memory | Depth 5 BFS | Full graph BFS | Shortest path |
|----------|--------|-------------|----------------|---------------|
| Scale-free | 2.4GB | 15.0s (5M nodes) | 15.0s | 3 hops / 2.0s |
| Small-world | 2.4GB | 20ms (18K nodes) | 11.1s | 1 hop / 1.7ms |
| Barbell | 3.9GB | 7.1s (2.5M nodes) | 25.2s | 21 hops / 8.8s |
| DLA | 1.0GB | 3.2ms (4.5K nodes) | 8.3s | 17 hops / 1.3s |

For comparison, AGE hangs indefinitely at depth 6 on a 200-node graph.

### What's next (Phase 2)

Wire the core engine into a pgrx PostgreSQL extension with SPI loading from AGE tables, GUC configuration, and Option 0 deployment into the running container.

## Test plan

- [x] `cargo test` — 13 unit tests (chains, stars, cycles, disconnected graphs, depth limits, undirected traversal, app ID resolution)
- [x] `cargo build --release` — compiles clean
- [x] `graph-accel-bench all 5000000` — all 6 topologies generate and benchmark at target scale
- [ ] Review ADR-201 benchmark data and projected speedup tables

### DIFF
--- a/graph-accel/bench/src/main.rs
+++ b/graph-accel/bench/src/main.rs
@@ -1,4 +1,5 @@
 use graph_accel_core::Graph;
+use std::collections::VecDeque;
 use std::time::Instant;
 
 fn main() {
@@ -354,9 +355,11 @@ fn gen_dla(node_count: u64) -> Graph {
 
     graph.add_node(0, "Seed".into(), Some("c_0".into()));
 
-    // Track active "surface" nodes — recent additions that new particles attach to
-    // This keeps the growth at the frontier, like real DLA
-    let mut surface: Vec<u64> = vec![0];
+    // Track active "surface" nodes — recent additions that new particles attach to.
+    // This keeps the growth at the frontier, like real DLA.
+    // VecDeque for O(1) pop_front when evicting oldest surface nodes.
+    let mut surface: VecDeque<u64> = VecDeque::with_capacity(10001);
+    surface.push_back(0);
     let surface_max = 10000usize;
 
     for new_node in 1..node_count {
@@ -377,9 +380,9 @@ fn gen_dla(node_count: u64) -> Graph {
         }
 
         // Maintain surface: add new node, evict oldest if over limit
-        surface.push(new_node);
+        surface.push_back(new_node);
         if surface.len() > surface_max {
-            surface.remove(0);
+            surface.pop_front();
         }
     }
 
--- a/graph-accel/core/src/graph.rs
+++ b/graph-accel/core/src/graph.rs
@@ -6,6 +6,9 @@ pub type NodeId = u64;
 /// Interned relationship type index (avoids storing duplicate strings per edge).
 pub type RelTypeId = u16;
 
+/// Maximum number of distinct relationship types (u16::MAX).
+pub const MAX_REL_TYPES: usize = u16::MAX as usize;
+
 /// Metadata about a node.
 #[derive(Debug, Clone)]
 pub struct NodeInfo {
@@ -20,6 +23,19 @@ pub struct Edge {
     pub rel_type: RelTypeId,
 }
 
+/// A record describing an edge to load into the graph.
+/// Named fields prevent the field-ordering bugs that tuples invite.
+#[derive(Debug, Clone)]
+pub struct EdgeRecord {
+    pub from_id: NodeId,
+    pub to_id: NodeId,
+    pub rel_type: String,
+    pub from_label: String,
+    pub to_label: String,
+    pub from_app_id: Option<String>,
+    pub to_app_id: Option<String>,
+}
+
 /// In-memory graph: adjacency lists + node metadata + relationship type interning.
 ///
 /// Edges are stored bidirectionally — `outgoing[a]` contains edges from a,
@@ -31,6 +47,8 @@ pub struct Graph {
     app_id_index: HashMap<String, NodeId>,
     rel_types: Vec<String>,
     rel_type_map: HashMap<String, RelTypeId>,
+    /// Hint for Vec pre-allocation in add_edge. Set by with_capacity().
+    estimated_avg_degree: usize,
 }
 
 impl Graph {
@@ -42,10 +60,14 @@ impl Graph {
             app_id_index: HashMap::new(),
             rel_types: Vec::new(),
             rel_type_map: HashMap::new(),
+            estimated_avg_degree: 4,
         }
     }
 
     /// Pre-allocate for a known graph size.
+    ///
+    /// `edge_count` is used to estimate average edges per node for Vec pre-allocation
+    /// in `add_edge`, avoiding repeated re-allocation during bulk loading.
     pub fn with_capacity(node_count: usize, edge_count: usize) -> Self {
         Self {
             outgoing: HashMap::with_capacity(node_count),
@@ -54,14 +76,29 @@ impl Graph {
             app_id_index: HashMap::with_capacity(node_count),
             rel_types: Vec::new(),
             rel_type_map: HashMap::new(),
+            estimated_avg_degree: if node_count > 0 {
+                (edge_count / node_count).max(1)
+            } else {
+                4
+            },
         }
     }
 
     /// Intern a relationship type string, returning its compact ID.
+    ///
+    /// # Panics
+    /// Panics if more than 65,535 distinct relationship types are interned.
+    /// In the pgrx extension this panic is caught by `#[pg_guard]` and
+    /// converted to a Postgres ERROR.
     pub fn intern_rel_type(&mut self, rel_type: &str) -> RelTypeId {
         if let Some(&id) = self.rel_type_map.get(rel_type) {
             return id;
         }
+        assert!(
+            self.rel_types.len() < MAX_REL_TYPES,
+            "graph_accel: exceeded maximum of {} distinct relationship types",
+            MAX_REL_TYPES
+        );
         let id = self.rel_types.len() as RelTypeId;
         self.rel_types.push(rel_type.to_string());
         self.rel_type_map.insert(rel_type.to_string(), id);
@@ -69,8 +106,9 @@ impl Graph {
     }
 
     /// Resolve a RelTypeId back to its string name.
-    pub fn rel_type_name(&self, id: RelTypeId) -> &str {
-        &self.rel_types[id as usize]
+    /// Returns None if the ID is out of range.
+    pub fn rel_type_name(&self, id: RelTypeId) -> Option<&str> {
+        self.rel_types.get(id as usize).map(|s| s.as_str())
     }
 
     /// Register a node with metadata.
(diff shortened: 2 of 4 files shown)


aaronsb:
## Code Review -- graph-accel Phase 1

**Scope**: 915 lines across 4 Rust files. Pure data structure + traversal engine with benchmark harness. All 13 tests pass. Clippy reports 4 warnings (3 in core, 1 in bench). No Postgres dependencies yet.

**Overall assessment**: Solid foundation. The adjacency list design is clean and well-suited for the pgrx wrapping in Phase 2. The BFS implementations are correct. I found one real bug in `shortest_path`, several performance concerns that will compound at 50M edges, and a handful of API design issues worth addressing before the pgrx layer locks in the interface.

---

### BUG: `shortest_path` reconstruct_path may see incomplete visited map

**Location**: `core/src/traversal.rs:119-123`

```rust
if !visited.contains_key(&edge.target) {
    visited.insert(edge.target, (current, edge.rel_type));

    if edge.target == target {
        return Some(reconstruct_path(graph, &visited, start, target));
    }
```

When the target is found, `reconstruct_path` walks backward through `visited` from `target` to `start`. The target node was *just* inserted, so its parent chain back to `start` is complete -- each node in that chain was visited at an earlier BFS layer, so their entries exist. **This is actually correct for BFS** because BFS explores layer-by-layer and the parent chain is always populated before deeper layers are reached.

However: this relies on `visited.insert(edge.target, ...)` happening *before* the `reconstruct_path` call on line 123. The clippy suggestion to use `Entry::Vacant` would change this -- after `e.insert(...)`, calling `reconstruct_path` borrows `&visited` while `e` still holds a mutable borrow. The compiler would reject it. If someone naively restructures to satisfy clippy here, they could introduce a subtle issue. Worth a `#[allow(clippy::map_entry)]` with a comment explaining why, or restructure the early-exit to happen after the `if` block.

**Verdict**: Not a bug today, but a correctness trap waiting to happen during refactoring.

---

### PERFORMANCE: Vec cloning in BFS hot loop is O(depth * edges_visited)

**Location**: `core/src/traversal.rs:37-54`

```rust
let mut visited: HashMap<NodeId, (u32, Vec<RelTypeId>)> = HashMap::new();
let mut queue: VecDeque<(NodeId, u32, Vec<RelTypeId>)> = VecDeque::new();
// ...
let mut new_path = path_types.clone();  // <-- clone per edge
new_path.push(edge.rel_type);
visited.insert(edge.target, (depth + 1, new_path.clone()));  // <-- second clone
queue.push_back((edge.target, depth + 1, new_path));
```

Every node visited clones and stores a `Vec<RelTypeId>` of length proportional to its depth. At 5M nodes with average depth 10, that is 5M vectors averaging 5 elements each = ~25M heap allocations just for path tracking. The `new_path` is cloned *twice* per node -- once into `visited` and once into `queue`.

**Why it matters**: On the scale-free benchmark at 50M edges, this is likely a significant fraction of the 15s BFS time. The allocator pressure will also cause cache thrashing.

**Suggestion**: Store only the parent pointer (like `shortest_path` does) and reconstruct paths lazily in the result-building phase. This eliminates all Vec allocations from the hot loop:

```rust
// In the BFS loop, store only parent + edge type
visited: HashMap<NodeId, (u32, NodeId, RelTypeId)>  // (depth, parent, edge_type)

// Reconstruct path_types only when building NeighborResult
fn trace_path(visited: &HashMap<...>, start: NodeId, target: NodeId) -> Vec<RelTypeId> { ... }
```

This converts O(N * avg_depth) heap allocations into O(N) entries with fixed-size values.

---

### PERFORMANCE: HashMap vs flat arrays at 50M edges

**Location**: `core/src/graph.rs:27-34`

```rust
pub struct Graph {
    outgoing: HashMap<NodeId, Vec<Edge>>,
    incoming: HashMap<NodeId, Vec<Edge>>,
    nodes: HashMap<NodeId, NodeInfo>,
    ...
}
```

HashMap is the right choice for a generic graph library, but the ADR states AGE's internal IDs are sequential integers. If nodes are assigned IDs 0..N, a `Vec<Vec<Edge>>` (or a CSR representation) would eliminate all hash overhead in the traversal hot path. At 5M nodes, each BFS step does two HashMap lookups (outgoing + incoming) = 10M+ hash probes for a full traversal.

**Why it matters**: The hash overhead is hidden in the benchmark because generation and traversal are measured together, but at 50M edges the constant factor matters. A CSR (Compressed Sparse Row) representation would also improve cache locality dramatically -- all edges for a node are contiguous in memory.

**Suggestion**: This is a Phase 2 consideration. For now, the HashMap is fine and keeps the API flexible. But consider adding a `CompactGraph` or `CsrGraph` alternative that the pgrx loader can populate when it knows the ID space is dense. The current `Graph` trait surface is small enough that swapping the backing store later is feasible.

---

### PERFORMANCE: `surface.remove(0)` in DLA generator is O(n)

**Location**: `bench/src/main.rs:381-382`

```rust
surface.push(new_node);
if surface.len() > surface_max {
    surface.remove(0);  // O(n) shift of entire Vec
}
```

`Vec::remove(0)` shifts all elements left. With `surface_max = 10000`, this is 10K element copies per node after the surface fills. At 5M nodes, that is ~50 billion element copies.

**Suggestion**: Use a `VecDeque` instead, which provides O(1) `push_back` + `pop_front`. This is a benchmark-only issue so it does not affect correctness, but it likely makes the DLA generator 5-10x slower than necessary.

---

### SAFETY: `intern_rel_type` silently truncates at u16::MAX

**Location**: `core/src/graph.rs:61-68`

```rust
pub type RelTypeId = u16;

pub fn intern_rel_type(&mut self, rel_type: &str) -> RelTypeId {
    let id = self.rel_types.len() as RelTypeId;  // wraps at 65536
    ...
}
```

If more than 65,535 unique relationship types are interned, the cast wraps silently and two different strings map to the same ID. In practice, knowledge graphs have far fewer relationship types, but this is the kind of invariant that should be enforced with an assertion or `Result` return.

**Suggestion**: Add a bounds check:
```rust
assert!(self.rel_types.len() < u16::MAX as usize, "exceeded max relationship types (65535)");
```

Or, better for a library, return `Result<RelTypeId, GraphError>` so the pgrx layer can convert it to a Postgres ERROR rather than panicking the backend process. (Even with `#[pg_guard]`, panics have overhead and show less helpful error messages than explicit errors.)

---

### SAFETY: `rel_type_name` will panic on invalid RelTypeId

**Location**: `core/src/graph.rs:72-74`

```rust
pub fn rel_type_name(&self, id: RelTypeId) -> &str {
    &self.rel_types[id as usize]
}
```

Direct index without bounds checking. If a stale `RelTypeId` from a previous graph version is used, this panics. In the pgrx context, `#[pg_guard]` would catch it, but the error message would be opaque ("index out of bounds").

**Suggestion**: Return `Option<&str>` or add a debug_assert.

---

### API DESIGN: `load_edges` tuple has 5 unnamed fields

**Location**: `core/src/graph.rs:98-101`

```rust
pub fn load_edges<I>(&mut self, edges: I)
where
    I: IntoIterator<Item = (NodeId, NodeId, String, String, String)>,
```

And `load_edges_with_app_ids` has a 7-tuple. These are error-prone at the call site -- it is easy to swap `from_label` and `to_label`, or `from_app_id` and `to_app_id`, and the compiler will not catch it since they are all the same type.

**Suggestion**: Define a struct for the edge row:
```rust
pub struct EdgeRow {
    pub from: NodeId,
    pub to: NodeId,
    pub rel_type: String,
    pub from_label: String,
    pub to_label: String,
    pub from_app_id: Option<String>,
    pub to_app_id: Option<String>,
}
```

This also makes the SPI-to-Rust mapping in Phase 2 more readable.

---

### API DESIGN: `memory_usage()` undercounts significantly

**Location**: `core/src/graph.rs:175-184`

```rust
pub fn memory_usage(&self) -> usize {
    let nodes_mem = self.nodes.len() * (size_of::<NodeId>() + size_of::<NodeInfo>() + 40);
    let out_edges: usize = self.outgoing.values().map(|v| v.len() * size_of::<Edge>()).sum();
    let in_edges: usize = self.incoming.values().map(|v| v.len() * size_of::<Edge>()).sum();
    let index_mem = self.app_id_index.len() * 80;
    nodes_mem + out_edges + in_edges + index_mem
}
```

Missing:
1. HashMap bucket overhead (each HashMap has ~1.125x capacity overhead for robin hood hashing)
2. Vec capacity overhead (Vecs often have 2x their len in capacity after growth)
3. String heap allocations inside `NodeInfo` (label, app_id) -- the magic `40` probably attempts to cover this but `label` length varies
4. The `rel_types` and `rel_type_map` fields entirely
5. The `outgoing`/`incoming` HashMap entry overhead (~56 bytes per entry for the hashmap itself, not just the values)

The benchmark output shows "~992MB" for 5M nodes with 5M edges. The actual RSS is likely 1.5-2x higher. This matters because `max_memory_mb` in Phase 2 needs accurate accounting to prevent OOM.

**Suggestion**: Either make the estimate more accurate by accounting for HashMap overhead and Vec capacity, or rename the function to `estimated_memory_usage_bytes()` and document that it undercounts by ~30-50%. The pgrx layer should use the allocator's actual tracked size rather than this estimate.

---

### COMPILER WARNING: `edge_count` parameter unused in `with_capacity`

**Location**: `core/src/graph.rs:49`

```rust
pub fn with_capacity(node_count: usize, edge_count: usize) -> Self {
```

`edge_count` is accepted but never used to pre-allocate the `outgoing`/`incoming` HashMaps' inner Vecs. The HashMap capacity is set to `node_count` (the number of *buckets*), but individual `Vec<Edge>` entries are not pre-allocated. This means every `add_edge` call triggers Vec growth with reallocations.

**Suggestion**: Either use `edge_count` to estimate average degree and pre-allocate inner Vecs during `load_edges`, or rename the parameter to `_edge_count` until it is actually used. The current state is misleading -- callers think they are getting pre-allocation but they are not.

---

### TEST COVERAGE: What is missing

The 13 tests cover the happy paths well. Missing coverage:

1. **Empty graph**: `bfs_neighborhood` and `shortest_path` on a graph with zero nodes and zero edges. The `start` node would not exist in the graph -- does this produce an empty result or panic?

2. **Start node not in graph**: What happens when `bfs_neighborhood(graph, 999, 5)` is called with a node ID that was never added? Currently it would return `nodes_visited: 1` and an empty `neighbors` list (since `visited` starts with the start node, and `neighbors_all` returns empty slices for unknown nodes). The `filter(|(id, _)| *id != start)` removes it from results. This works but should be tested to lock in the behavior.

3. **Self-loops**: `add_edge(1, 1, rt)` -- does BFS handle a node that is its own neighbor? Currently `neighbors_all` would yield the self-loop edge, but `visited.contains_key` would prevent re-visiting. So it works, but worth a test.

4. **Parallel edges**: Multiple edges between the same pair of nodes with different rel_types. BFS would visit the target only once (first edge wins). Is that the intended behavior? The `path_types` would record only the first edge's type. Should be documented and tested.

5. **`shortest_path` with max_hops=0**: Returns `None` unless `start == target`. Worth a test.

6. **`intern_rel_type` idempotency**: Calling it twice with the same string should return the same ID. Tested implicitly but worth an explicit test since correctness depends on it.

7. **Large fan-out**: A node with 100K+ edges (realistic for hub nodes in scale-free graphs). The benchmark exercises this but there is no unit test asserting correctness at that scale.

---

### MINOR: Clippy warnings should be addressed

3 `map_entry` warnings in `traversal.rs` and 1 `type_complexity` warning in `bench/src/main.rs`. For the `map_entry` in `shortest_path`, use `#[allow]` with a comment as noted above. The others can be mechanically fixed.

---

### POSITIVE OBSERVATIONS

Things that are well done:

- **Dual adjacency storage** (outgoing + incoming) is the right call for undirected BFS. Avoids a full-graph scan to find incoming edges.
- **Relationship type interning** reduces per-edge memory from ~40 bytes (String) to 2 bytes (u16). At 50M edges that is ~1.9GB savings.
- **The `Edge` struct is `Copy`** (4 bytes target + 2 bytes rel_type = 10 bytes with padding). This is cache-friendly and avoids heap allocations per edge.
- **The benchmark topology selection is excellent**. Six generators that exercise different pathological cases. The barbell graph in particular is a great stress test for bottleneck pathfinding.
- **Clean module separation**. `graph.rs` is data structure only, `traversal.rs` is algorithms only. This will map directly to separate files in the pgrx extension.
- **The ADR is thorough** -- prior art, deployment options, configuration, safety model, and honest performance analysis.

---

### Summary

| Category | Finding | Severity |
|----------|---------|----------|
| Correctness | `reconstruct_path` clippy trap | Low (works today, document it) |
| Performance | Vec cloning in BFS hot loop | **High** (dominates large traversals) |
| Performance | HashMap vs dense array | Medium (Phase 2 consideration) |
| Performance | `surface.remove(0)` in DLA | Low (bench only) |
| Safety | `intern_rel_type` u16 overflow | Medium (silent corruption) |
| Safety | `rel_type_name` panic on bad ID | Low (pg_guard catches it) |
| API | 5/7-tuple load functions | Medium (error-prone interface) |
| API | `memory_usage` undercounts | Medium (affects Phase 2 memory cap) |
| Build | Unused `edge_count` parameter | Low (misleading API) |
| Tests | Missing edge cases | Medium (6 categories listed above) |

**Recommendation**: Address the Vec cloning performance issue and the u16 overflow safety issue before Phase 2. The rest can be addressed incrementally. The architecture and module structure are sound for wrapping with pgrx.

PR #264, opened 2026-02-01
feat(graph-accel): pgrx extension shell — Phase 2 of ADR-201

aaronsb:
## Summary

Adds `graph-accel/ext/` — a pgrx 0.16.1 PostgreSQL extension that wraps the core in-memory traversal engine (Phase 1) and exposes it as SQL functions against Apache AGE graphs.

- **4 SQL functions**: `graph_accel_status()`, `graph_accel_load()`, `graph_accel_neighborhood()`, `graph_accel_path()`
- **7 GUC parameters**: source_graph, max_memory_mb, node_id_property, node_labels, edge_types, auto_reload, reload_debounce_sec
- **Per-backend state**: `thread_local! + RefCell<Option<Graph>>` — shared memory deferred to Phase 3
- **SPI loading**: reads `ag_catalog.ag_label` for label discovery, loads per-table with filtering, parses agtype properties via serde_json
- **Dual node resolution**: tries app_id (concept_id) first, falls back to AGE graphid

## Key design decisions

- Per-backend state instead of shared memory for Phase 2 simplicity
- `max_memory_mb` GUC uses Sighup context (changes to Postmaster in Phase 3 with shared memory)
- Per-label-table SPI queries avoid inheritance scans and give relationship type for free
- No path found = empty result set (SQL convention), not an error

## Verified

- `cargo check` — all 3 workspace crates compile
- `cargo test -p graph-accel-core` — 26 tests pass
- `cargo pgrx run pg17` — extension loads into PG 17.7, status returns "not_loaded", all GUCs visible with correct defaults

## Test plan

- [ ] Option 0 deployment: copy .so into live `knowledge-graph-postgres` container
- [ ] `CREATE EXTENSION graph_accel` against live database
- [ ] `graph_accel_load('knowledge_graph')` — verify SPI reads AGE tables
- [ ] `graph_accel_neighborhood()` and `graph_accel_path()` against real concepts

### DIFF
--- a/graph-accel/ext/src/load.rs
+++ b/graph-accel/ext/src/load.rs
@@ -2,6 +2,7 @@ use std::time::Instant;
 
 use graph_accel_core::Graph;
 use pgrx::prelude::*;
+use pgrx::spi::{quote_identifier, quote_literal};
 
 use crate::guc;
 use crate::state::{self, GraphState};
@@ -26,13 +27,15 @@ fn graph_accel_load(
             error!("graph_accel: source_graph not set and no graph_name argument provided");
         });
 
+    validate_name(&gname);
+
     let (node_count, edge_count) = Spi::connect(|client| {
         // Verify graph exists
         let exists = client
             .select(
                 &format!(
-                    "SELECT 1 FROM ag_catalog.ag_graph WHERE name = '{}'",
-                    sanitize_ident(&gname)
+                    "SELECT 1 FROM ag_catalog.ag_graph WHERE name = {}",
+                    quote_literal(&gname)
                 ),
                 None,
                 &[],
@@ -127,9 +130,9 @@ fn load_label_catalog(
         "SELECT l.name, l.kind::text \
          FROM ag_catalog.ag_label l \
          JOIN ag_catalog.ag_graph g ON l.graph = g.graphid \
-         WHERE g.name = '{}' \
+         WHERE g.name = {} \
            AND l.name NOT LIKE '_ag%%'",
-        sanitize_ident(graph_name)
+        quote_literal(graph_name)
     );
 
     let mut labels = Vec::new();
@@ -162,9 +165,9 @@ fn load_vertices(
     graph: &mut Graph,
 ) -> Result<(), pgrx::spi::SpiError> {
     let query = format!(
-        "SELECT id::text, properties::text FROM {}.\"{}\"",
-        sanitize_ident(graph_name),
-        sanitize_ident(label_name)
+        "SELECT id::text, properties::text FROM {}.{}",
+        quote_identifier(graph_name),
+        quote_identifier(label_name)
     );
 
     let table = client.select(&query, None, &[])?;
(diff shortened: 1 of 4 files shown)


aaronsb:
## Second Review Pass — Fix Commit `061c7248`

Reviewed the three fixes from the first pass: SQL injection remediation, `assert!()` replacement, and `resolve_node` deduplication.

### 1. SQL Injection Fix — Verified Clean

All four `format!()` calls in `load.rs` now use the correct pgrx quoting function for their position:

| Location | Position | Function | Correct? |
|---|---|---|---|
| `load.rs:36-38` — graph exists check | WHERE value | `quote_literal()` | Yes |
| `load.rs:129-135` — label catalog | WHERE value | `quote_literal()` | Yes |
| `load.rs:167-171` — vertex query | FROM schema.table | `quote_identifier()` | Yes |
| `load.rs:208-211` — edge query | FROM schema.table | `quote_identifier()` | Yes |

The old `sanitize_ident()` function is **fully removed** — confirmed via grep, zero occurrences remain. No remaining hand-rolled quoting anywhere in the extension.

The `validate_name()` at line 30 runs **before** entering the SPI block, which is the correct placement — fail fast before any database interaction. The `quote_literal()`/`quote_identifier()` calls inside SPI provide belt-and-suspenders defense. This is sound defense in depth.

**Note on label names from catalog:** Label names at lines 72 and 83 flow from `load_label_catalog()` (database-sourced via `ag_catalog.ag_label`) directly into `load_vertices()`/`load_edges()` where they hit `quote_identifier()`. These are trusted values from the system catalog, and `quote_identifier()` handles them correctly regardless. No issue here.

### 2. Error Handling Fix — Verified

`validate_name()` (load.rs:253-256) now uses `pgrx::error!()` instead of `assert!()`. This is correct — `error!()` raises a PostgreSQL ERROR that the client can catch and handle, while `assert!()` would `panic!()` the process and crash the backend. Confirmed zero remaining `assert!()` or `panic!()` calls anywhere in `ext/src/` (the one grep hit is in the doc comment on line 252 which just explains the change).

### 3. Dedup of `resolve_node` — Verified Identical

The new `state::resolve_node()` (state.rs:40-52) is **character-for-character identical** to the two copies that were removed from `neighborhood.rs` and `path.rs`, with one correct difference: it uses the fully qualified `pgrx::error!()` instead of the bare `error!()` since `state.rs` does not have `use pgrx::prelude::*`. Both callers now invoke `state::resolve_node(&gs.graph, ...)`.

**On the RefCell/error! interaction:** `resolve_node()` is called from within `with_graph()` closures (neighborhood.rs:20, path.rs:21-22) which hold a `RefCell::borrow()`. If `pgrx::error!()` fires, it performs a Postgres `longjmp` which unwinds the Rust stack. The `RefCell` borrow guard gets dropped during this unwind, so there is no borrow leak. This is safe.

### 4. Build Verification

- `cargo check` passes cleanly (only the expected `dead_code` warning for `load_time_ms`/`loaded_at` fields reserved for Phase 3 staleness tracking)
- `cargo test -p graph-accel-core` — all 26 tests pass
- No regressions introduced

### 5. Remaining Items from First Pass (Re-verified)

- **guc.rs** — All 7 GUCs properly registered with `Sighup` context, correct defaults and ranges. `get_string()` helper handles None/empty correctly. No changes needed.
- **status.rs** — Clean use of `with_graph()`, proper fallback for unloaded state. No changes needed.
- **lib.rs** — `_PG_init()` calls `register_gucs()`, `pg_module_magic!()` present, two `pg_test` functions cover status and GUC defaults. No changes needed.

### Assessment

All three findings from the first review are correctly addressed. No new issues introduced. The code is clean for merge, pending the Option 0 deployment test (running against a live container with actual AGE data) documented in the PR test plan.

PR #265, opened 2026-02-01
fix(graph-accel): Option 0 deployment + benchmark findings

aaronsb:
## Summary

- Fix two bugs discovered during live container deployment: `ag_label.name` type cast (`name` → `text`) and GUC contexts (Sighup → Userset for Phase 2 per-backend state)
- Add benchmark findings document tracing the full query chain from web UI through API to Cypher, with measured AGE baselines and graph_accel results
- Add automated comparison test script validating timing and data correctness at depths 1-5

## Results (236 concepts, 2159 edges)

| Depth | AGE | graph_accel | Speedup |
|-------|-----|-------------|---------|
| 1 | 3,607ms | 0.181ms | 19,926x |
| 2 | 508ms | 0.091ms | 5,577x |
| 3 | 1,861ms | 0.144ms | 12,924x |
| 4 | 11,569ms | 0.309ms | 37,440x |
| 5 | 93,174ms | 0.423ms | **220,270x** |
| 6+ | Hangs | 0.377ms | ∞ |

Data correctness: depth-1 exact match (11/11 concepts), depth 2+ strict superset (all AGE concepts present + extras via cross-type paths).

## Test plan

- [x] Extension loads in live container without segfault
- [x] `graph_accel_load('knowledge_graph')` loads 788 nodes / 2,159 edges in 22ms
- [x] `benchmark-comparison.sh` passes all checks (timing + data correctness)
- [x] Depth 1 concept sets identical between AGE and graph_accel

### DIFF
--- a/graph-accel/tests/benchmark-comparison.sh
+++ b/graph-accel/tests/benchmark-comparison.sh
@@ -7,6 +7,13 @@
 # Requirements:
 #   - knowledge-graph-postgres container running with AGE graph loaded
 #   - graph_accel extension installed (Option 0 or later deployment)
+#   - PostgreSQL 17.x (psql output parsing assumes 17.x formatting)
+#   - bc (for floating-point speedup calculation)
+#
+# Notes:
+#   - Operator tooling — concept_id is interpolated into Cypher strings.
+#     Not suitable for untrusted input without parameterization.
+#   - Concept ID grep filter assumes sha256: or c_ prefixes (current dataset).
 #
 # Usage:
 #   ./graph-accel/tests/benchmark-comparison.sh


aaronsb:
## Review: Option 0 Deployment Fixes + Benchmark Findings

505 additions, 14 deletions across 6 files. Two bug fixes from live deployment, one benchmark doc, one test script, ADR update, tracking file update.

---

### 1. GUC Context Change: Sighup to Userset (`guc.rs`)

**Assessment: Appropriate for Phase 2, with a clear caveat documented**

The comment on line 32-33 does the right thing: it explains *why* Userset now, and *when* it tightens:

```rust
// Phase 2: Userset context allows per-session SET for easy testing.
// Phase 3 (shared memory) will tighten to Sighup/Postmaster where appropriate.
```

This is safe in Phase 2 because the graph state is per-backend (`thread_local!` in `state.rs`). Any user who runs `SET graph_accel.max_memory_mb = 131072` only affects their own backend's allocation cap -- they cannot exhaust shared memory or impact other connections. The `MAX_MEMORY_MB` default is 4096 and the hard ceiling is 131072 (128GB), which is checked at load time, not at SET time.

**One thing to keep in mind for Phase 3**: When state moves to shared memory, `max_memory_mb` *must* become `GucContext::Postmaster` (requires restart to change). Similarly, `source_graph` and `auto_reload` should become `Sighup` since they would affect the shared structure. The existing comment captures this intent, which is good. Just flagging it so it does not get lost during the Phase 3 transition.

**Security assessment**: No escalation risk. Userset means any connected role can SET these, but since the graph is per-backend and read-only, the worst a user can do is waste their own backend's memory or load a different AGE graph into their own session. There is no write path and no cross-backend effect.

---

### 2. Type Cast Fix: `l.name::text` (`load.rs:130`)

**Assessment: Correct fix**

The Postgres `name` type (Oid 19) is a fixed-length 63-byte type used internally for identifiers. pgrx's SPI `get_by_name::<String>()` expects `text` (Oid 25). Without the cast, pgrx would fail to deserialize the column because the Oid does not match its expected type mapping.

The cast `l.name::text` is the standard Postgres approach -- it is a no-cost cast (binary compatible, the planner does not generate a function call). The `l.kind::text` cast was already present, so this makes the two columns consistent. The `WHERE g.name = {}` comparison uses `quote_literal` which produces a text literal, so the implicit cast on the comparison side worked fine -- only the SELECT output column needed the explicit cast for pgrx deserialization.

No concerns here.

---

### 3. Benchmark Comparison Script (`benchmark-comparison.sh`)

**Overall assessment: Functional test script with good structure. A few items to note.**

**What works well:**
- `set -euo pipefail` -- strict mode from the start
- Prerequisite checks (concept exists, extension installed) before running the expensive parts
- Data correctness validation at depth 1 with exact diff
- Superset check at higher depths (GA >= AGE) is the right invariant
- Clean exit codes: 0 for pass, 1 for fail
- Accepts a custom concept ID as `$1`

**Items to be aware of:**

**(a) CONCEPT_ID flows into Cypher strings without parameterization (lines 48-49, 87, 122-126, 197-198, 204-207)**

The `$CONCEPT_ID` variable is interpolated directly into Cypher dollar-quoted strings:

```bash
MATCH (c:Concept {concept_id: '$CONCEPT_ID'}) RETURN c.label
```

This is a test script running inside a controlled container via `docker exec`, not an API endpoint, so the practical risk is low. But if someone passes a crafted concept ID like `'}) RETURN 1 as label //$`, it would break out of the Cypher string context. Since this is operator tooling (not user-facing), this is acceptable as-is. If the script is ever promoted to CI or exposed more broadly, parameterization through psql variables (`-v concept_id=...`) would be the safer approach.

**(b) Output parsing is fragile (lines 98-99, 130, 138-144)**

The grep/awk chains to extract counts and timing from psql output depend on specific output formatting:

```bash
count=$(echo "$output" | grep -E '^\s*[0-9]+' | tr -d ' ')
time_ms=$(echo "$output" | grep "^Time:" | tail -1 | grep -oP '[\d.]+' | head -1)
```

This works for the current psql version and locale, but psql output formatting can vary between versions (e.g., column alignment, header presence). The `-t -A` flags in `psql_cmd` suppress headers and alignment, which helps. The `psql_timed` function does not use these flags since it needs `\timing` output, which is inherently human-formatted.

This is a known trade-off in psql-based benchmarking scripts. Documenting the expected psql version (17.x) in the script header would help future maintainers.

**(c) The `ga_d1` grep pattern assumes concept ID prefixes (line 208)**

```bash
| grep '^sha256:\|^c_' | sort)
```

This filters graph_accel output to lines starting with `sha256:` or `c_`. If concept IDs ever use a different prefix (e.g., `uuid:`, or plain strings), this filter would silently drop results, causing a false mismatch. Since this is a test for a specific dataset where the ID format is known, it is fine. Just a note for generalization.

**(d) `bc -l` dependency (line 172)**

The script uses `bc` for floating-point arithmetic. This is available on Debian/Ubuntu but not always in minimal containers. Since the script runs on the host (via `docker exec`), not inside the container, this should be fine on any standard workstation.

---

### 4. Benchmark Findings Document (`benchmark-findings.md`)

**Assessment: Thorough and well-structured. Good companion to the ADR.**

Positive observations:
- The full query chain trace (Web UI through API to Cypher) is valuable for understanding what the extension is actually replacing
- Explicitly documents the depth-1 anomaly (3.6s for 1-hop, faster than depth 2) with a plausible explanation (UNION ALL across 120 edge tables)
- Clearly explains the node count discrepancy between AGE and graph_accel (Concept-only vs all types)
- Data correctness section documents exact match at depth 1 and explains the superset behavior at depth 2+
- Reproduction SQL included in full

**One documentation nit:**

The PR body's results table shows different numbers than the benchmark-findings.md:

| Source | Depth 1 time | Depth 2 time |
|--------|-------------|-------------|
| PR body | 3,607ms | 508ms |
| Findings doc | 3,644ms | 471ms |
| ADR update | 3,644ms | 471ms |

The PR body numbers appear to be from a different run. This is fine -- benchmark variance across runs is expected -- but worth noting that the ADR and findings doc are consistent with each other, which is what matters for the committed record. The PR body is ephemeral.

---

### 5. ADR-201 Update

**Assessment: Clean addition, consistent with findings doc**

The new SQL-measured timing table slots in naturally after the existing API-measured table. The cross-reference to `benchmark-findings.md` ties the two documents together. The label for the existing table ("API-measured timing") adds clarity that was previously implicit.

Numbers match between the ADR and the findings doc. Good.

---

### 6. Tracking File (`.claude/todo-adr-201-graph-accel.md`)

**Assessment: Accurate state update**

- Phase 2 Option 0 deployment marked complete with bullet-point results
- Phase 4 items that were implemented as part of Phase 2 (node_labels, edge_types, node_id_property) correctly moved to `[x]`
- New TODO added: "Edge filtering: skip edges where source/target not in loaded node set" -- this captures the cross-type traversal finding from the benchmark, which is a good example of discovery-driven planning
- Phase 5 benchmark item marked complete with cross-reference

---

### Summary

The two code changes are correct and well-motivated: the type cast fixes a real pgrx SPI deserialization issue, and the GUC context change is appropriate for Phase 2's per-backend architecture with a clear plan to tighten in Phase 3.

The benchmark documentation is thorough and honest about methodology differences (Concept-only vs all-types traversal, superset behavior). The test script is functional for its purpose as operator tooling.

No blocking issues. The Phase 3 GUC tightening is the main thing to track going forward.

PR #285, opened 2026-02-03
feat: unified smart search with exploration tracking and query program

aaronsb:
## Summary

Replaces the three separate search tabs (concept, neighborhood, path) with a unified progressive interface. Introduces exploration tracking that records graph actions as ordered Cypher statements, enabling save/load/replay of graph explorations. Establishes a unified query program IR (`{ op: '+' | '-', cypher: string }[]`) that all query modes compile to.

### Search UX
- Single progressive search flow: select concept -> adjust depth -> optionally add destination
- Parameter-presence model: mode is derived from which params are set, not stored
- Additive graph loading: "Load" replaces, "Add to Graph" merges
- Warmth physics slider for force graph tuning

### Exploration Tracking (Phase 1)
- Every graph action (explore, follow, add-adjacent, remove, load-path) records a step
- Steps include semantic metadata and equivalent Cypher statements
- `+/-` set algebra operators for additive/subtractive graph operations
- Undo last step support

### Saved Queries (Phase 2)
- Save/load/delete explorations via query definitions API
- Exploration type stores `{ op, cypher }[]` statements
- Replays Cypher sequentially with correct AGE ID mapping
- Saved queries panel in icon rail sidebar

### Unified Query Program
- Cypher editor now participates in the exploration session
- Multi-statement `+/-` script execution (was single statement only)
- Fixed: Cypher editor bypassed rawGraphData pipeline, used wrong IDs
- Export to Editor: push exploration session as editable Cypher text
- Round-trip: smart search -> save -> load -> export to editor -> edit -> execute -> save

### Canvas Context Menu
- Travel submenu: find path between origin/destination, animate camera through nodes
- Send to Polarity Explorer: sets origin/destination as poles
- Send Path to Reports: creates traversal report with per-path tables

### Bug Fixes
- Cypher endpoint misidentified edges as nodes (AGE `start_id`/`end_id` check)
- Query definition create response missing `definition` field
- Stale URL params replayed on navigation (clear on mount without params)
- getSubgraph missing relationship target hydration

## Test plan
- [ ] Smart search: select concept, adjust depth, load graph
- [ ] Add destination: path mode activates, find paths, select, load
- [ ] Add to Graph: merges new data without replacing
- [ ] Remove from Graph: right-click node, remove, graph updates
- [ ] Save exploration: build graph, save, verify in saved queries list
- [ ] Load exploration: click saved query, graph rebuilds with correct edges
- [ ] Cypher editor: write `+ MATCH ...;` script, execute, graph renders
- [ ] Cypher editor multi-statement: use `+/-` operators, verify set algebra
- [ ] Export to Editor: build via smart search, click code icon, editor populated
- [ ] Round-trip: save -> load -> export to editor -> edit -> execute -> save again
- [ ] Clear graph: click eraser, graph clears, load saved query works
- [ ] Travel: set origin+destination, right-click canvas, Travel > From Origin
- [ ] Send to Polarity: right-click canvas, Send to Polarity Explorer
- [ ] Send Path to Reports: right-click canvas, creates traversal report
- [ ] Traversal report: CSV and Markdown export work

### DIFF
--- a/web/src/components/shared/SearchBar.tsx
+++ b/web/src/components/shared/SearchBar.tsx
@@ -23,7 +23,8 @@ import { ModeDial } from './ModeDial';
 import { apiClient } from '../../api/client';
 import { BlockBuilder } from '../blocks/BlockBuilder';
 import { stepToCypher, parseCypherStatements } from '../../utils/cypherGenerator';
-import { mapCypherResultToRawGraph } from '../../utils/cypherResultMapper';
+import { mapCypherResultToRawGraph, extractGraphFromPath } from '../../utils/cypherResultMapper';
+import type { PathResult } from '../../utils/cypherResultMapper';
 import {
   ConceptSearchInput,
   SliderControl,
@@ -55,18 +56,18 @@ export const SearchBar: React.FC = () => {
   // === UNIFIED SEARCH STATE ===
   // Primary concept (the starting point for any search)
   const [primaryQuery, setPrimaryQuery] = useState('');
-  const [selectedPrimary, setSelectedPrimary] = useState<any>(null);
+  const [selectedPrimary, setSelectedPrimary] = useState<{ concept_id: string; label: string } | null>(null);
   const [depth, setDepth] = useState(1);
 
   // Destination concept (optional — triggers path mode)
   const [destinationQuery, setDestinationQuery] = useState('');
-  const [selectedDestination, setSelectedDestination] = useState<any>(null);
+  const [selectedDestination, setSelectedDestination] = useState<{ concept_id: string; label: string } | null>(null);
   const [showDestination, setShowDestination] = useState(false);
   const [maxHops, setMaxHops] = useState(5);
 
   // Path search state
-  const [pathResults, setPathResults] = useState<any>(null);
-  const [selectedPath, setSelectedPath] = useState<any>(null);
+  const [pathResults, setPathResults] = useState<{ paths: PathResult[]; count: number; error?: string } | null>(null);
+  const [selectedPath, setSelectedPath] = useState<PathResult | null>(null);
   const [isLoadingPath, setIsLoadingPath] = useState(false);
 
   // Cypher editor state
@@ -128,12 +129,12 @@ LIMIT 50`);
 
   // === HANDLERS ===
 
-  const handleSelectPrimary = (concept: any) => {
+  const handleSelectPrimary = (concept: { concept_id: string; label: string }) => {
     setSelectedPrimary(concept);
     setPrimaryQuery(concept.label);
   };
 
-  const handleSelectDestination = (concept: any) => {
+  const handleSelectDestination = (concept: { concept_id: string; label: string }) => {
     setSelectedDestination(concept);
     setDestinationQuery(concept.label);
   };
@@ -190,16 +191,17 @@ LIMIT 50`);
         max_hops: maxHops,
       });
       setPathResults(result);
-    } catch (error: any) {
+    } catch (error: unknown) {
       console.error('Failed to find paths:', error);
 
+      const err = error as { code?: string; response?: { data?: { detail?: string } }; message?: string };
       let errorMessage = 'Failed to find paths';
-      if (error.code === 'ECONNABORTED') {
+      if (err.code === 'ECONNABORTED') {
         errorMessage = `Search timed out. Try reducing max hops.`;
-      } else if (error.response?.data?.detail) {
-        errorMessage = error.response.data.detail;
-      } else if (error.message) {
-        errorMessage = error.message;
+      } else if (err.response?.data?.detail) {
+        errorMessage = err.response.data.detail;
+      } else if (err.message) {
+        errorMessage = err.message;
       }
 
       setPathResults({ error: errorMessage, count: 0, paths: [] });
@@ -232,40 +234,7 @@ LIMIT 50`);
       maxHops,
     });
 
-    const conceptNodes: any[] = [];
-    const conceptRelTypes: string[][] = [];
-    let pendingRels: string[] = [];
-
-    for (let i = 0; i < selectedPath.nodes.length; i++) {
-      const node = selectedPath.nodes[i];
-      if (node.id && node.id !== '') {
-        conceptNodes.push(node);
-        conceptRelTypes.push(pendingRels);
-        pendingRels = [];
-      }
-      if (i < selectedPath.relationships.length) {
-        pendingRels.push(selectedPath.relationships[i]);
-      }
-    }
-
-    const nodes = conceptNodes.map((node: any) => ({
-      concept_id: node.id,
-      label: node.label,
-      description: node.description,
-      ontology: 'default',
-      grounding_strength: node.grounding_strength,
-    }));
-
-    const links: any[] = [];
-    for (let i = 0; i < conceptNodes.length - 1; i++) {
-      const rels = conceptRelTypes[i + 1];
-      const relType = rels.find(r => r !== 'APPEARS' && r !== 'SCOPED_BY') || rels[0] || 'CONNECTED';
-      links.push({
-        from_id: conceptNodes[i].id,
-        to_id: conceptNodes[i + 1].id,
-        relationship_type: relType,
-      });
-    }
+    const { nodes, links, conceptNodeIds } = extractGraphFromPath(selectedPath);
 
     if (loadMode === 'clean') {
       setGraphData(null);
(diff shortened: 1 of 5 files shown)


aaronsb:
## Code Review: Unified Smart Search with Exploration Tracking

**Scope**: 30 files, +2535/-1162 lines. Large feature branch with new exploration session system, unified search model, Cypher generator/parser, and significant context menu expansion.

**Overall assessment**: The architecture is sound. The parameter-presence model (`deriveMode`) is a clean improvement over the explicit mode enum, and the exploration session IR (intermediate representation) of `{ op, cypher }[]` that all query modes compile to is a good unifying abstraction. The Cypher generator/parser roundtrip is well-structured. There are a few concrete issues to address.

---

### Bug: Exploration step always records `+` regardless of loadMode

**Location**: `web/src/components/shared/SearchBar.tsx:162`

```typescript
useGraphStore.getState().addExplorationStep({
  action: 'explore',
  op: loadMode === 'add' ? '+' : '+',  // <-- always '+'
  cypher: stepToCypher(stepParams),
  ...
});
```

The ternary `loadMode === 'add' ? '+' : '+'` evaluates to `'+'` in both branches. This appears to be an oversight -- a clean load that replaces the graph should arguably still be `'+'` (it is additive in the query program sense), but the expression reads as an accidental copy-paste. If both branches are intentionally `'+'`, the ternary should be removed and replaced with the literal `'+'` to avoid confusing future readers. If clean loads were intended to get a different operator, the second branch needs the correct value.

---

### Significant duplication: Path node extraction logic appears 3 times

**Locations**:
- `web/src/hooks/useGraphData.ts:195-241` (in `useFindConnection`)
- `web/src/components/shared/SearchBar.tsx:235-268` (in `handleLoadPath`)
- `web/src/explorers/common/useGraphContextMenu.ts:196-226` (in `handleTravelPath`)

All three implement the same algorithm: iterate path nodes, filter empty-ID nodes (Source/Ontology vertices from AGE traversal), collect relationship types between consecutive concept nodes, and build a `{ nodes, links }` payload. The code is nearly identical across all three call sites.

This is a strong extraction candidate. A shared utility like `extractConceptPathFromRawPath(path) => { nodes, links }` would eliminate ~30 lines of duplication per call site and ensure the filtering logic stays consistent if the AGE response format evolves. The `cypherResultMapper.ts` module would be a natural home for this.

---

### Type safety: Pervasive `any` on concept/path types

Several interfaces and handler signatures use `any` where narrower types exist or could be defined:

- `web/src/components/shared/search/ConceptSearchInput.tsx:11-13` -- `results: any[]`, `onSelect: (concept: any) => void`
- `web/src/components/shared/search/PathResults.tsx:5-8` -- `pathResults: any`, `selectedPath: any`, `onSelectPath: (path: any) => void`
- `web/src/components/shared/SearchBar.tsx:58` -- `selectedPrimary: useState<any>(null)`
- `web/src/components/shared/SearchBar.tsx:68-69` -- `pathResults: useState<any>(null)`, `selectedPath: useState<any>(null)`
- `web/src/store/graphStore.ts:150` -- `rawGraphData: { nodes: any[]; links: any[] } | null`

The API response types already exist in `web/src/types/graph.ts` (`APIGraphNode`, `APIGraphLink`). For the search results, the search API returns a known shape. For path results, the `findConnection` response has a definable structure. Narrowing these would catch ID mapping bugs at compile time rather than runtime -- especially relevant given the AGE internal ID vs concept_id theme in this PR.

Not a blocker for this PR, but worth noting: the `rawGraphData` store field using `any[]` for nodes and links means the entire pipeline downstream loses type information. The `APIGraphNode[]` / `APIGraphLink[]` types would be a drop-in improvement.

---

### Race condition risk: Exploration step recorded before API call succeeds

**Locations**:
- `web/src/explorers/common/useGraphContextMenu.ts:82-102` (`handleFollowConcept`)
- `web/src/explorers/common/useGraphContextMenu.ts:107-131` (`handleAddToGraph`)

Both handlers call `store.addExplorationStep(...)` *before* the `apiClient.getSubgraph()` call. If the API call fails (which is caught and shows an alert), the exploration session now contains a step that never produced graph data. On replay, this step would attempt to execute its Cypher equivalent and may or may not succeed depending on why the original call failed.

Consider moving `addExplorationStep` into the success path (after the API response), or adding a rollback mechanism (e.g., removing the last step on catch). The same pattern appears in `handleTravelPath` at line 157, where the step is recorded before the `mergeRawGraphData` call.

For comparison, `SearchBar.tsx:handleLoadExplore` (line 160) does the same pre-recording, but since it delegates to `setSearchParams` -> React Query (which has its own retry/error handling), the failure mode is different.

---

### Missing `graphData: null` on subtract operations

**Location**: `web/src/store/graphStore.ts:324-357` (`subtractRawGraphData`)

When `setRawGraphData` is called, the caller typically also calls `setGraphData(null)` to force re-transformation. But `subtractRawGraphData` only updates `rawGraphData` without clearing `graphData`. The downstream `useEffect` in `ExplorerView` that watches `rawGraphData` and runs `explorerPlugin.dataTransformer` handles re-transformation, so this likely works in practice. However, there is a brief window where `graphData` contains nodes that no longer exist in `rawGraphData`. If any component reads `graphData` synchronously between the store update and the next React render cycle, it could reference stale nodes.

This is a low-severity issue since React's batched updates usually prevent intermediate reads, but it would be more defensive to clear `graphData` inside `subtractRawGraphData` the same way `clearExploration` does.

---

### ExplorerView: `useEffect` dependencies are incomplete

**Location**: `web/src/views/ExplorerView.tsx` (around the data-loading effect)

```typescript
useEffect(() => {
    const newData = mode === 'path' ? pathData : exploreData;
    if (!newData || newData === lastProcessedData.current) return;
    lastProcessedData.current = newData;
    ...
    if (loadModeRef.current === 'add') {
      mergeRawGraphData(graphPayload);
    } else {
      setGraphData(null);
      setRawGraphData(graphPayload);
    }
  }, [exploreData, pathData, mode]);
```

The effect uses `mergeRawGraphData`, `setGraphData`, and `setRawGraphData` from the store but they are not in the dependency array. Zustand store functions are stable references (they don't change between renders), so this will not cause bugs in practice. But the ESLint react-hooks/exhaustive-deps rule would flag it. Worth noting but not a functional issue.

---

### `handleRemoveFromGraph` action label is misleading

**Location**: `web/src/explorers/common/useGraphContextMenu.ts:135-151`

```typescript
store.addExplorationStep({
  action: 'add-adjacent',  // <-- this is a removal, not an add
  op: '-',
  cypher: stepToCypher({ action: 'add-adjacent', ... }),
  ...
});
```

The `action` is `'add-adjacent'` but this is semantically a removal. The `op: '-'` correctly marks it as subtractive, and the Cypher it generates via `stepToCypher` will be the neighborhood MATCH (which is correct for replay -- subtracting a neighborhood means "remove these nodes"). However, in the generated Cypher script comments, this will appear as:

```
-- Step N: add-adjacent "NodeLabel"
- MATCH (c:Concept)-[r]-(n:Concept) WHERE c.label = 'NodeLabel' RETURN c, r, n;
```

The comment says "add-adjacent" for a subtraction step, which could be confusing when reading exported scripts. Consider adding a `'remove'` action type, or at minimum adjusting the label generation in `generateCypher` to account for the op.

---

### What looks good

- **`deriveMode` pattern**: Eliminating the explicit mode enum in favor of parameter-presence is a clean design. `DerivedMode` being a pure function of `SearchParams` means no possibility of mode/params getting out of sync.

- **`cypherGenerator.ts` / `cypherResultMapper.ts`**: Well-structured modules with clear doc comments. The `escapeCypher` function handles the right escape cases. The `parseCypherStatements` parser correctly handles continuation lines, blank-line flushing, and default-to-additive for unprefixed statements.

- **AGE ID mapping in `cypherResultMapper.ts`**: The `internalToConceptId` map correctly remaps relationship endpoints from AGE internal vertex IDs to `concept_id` property values. This is the right place to centralize that mapping.

- **`subtractRawGraphData`**: The cascading link removal (removing links that reference removed nodes via `remainingNodeIds` set) is correct and handles both `from_id`/`to_id` and D3-mutated `source.id`/`target.id` forms.

- **`usePathEnrichment` guard rails**: Capping enrichment depth at 2 and skipping if >50 nodes prevents accidental explosion. The `useQueries` pattern for parallel fetching is appropriate here.

- **Migration 050**: Properly idempotent with the `schema_migrations` check and `ON CONFLICT DO NOTHING`.

- **Settings panel relocation to sidebar**: Moving graph settings from floating overlays inside the explorer to the IconRailPanel sidebar is a good UX improvement. The `embedded` prop on `GraphSettingsPanel` and `Settings3DPanel` is a clean way to support both contexts.

---

### Summary

| Category | Count | Severity |
|----------|-------|----------|
| Bugs / logic errors | 1 (always-`+` ternary) | Low -- cosmetic/confusing but not functionally broken |
| Race conditions | 1 (step before API) | Medium -- causes phantom steps on failure |
| Duplication | 1 (path extraction x3) | Medium -- maintenance risk |
| Type safety | 1 (pervasive `any`) | Low -- existing pattern, not new debt |
| Misleading semantics | 1 (remove uses add-adjacent action) | Low -- affects script readability |
| Stale state window | 1 (subtract without clearing graphData) | Low |

None of these are merge-blockers. The most actionable items are the path extraction duplication (extract to shared utility) and the step-before-API-call pattern (move recording to success path).

---

PR #291, opened 2026-02-04
feat: containerize graph_accel extension (ADR-201 Phase 5c)

aaronsb:
## Summary

- Add container build pipeline for graph_accel: multi-stage Dockerfile.build compiles inside `apache/age` for ABI compatibility, `build-in-container.sh` wrapper extracts artifacts to `dist/pg17/`
- Create `kg-postgres` image (`docker/Dockerfile.postgres`): thin layer over `apache/age` with pre-built extension + schema init scripts
- Update all 5 compose overlays (base builds locally, ghcr/prod/standalone use GHCR image, dev uses stock apache/age)
- Add schema init (`11_graph_accel.sql`) and migration (`051_graph_accel_extension.sql`) with conditional CREATE EXTENSION
- Add `postgres` target to `publish.sh`, update `deploy-option0.sh` to prefer `dist/` artifacts
- Pre-built artifacts (`.so`, `.control`, `.sql`) committed for drop-in use with any `apache/age` container

## Test plan

- [x] `build-in-container.sh` produces three artifacts in `dist/pg17/`
- [x] `docker build -f docker/Dockerfile.postgres .` completes in <10s
- [x] Extension loads: `graph_accel_status()` returns 945 nodes, 2664 edges
- [x] `make test` passes (978 passed, 0 failed)
- [x] Dev mode: stock `apache/age` + `deploy-option0.sh` installs extension
- [ ] Fresh `operator.sh init` creates extension via init script
- [ ] `operator.sh upgrade` applies migration 051

### DIFF
--- a/graph-accel/README.md
+++ b/graph-accel/README.md
@@ -245,23 +245,29 @@ cargo build --release -p graph-accel-bench
 
 ### Pre-built artifacts (recommended)
 
-Pre-built artifacts for `apache/age` (PostgreSQL 17, Debian trixie) are in `dist/pg17/`:
+Pre-built artifacts for `apache/age` (PostgreSQL 17, Debian trixie) are in `dist/pg17/`, organized by architecture:
 
 ```
 dist/pg17/
-├── graph_accel.so          # Shared library
-├── graph_accel.control     # Extension metadata
-└── graph_accel--0.5.0.sql  # Function definitions
+├── amd64/
+│   ├── graph_accel.so          # Shared library (x86-64)
+│   ├── graph_accel.control     # Extension metadata
+│   └── graph_accel--0.5.0.sql  # Function definitions
+└── arm64/
+    ├── graph_accel.so          # Shared library (aarch64)
+    ├── graph_accel.control
+    └── graph_accel--0.5.0.sql
 ```
 
-Copy these three files into any `apache/age` container:
+Copy the three files matching your container's architecture into any `apache/age` container:
 
 ```bash
-docker cp dist/pg17/graph_accel.so \
+# Use amd64/ or arm64/ to match your container architecture
+docker cp dist/pg17/amd64/graph_accel.so \
   my-container:/usr/lib/postgresql/17/lib/
-docker cp dist/pg17/graph_accel.control \
+docker cp dist/pg17/amd64/graph_accel.control \
   my-container:/usr/share/postgresql/17/extension/
-docker cp dist/pg17/graph_accel--0.5.0.sql \
+docker cp dist/pg17/amd64/graph_accel--0.5.0.sql \
   my-container:/usr/share/postgresql/17/extension/
 
 docker exec my-container psql -U postgres -d mydb \
@@ -275,8 +281,10 @@ docker exec my-container psql -U postgres -d mydb \
 To rebuild for the current `apache/age` image (guarantees ABI compatibility):
 
 ```bash
-./build-in-container.sh
-# → produces dist/pg17/{graph_accel.so, .control, .sql}
+./build-in-container.sh              # host arch only (fast)
+./build-in-container.sh --all        # amd64 + arm64
+./build-in-container.sh --platform linux/arm64  # specific arch
+# → produces dist/pg17/<arch>/{graph_accel.so, .control, .sql}
 ```
 
 This spins up a temporary Docker container with the `apache/age` base, installs the Rust toolchain + pgrx, compiles the extension, and extracts the artifacts. Docker layer caching makes rebuilds fast.
@@ -285,9 +293,10 @@ This spins up a temporary Docker container with the `apache/age` base, installs 
 
 ```dockerfile
 FROM apache/age
-COPY dist/pg17/graph_accel.so /usr/lib/postgresql/17/lib/
-COPY dist/pg17/graph_accel.control /usr/share/postgresql/17/extension/
-COPY dist/pg17/graph_accel--*.sql /usr/share/postgresql/17/extension/
+ARG TARGETARCH
+COPY dist/pg17/${TARGETARCH}/graph_accel.so /usr/lib/postgresql/17/lib/
+COPY dist/pg17/${TARGETARCH}/graph_accel.control /usr/share/postgresql/17/extension/
+COPY dist/pg17/${TARGETARCH}/graph_accel--*.sql /usr/share/postgresql/17/extension/
 ```
 
 ### Development deployment
@@ -310,7 +319,7 @@ graph-accel/
 ├── Cargo.toml          # Workspace: core, bench, ext
 ├── build-in-container.sh  # Build inside apache/age for ABI compatibility
 ├── Dockerfile.build    # Multi-stage Dockerfile for container builds
-├── dist/pg17/          # Pre-built artifacts for apache/age (PG 17)
+├── dist/pg17/{amd64,arm64}/  # Pre-built artifacts for apache/age (PG 17)
 ├── core/               # Pure Rust traversal engine
 │   └── src/
 │       ├── graph.rs    #   Adjacency list, node index, rel-type interning


aaronsb:
## Code Review: PR #291 -- Containerize graph_accel extension (ADR-201 Phase 5c)

**Scope**: 21 files, 847 additions, 93 deletions across 5 commits. Significant change introducing a multi-arch container build pipeline for a Rust pgrx PostgreSQL extension, plus Docker Compose integration across all deployment modes (dev, ghcr, standalone, prod).

---

### Assessment: Solid, well-structured work with one documentation gap

The overall architecture is clean and carefully layered. The digest-pinned base images, the per-arch artifact layout, the `target_multi_arch` fallback logic, and the defensive schema migrations all demonstrate real production thinking. I have one issue to flag and a few observations worth capturing.

---

### Issue: README uses stale flat `dist/pg17/` paths (no arch subdir)

**Location**: `graph-accel/README.md` lines 248-265, 279, 288-290, 313

**Problem**: The README was rewritten as part of this PR, but the new text references the **old flat** path layout (`dist/pg17/graph_accel.so`) rather than the per-arch layout (`dist/pg17/amd64/graph_accel.so`) that every operational file in this PR uses.

Affected sections:
- The directory tree diagram (line 250-254) shows a flat `dist/pg17/` with no arch subdirs
- The manual `docker cp` examples (lines 260-265) reference `dist/pg17/graph_accel.so` etc.
- The `build-in-container.sh` output comment (line 279) says `dist/pg17/{...}`
- The example Dockerfile snippet (lines 288-290) uses `COPY dist/pg17/graph_accel.so`
- The project structure tree (line 313) shows `dist/pg17/` without the arch level

Meanwhile, the actual Dockerfiles, build scripts, deploy scripts, and .gitignore all consistently use `dist/pg17/<arch>/`. Someone following the README instructions would get confused or hit file-not-found errors.

**Suggestion**: Update the README to reflect the actual per-arch structure:

```
dist/pg17/
├── amd64/
│   ├── graph_accel.so
│   ├── graph_accel.control
│   └── graph_accel--0.5.0.sql
└── arm64/
    ├── graph_accel.so
    ├── graph_accel.control
    └── graph_accel--0.5.0.sql
```

And update the docker cp examples similarly (e.g., `dist/pg17/amd64/graph_accel.so`). The example Dockerfile snippet should use `COPY dist/pg17/${TARGETARCH}/...` to match what `Dockerfile.postgres` actually does, or document a single-arch simple case with a note that multi-arch uses the `${TARGETARCH}` pattern.

**Severity**: Non-blocking but should fix before merge. Documentation that contradicts the implementation will confuse anyone trying to use these artifacts manually.

---

### Verification of checklist items

**1. Per-arch artifact structure**: Consistent. `.gitignore` exceptions cover both `amd64/` and `arm64/` with the full parent chain (`!graph-accel/dist/`, `!graph-accel/dist/pg17/`, `!graph-accel/dist/pg17/amd64/`, then the file globs). `Dockerfile.build` stage 2 outputs to `/pg17/${TARGETARCH}/`. `Dockerfile.postgres` copies from `${TARGETARCH}/`. `build-in-container.sh` cleans and writes to `$DIST_DIR/pg17/$arch`. `deploy-option0.sh` resolves `$DIST_ARCH` from the container's `uname -m`. All aligned.

**2. Dockerfile.postgres TARGETARCH**: Correct. `ARG TARGETARCH` is declared (line 24) and used in all three COPY instructions (lines 32-34). Docker BuildKit auto-populates this during `docker buildx build --platform`.

**3. Dockerfile.build TARGETARCH**: Correct. The `artifacts` stage (from scratch) declares `ARG TARGETARCH` and uses it in `COPY --from=builder /artifacts/ /pg17/${TARGETARCH}/`. Combined with `--output type=local,dest=$DIST_DIR`, this produces the correct per-arch directory.

**4. Digest pinning**: Both Dockerfiles pin to the identical digest `sha256:4ba7dd683fc83fcac0e603f76448f6afba0507cf4d7d21673cede0d6ff0265b2`. Both include comments instructing to update them together. This is the correct pattern for ABI compatibility.

**5. publish.sh multi-arch logic**: Well-implemented. The `target_multi_arch` variable is initialized per-target from `use_multi_arch` (line 641), then the postgres case can demote it to `false` if arm64 artifacts are missing (line 178). Both the buildx build branch (line 696) and the single-arch push branch (line 732) correctly reference `target_multi_arch` instead of `use_multi_arch`. The amd64 directory check gates the entire postgres target (`continue` if missing), which prevents a broken build.

**6. deploy-option0.sh arch detection**: Correct. It runs `docker exec "$CONTAINER" uname -m` to get the *container's* architecture (not the host's), maps to the dist arch name, and uses `$DIST_DIR` consistently. The fallback to `amd64` for unknown architectures is a reasonable default.

**7. build-in-container.sh flags**: `--all` sets both platforms, `--platform linux/arm64` adds to the platforms array, bare invocation defaults to host arch via `uname -m` mapping. The buildx builder is only created when cross-platform or cross-arch is needed (line 62 check). The platform argument parsing has a minor quirk (see observation below) but functions correctly for documented usage.

**8. Docker Compose overlays**:
- **Base** (`docker-compose.yml`): Builds locally from `Dockerfile.postgres`, volume-mounts init scripts.
- **GHCR** (`docker-compose.ghcr.yml`): Overrides image to `kg-postgres:latest` from GHCR.
- **Dev** (`docker-compose.dev.yml`): Overrides back to stock `apache/age` with volume mounts for init scripts. Extension deployed via `deploy-option0.sh`.
- **Standalone** (`docker-compose.standalone.yml`): Uses `volumes: !reset` to drop the init script mounts (baked into GHCR image). Keeps data and import volumes.
- **Prod** (`docker-compose.prod.yml`): Uses GHCR `kg-postgres:latest`, removes the `00_baseline.sql` mount that was previously there. Correctly relies on baked-in init scripts.

All overlays are internally consistent and compose correctly in their intended chains.

**9. Schema migration safety**: Both `schema/11_graph_accel.sql` and `schema/migrations/051_graph_accel_extension.sql` check `pg_available_extensions` before attempting `CREATE EXTENSION`. The migration additionally checks `schema_migrations` for idempotency and wraps in a transaction. The init script runs only on first database creation (Docker entrypoint behavior). Safe and defensive.

**10. .gitignore exceptions**: Cover both arch subdirectories with the full unignore chain from `!graph-accel/dist/` through to the file-level globs for `*.so`, `*.control`, `*.sql`. This is the correct git pattern for selectively committing files under an otherwise-ignored `dist/` rule. The global `*.so` ignore (line 25) is properly overridden by the more specific `!graph-accel/dist/pg17/amd64/*.so` pattern.

**11. Leftover flat-path references**: All operational code (Dockerfiles, build scripts, deploy scripts, publish.sh, .gitignore, compose files) uses the per-arch path. The only flat `dist/pg17/graph_accel*` references are in `graph-accel/README.md` (documented as the issue above).

---

### Observations (non-blocking)

**build-in-container.sh argument parsing quirk**: The `--platform` flag matches as a case but does nothing (line 38-39). The actual platform value (`linux/amd64` or `linux/arm64`) is caught by the next case branch. This means `--platform linux/arm64` works, but `--platform=linux/arm64` would fail with "Unknown argument". The documented usage only shows the space-separated form, so this is fine in practice, but the two-token parsing model could be made more explicit with a comment, or by consuming `--platform` + shift to get the next arg.

**Binary artifacts in git**: Committing the `.so` files (1.3MB amd64, 1MB arm64) is a conscious trade-off documented in the PR. It avoids requiring a Rust toolchain for `docker build`, which is the right call for this project's deployment model. Worth noting that these will accumulate on version bumps -- consider `.gitattributes` with `filter=lfs` if the .so files start changing frequently.

**Dockerfile.build layer caching**: The source COPY happens after all toolchain installation, so Rust source changes get good cache hits on the toolchain layers. The `cargo pgrx package` step will re-run on any source change, which is correct. One small optimization: copying `Cargo.toml`/`Cargo.lock` first and running a dummy build could cache dependency downloads, but given this is a periodic rebuild (not CI), it is a minor point.

---

### Summary

The multi-arch containerization is architecturally sound. The digest pinning, per-arch artifact layout, defensive schema migrations, and graceful fallback paths in publish.sh all demonstrate careful engineering. The Docker Compose overlay strategy correctly differentiates dev (stock image + deploy-option0), base (local build), and production (GHCR image with baked-in init) modes.

Fix the README flat-path references to match the actual per-arch directory structure, and this is ready to merge.

PR #304, opened 2026-02-07
feat: route web client through GraphProgram execution

aaronsb:
## Summary

- Routes all web client Cypher execution through `POST /programs/execute` instead of raw `POST /query/cypher`
- Reroutes 4 execution points: SearchBar (multi-statement editor), BlockBuilder (block-compiled Cypher), useQueryReplay (saved query replay), DocumentExplorerWorkspace (inter-concept edges)
- Removes `executeCypherQuery()` from web API client — zero callers remain after rerouting
- Adds `programBuilder.ts` helpers (`cypherToStatement`, `statementsToProgram`) to construct GraphProgram ASTs from web app actions
- Adds `mapWorkingGraphToRawGraph()` to bridge server-side WorkingGraph format to client-side RawGraphData

## Context

The web app was the last client sending unnotarized Cypher directly to the database via `POST /query/cypher`. CLI and MCP already use `POST /programs/execute` exclusively. This PR closes the gap — all query execution now goes through the notarized GraphProgram pipeline with validation, safety gates, and execution logging.

The block builder is the human equivalent of the MCP program tool for agents. Both compose GraphPrograms, both now execute through the same pipeline.

## Test plan

- [ ] `cd web && npx tsc -b --noEmit` — type check passes
- [ ] Cypher editor: execute multi-statement script, graph renders
- [ ] Block builder: compile and execute blocks, graph renders
- [ ] Save exploration → reload → replay works
- [ ] Document explorer: inter-concept edges load
- [ ] `grep -r executeCypherQuery web/src/` → zero results

### DIFF
--- a/api/app/lib/polarity_axis.py
+++ b/api/app/lib/polarity_axis.py
@@ -312,11 +312,12 @@ def discover_candidate_concepts_parallel(
     validate_concept_id(positive_pole_id)
     validate_concept_id(negative_pole_id)
 
-    # Fall back to sequential for max_hops > 2 (safety)
+    # Parallel discovery only supports 1-2 hops structurally; higher hops use
+    # the sequential variable-length path query (fast with graph_accel).
     if max_hops > 2:
-        logger.warning(
-            f"max_hops={max_hops} exceeds parallel query limit (2), "
-            "falling back to sequential implementation"
+        logger.info(
+            f"max_hops={max_hops} exceeds parallel implementation limit (2), "
+            "using sequential variable-length path query"
         )
         return discover_candidate_concepts(
             positive_pole_id=positive_pole_id,


aaronsb:
## Review: Route web client through GraphProgram execution

**Scope**: 287 additions, 71 deletions across 9 files. Standard-tier change.

---

### 1. Execution Point Reroutes (4/4 complete and correct)

All four execution points are properly rerouted:

| Site | File | Pattern |
|------|------|---------|
| Cypher editor | `SearchBar.tsx:303-305` | `statementsToProgram` -> `executeProgram` -> `mapWorkingGraphToRawGraph` |
| Block builder | `BlockBuilder.tsx:635-658` | `cypherToStatement` -> `executeProgram` -> `mapWorkingGraphToRawGraph` |
| Query replay | `useQueryReplay.ts:53-57` | `statementsToProgram` -> `executeProgram` -> `mapWorkingGraphToRawGraph` |
| Doc explorer | `DocumentExplorerWorkspace.tsx:222-228` | `cypherToStatement` -> `executeProgram` -> `mapWorkingGraphToRawGraph` |

Each follows the same three-step pattern (build AST, execute, map result), which is good for consistency.

### 2. mapWorkingGraphToRawGraph field mapping -- correct

`cypherResultMapper.ts:187-209` maps all fields from `WorkingGraph` to `RawGraphData`:

- **Nodes**: `concept_id`, `label`, `ontology` (null-coalesced to `'default'`), `description`, plus properties-bag extractions for `search_terms`, `grounding_strength`, `diversity_score`, `evidence_count`
- **Links**: `from_id`, `to_id`, `relationship_type`, `category`, `confidence`, `grounding_strength`

This aligns with the `RawGraphNode` and `RawGraphLink` interfaces. The null-to-undefined coercion (`?? undefined`) is correct for TypeScript optional fields.

The server-side Python models (`RawNode`, `RawLink` in `program.py:391-407`) match the TypeScript `ProgramRawNode` and `ProgramRawLink` types structurally -- field names, types, and optionality all align.

### 3. Type safety: `as unknown as Record<string, unknown>` casts

Four call sites use this cast pattern to pass a typed `GraphProgram` to `executeProgram` which accepts `Record<string, unknown>`. This is a pragmatic workaround for the structural mismatch between `GraphProgram` (typed AST with `version: 1` literal type) and the `Record<string, unknown>` the API client expects.

**Assessment**: Acceptable. The alternative would be changing `executeProgram` to accept `GraphProgram | Record<string, unknown>`, but since the server deserializes from untyped JSON anyway, the cast is cosmetic. The `tsc -b --noEmit` check passes clean, confirming no type holes escape.

**Minor improvement opportunity**: The `executeProgram` signature could accept `program?: GraphProgram | Record<string, unknown>` to eliminate the cast at all call sites. Low priority -- the cast is safe and the pattern is already established in the codebase (e.g., `reportStore.ts` uses the same idiom).

### 4. executeCypherQuery removal -- clean

`grep -r executeCypherQuery web/src/` returns zero results. The method is fully removed from `client.ts` and no callers remain. The `/query/cypher` endpoint path also returns zero references in web/src. Clean removal.

### 5. SearchBar multi-statement handling -- behavioral change worth noting

**Before**: Each statement executed individually in a loop. Each result was merged or subtracted incrementally (`stmt.op === '+' ? merge : subtract`). This meant the `-` operator actually removed nodes from the client-side graph.

**After**: All statements execute as a single `GraphProgram`. The server-side executor handles the set algebra. The final `WorkingGraph` is the complete result, and the client does a single `mergeRawGraphData(mapped)`.

This is correct -- the server-side executor applies `+` and `-` operators during execution, so the returned `WorkingGraph` already reflects all operations. The client no longer needs to know about `-` semantics.

**One dead import**: `mapCypherResultToRawGraph` is still imported in `SearchBar.tsx:26` but never called in the function body. It should be removed from the import to avoid confusion. (Not a bug, just dead code.)

**One dead destructured variable**: `subtractRawGraphData` is still destructured from `useGraphStore` in `useQueryReplay.ts:34` and listed in the dependency array at line 102, but is never called. Same category -- dead reference.

### 6. BlockBuilder reshaping -- works but leaves vestigial code

`BlockBuilder.tsx:641-658` correctly maps `WorkingGraph` results back into the `{ nodes, relationships }` shape that the downstream `internalToConceptId` and `rawNodes` processing expects. The mapping is field-complete.

**However**, the code at lines 669-674 still builds an `internalToConceptId` map and the comment says "No AGE internal ID mapping needed". The comment is accurate -- the map is now a no-op identity mapping (since `n.id` is already the `concept_id`). The map is still technically *used* downstream (it feeds into the old reshape at lines 682-688), so it cannot simply be deleted without refactoring the downstream code. But the comment correctly documents that this is vestigial.

The downstream reshape at lines 682-688 also re-extracts properties that were *just* placed there by the mapping at lines 641-658. This is a double-mapping: `WorkingGraph -> { nodes, relationships } -> rawNodes`. It would be cleaner to feed the `mapped` result directly into the rawNodes/rawLinks pipeline, skipping the intermediate format entirely. That said, this is a refactoring opportunity, not a bug -- the current code produces correct output.

### 7. programBuilder.ts -- clean utility

`cypherToStatement` and `statementsToProgram` are simple, focused, and well-typed. The `metadata: { author: 'human' }` annotation in `statementsToProgram` is a good practice for distinguishing human-originated programs from agent-originated ones in execution logs.

### 8. API client methods -- correct routing

- `executeProgram` posts to `/programs/execute` with 60s timeout (matches the old `executeCypherQuery` timeout)
- `chainPrograms` posts to the same endpoint with `deck` body field and 120s timeout (reasonable for multi-program chains)
- `listPrograms` uses GET `/programs` with query params

The endpoint discrimination (single vs chain mode via presence of `deck` field) matches the server-side `ProgramExecuteRequest` validator.

### Summary

The rerouting is structurally sound. All four execution paths correctly construct program ASTs, execute through the notarized pipeline, and map results to the existing `RawGraphData` format. The `executeCypherQuery` removal is clean.

**Non-blocking findings** (cleanup, not correctness):
1. Dead import: `mapCypherResultToRawGraph` in `SearchBar.tsx:26`
2. Dead destructure: `subtractRawGraphData` in `useQueryReplay.ts:34` and dependency array at line 102
3. Vestigial double-mapping in `BlockBuilder.tsx:641-688` (correct but redundant intermediate format)

None of these block merge.

PR #309, opened 2026-02-08
feat: wire vocabulary lifecycle to LLM and fix CLI UX

aaronsb:
## Summary

- Wire vocabulary lifecycle stubs to real LLM decisions and database execution
- Extract shared `call_llm_sync()` utility, eliminating three duplicate provider-dispatch blocks
- Fix stale embedding bootstrap: resolve dimensions eagerly from provider chain + DB config
- Unify CLI job status naming with shared `resolveStatusFilter()` across all subcommands
- Add embedding dimension guards and remove silent fallback paths
- Replace outdated vocabulary categories doc with lifecycle guide

## Test plan

- [x] `pytest tests/test_pruning_strategies.py` — 38 passed
- [x] `pytest tests/test_synonym_detector.py` — 24 passed
- [x] `pytest tests/test_vocabulary_manager_integration.py` — 22 passed
- [x] `pytest tests/unit/lib/test_annealing_evaluator.py` — 14 passed
- [x] `pytest tests/` — 1175 passed, 23 skipped, 0 failures
- [x] Live `kg vocab consolidate --target 50` executed merges correctly
- [x] `kg job list -s pending/done/running` resolves aliases correctly

### DIFF
--- a/api/app/lib/annealing_evaluator.py
+++ b/api/app/lib/annealing_evaluator.py
@@ -15,6 +15,8 @@ import logging
 from dataclasses import dataclass
 from typing import Dict, Any, List, Optional
 
+from api.app.lib.llm_utils import call_llm_sync
+
 logger = logging.getLogger(__name__)
 
 
@@ -209,8 +211,6 @@ Respond with ONLY the JSON, no other text."""
 
 def _call_llm_sync(prompt: str, ai_provider) -> str:
     """Synchronous LLM call. Runs in a thread pool via _call_llm."""
-    from api.app.lib.llm_utils import call_llm_sync
-
     return call_llm_sync(
         ai_provider,
         prompt=prompt,
--- a/api/app/lib/pruning_strategies.py
+++ b/api/app/lib/pruning_strategies.py
@@ -50,6 +50,7 @@ import logging
 # Type imports for other ADR-032 modules
 from api.app.lib.synonym_detector import SynonymCandidate, SynonymStrength
 from api.app.lib.vocabulary_scoring import EdgeTypeScore
+from api.app.lib.llm_utils import call_llm_sync
 
 logger = logging.getLogger(__name__)
 
@@ -178,8 +179,6 @@ Consider:
 Respond with ONLY the JSON, no other text."""
 
     try:
-        from api.app.lib.llm_utils import call_llm_sync
-
         content = call_llm_sync(
             ai_provider,
             prompt=prompt,
@@ -613,8 +612,6 @@ class PruningStrategy:
             ValueError: If provider type is unsupported
             Exception: If the LLM call fails
         """
-        from api.app.lib.llm_utils import call_llm_sync
-
         return call_llm_sync(
             self.ai_provider,
             prompt=prompt,
--- a/api/app/lib/synonym_detector.py
+++ b/api/app/lib/synonym_detector.py
@@ -163,6 +163,8 @@ class SynonymDetector:
         if hasattr(provider, 'model_manager') and provider.model_manager is not None:
             try:
                 dims = provider.model_manager.get_dimensions()
+                # isinstance guard: MagicMock auto-creates attributes and returns
+                # non-int sentinels; treat those as "not available".
                 if isinstance(dims, int):
                     return dims
             except (RuntimeError, AttributeError):


aaronsb:
## Review: commit 6f3eafd7 — consolidate LLM dispatch + fix stale embedding bootstrap

Reviewed both fixes against the original code at 638bd880. Here is a structured assessment organized around the six review checklist items.

---

### 1. Does `call_llm_sync()` faithfully preserve behavior of all three original dispatch blocks?

**Verdict: Yes, with one intentional improvement and two subtle behavioral changes.**

The three original dispatch sites were:

| Site | system_msg | max_tokens | json_mode | top_p (Ollama) |
|------|-----------|-----------|-----------|---------------|
| `llm_evaluate_merge` (pruning) | "vocabulary expert" | 300 | OpenAI: yes, Ollama: yes | `ai_provider.top_p` (always) |
| `_call_llm_sync` (annealing) | "architect" | 400 | OpenAI: yes, Ollama: yes | **not sent** |
| `PruningStrategy._call_llm` | caller-supplied | 300 | **not used** | `ai_provider.top_p` (always) |

The consolidated `call_llm_sync()` faithfully routes `system_msg`, `max_tokens`, `temperature`, and `json_mode` through parameters — callers pass exactly the values they had before. All three call sites now pass the correct values. Good.

### 2. Behavioral differences introduced?

**Two items worth noting:**

**(a) `json_mode` now correctly not sent from `PruningStrategy._call_llm`** — The original `PruningStrategy._call_llm` (used by `_ai_review_synonym` and `_ai_review_low_value`) never used JSON mode. The refactored call at `pruning_strategies.py:618` passes `json_mode=False` by default, which is correct. No issue here.

**(b) Ollama `top_p` handling changed from unconditional to conditional** — This is the most noteworthy behavioral difference:

- **Original** `llm_evaluate_merge`: always sent `"top_p": ai_provider.top_p` — would raise `AttributeError` if `top_p` was missing.
- **Original** `PruningStrategy._call_llm`: same — always accessed `self.ai_provider.top_p` unconditionally.
- **Original** `_call_llm_sync` (annealing): never sent `top_p` at all.
- **New** `call_llm_sync`: uses `if hasattr(ai_provider, "top_p")` guard (line 82).

This is a **defensively better** approach — it unifies the three behaviors into one that doesn't crash if `top_p` is absent, while still forwarding it when present. The annealing evaluator gains `top_p` forwarding it didn't have before (harmless; Ollama ignores unrecognized options). The two pruning sites lose a hard crash on a missing attribute that was never a useful failure mode. This is a net improvement.

**No issue — the change is an improvement over the inconsistent originals.**

### 3. Is `isinstance(dims, int)` the right approach for MagicMock in tests?

**Location**: `synonym_detector.py:166, 176`

This is a pragmatic guard. When tests use `MagicMock()` as the `ai_provider`, `model_manager` is itself a `MagicMock`, and `get_dimensions()` returns a `MagicMock` (not an int). The `isinstance(dims, int)` check causes the method to return `None` in test contexts, which is the desired behavior — it lets `_expected_dims` stay `None` so the lazy fallback path at lines 550-551 and 565-566 can still set dims from actual embedding shapes.

**This is acceptable.** However, it implicitly encodes a test concern into production code. The alternative would be to check for `get_dimensions` as a callable returning a numeric type, or to have the test fixtures properly configure `model_manager.get_dimensions.return_value = 3`. The latter would be cleaner and let you drop the `isinstance` guard, but the current approach is a reasonable tradeoff given that mock providers are used extensively in this codebase.

**Minor suggestion (non-blocking):** Consider adding a brief inline comment on one of the `isinstance(dims, int)` lines explaining that this guards against MagicMock returns in tests. It's not obvious why a method named `get_dimensions()` would return a non-int.

### 4. Could the DB query in `_dims_from_db()` cause issues at init time?

**Location**: `synonym_detector.py:183-192`

This method does a deferred import of `load_active_embedding_config` and calls it, which internally creates an `AGEClient` and queries `kg_api.embedding_config`. The entire body is wrapped in a bare `except Exception: pass`.

**Two considerations:**

**(a) Connection not ready at init time** — If `SynonymDetector` is instantiated before the database pool is available (e.g., during module import or early app startup), the query will fail. The `except Exception: pass` handles this gracefully — `_dims_from_db()` returns `None`, and the detector falls back to lazy resolution. This is fine.

**(b) Circular imports** — The import chain is `synonym_detector -> embedding_config -> age_client`. Since the import is deferred (inside the function body, not at module level), it avoids circular import problems. This is the same pattern used elsewhere in the codebase (e.g., `_get_edge_type_embedding` at line 530 also does `from .age_client import AGEClient`).

**No issue.** The defensive `try/except` makes this safe for early-init scenarios.

### 5. Is `RuntimeError` on dimension mismatch the right failure mode?

**Location**: `synonym_detector.py:149-152`

When provider reports one dimension and DB config reports another, the code raises:

```
RuntimeError: Embedding dimension mismatch: provider reports 768, DB config says 1536.
Check embedding_config table.
```

**This is the right call.** A dimension mismatch means the operator switched embedding models without updating the DB config (or vice versa). Proceeding silently would produce garbage cosine similarities — every comparison would raise a `ValueError` downstream anyway (from `_cosine_similarity` at line 624). Failing loudly at init time with a clear error message and remediation hint ("Check embedding_config table") is much better than cryptic shape-mismatch errors scattered across async embedding operations.

The error message is actionable, which is important for an operator-managed system.

### 6. Local imports inside functions

Three local import sites:

| Location | Import |
|----------|--------|
| `annealing_evaluator.py:212` | `from api.app.lib.llm_utils import call_llm_sync` |
| `pruning_strategies.py:181` | `from api.app.lib.llm_utils import call_llm_sync` |
| `pruning_strategies.py:616` | `from api.app.lib.llm_utils import call_llm_sync` |

All three are deferred imports inside function bodies, not at module level.

**This is an established pattern in this codebase** — `synonym_detector.py` itself does `from .age_client import AGEClient` inside `_get_edge_type_embedding`, and `embedding_config.py` does the same. The motivation is avoiding circular imports in a codebase where `lib/` modules frequently cross-reference each other.

**However**, `llm_utils.py` has no imports from `api.app.lib` at all — it only imports `logging` and `typing`. There is no circular dependency risk from making these top-level imports. The deferred style is harmless (negligible performance cost for functions that make LLM calls), but switching to top-level imports in `annealing_evaluator.py` and `pruning_strategies.py` would be cleaner and signal to future readers that no circularity concern exists.

**Non-blocking suggestion.** The current pattern works; top-level imports would be marginally cleaner.

---

### Overall Assessment

Both fixes are solid.

**Fix 1 (LLM dispatch):** Successfully collapses three duplicate dispatch blocks into one. The `call_llm_sync` signature covers all the parameter variations needed by the three call sites. The `top_p` handling is actually improved (defensive `hasattr` vs. unconditional access). The docstring and `@verified` stamp are present. No behavioral regressions.

**Fix 2 (stale embedding bootstrap):** Eagerly resolving dimensions from both provider chain and DB config is the right architecture — it catches model-switch misconfigurations at init time rather than producing mysterious failures during synonym detection. The `isinstance(dims, int)` guard is pragmatic for test compatibility. The bare `except Exception: pass` in `_dims_from_db` is appropriate for a best-effort dimension check that should never block initialization.

**No blocking issues.** Two non-blocking suggestions noted above (comment on `isinstance` guard; consider top-level imports for `llm_utils`).

PR #311, opened 2026-02-09
feat: autonomous annealing proposal execution

aaronsb:
## Summary

- Annealing proposals previously stalled at `pending` because nothing auto-approved them — the `automation_level` setting existed in the DB but was never read
- Default mode is now **autonomous**: proposals are approved and dispatched for execution within the same annealing cycle, like vocabulary consolidation
- **HITL mode** preserved for diagnostics: set `automation_level = 'hitl'` to hold proposals for human review
- Stale pending proposals (past 7-day TTL) are now expired by the job scheduler cleanup cycle

## Changes

| File | What |
|------|------|
| `api/app/services/annealing_manager.py` | Return proposal IDs from cycle (was count-only) |
| `api/app/launchers/annealing.py` | Pass `automation_level` through job_data |
| `api/app/workers/annealing_worker.py` | Auto-approve + dispatch when autonomous; new `_auto_approve_and_dispatch()` |
| `api/app/services/job_scheduler.py` | Expire stale proposals in cleanup cycle |
| `schema/migrations/053_annealing_autonomous_default.sql` | Change default from `hitl` to `autonomous` |
| `docs/guides/SCHEDULED-JOBS.md` | Add annealing section with mermaid sequence diagram |

## Design

The auto-approve path in the worker reuses the same execution infrastructure as the human review endpoint:
1. Atomic `UPDATE ... WHERE status='pending'` prevents double-dispatch
2. Enqueue `proposal_execution` job via existing job queue
3. `ProposalExecutor` handles promotion/demotion (unchanged)

No changes to `proposal_execution_worker.py`, `proposal_executor.py`, or `routes/ontology.py`.

## Test plan

- [ ] Full test suite passes (1175 passed, 23 skipped)
- [ ] Migration 053 applies cleanly on existing install
- [ ] `kg ontology annealing --dry-run` shows candidates without executing
- [ ] `kg ontology annealing` in autonomous mode creates + executes proposals
- [ ] `kg ontology proposals --status pending` returns empty in autonomous mode (proposals don't linger)
- [ ] Setting `automation_level = 'hitl'` causes proposals to wait for review

### DIFF
--- a/api/app/routes/ontology.py
+++ b/api/app/routes/ontology.py
@@ -484,10 +484,11 @@ async def trigger_annealing_cycle(
     _: None = Depends(require_permission("ontologies", "write")),
 ):
     """
-    Trigger a manual annealing cycle (ADR-200 Phase 3b).
+    Trigger a manual annealing cycle (ADR-200).
 
     Scores all ontologies, recomputes centroids, identifies candidates,
-    and generates proposals for human review.
+    and generates proposals. In autonomous mode (default), proposals are
+    auto-approved and executed. In hitl mode, proposals wait for review.
 
     Use `dry_run=true` to preview candidates without generating proposals.
 
@@ -496,12 +497,33 @@ async def trigger_annealing_cycle(
     from api.app.services.job_queue import get_job_queue
 
     queue = get_job_queue()
+
+    # Read automation_level from DB so manual triggers respect the setting
+    automation_level = "autonomous"
+    try:
+        client = get_age_client()
+        conn = client.pool.getconn()
+        try:
+            with conn.cursor() as cur:
+                cur.execute(
+                    "SELECT value FROM kg_api.annealing_options "
+                    "WHERE key = 'automation_level'"
+                )
+                row = cur.fetchone()
+                if row:
+                    automation_level = row[0]
+        finally:
+            client.pool.putconn(conn)
+    except Exception:
+        pass  # Fall back to default
+
     job_id = queue.enqueue(
         job_type="ontology_annealing",
         job_data={
             "demotion_threshold": demotion_threshold,
             "promotion_min_degree": promotion_min_degree,
             "max_proposals": max_proposals,
+            "automation_level": automation_level,
             "dry_run": dry_run,
             "triggered_by": current_user.username if current_user else "api",
         }
--- a/api/app/services/job_scheduler.py
+++ b/api/app/services/job_scheduler.py
@@ -197,11 +197,12 @@ class JobScheduler:
         if deleted_failed > 0:
             logger.info(f"Deleted {deleted_failed} old failed jobs")
 
-        # Task 4: Expire stale annealing proposals past their TTL
+        # Task 4: Annealing proposal lifecycle cleanup
         try:
             conn = queue._get_connection()
             try:
                 with conn.cursor() as cur:
+                    # 4a. Expire stale pending proposals past their TTL
                     cur.execute("""
                         UPDATE kg_api.annealing_proposals
                         SET status = 'expired'
@@ -210,13 +211,50 @@ class JobScheduler:
                           AND expires_at < NOW()
                     """)
                     expired_proposals = cur.rowcount
+
+                    # 4b. Re-dispatch stranded approved proposals (approved
+                    # but no execution job dispatched, e.g. enqueue failed)
+                    cur.execute("""
+                        SELECT id FROM kg_api.annealing_proposals
+                        WHERE status = 'approved'
+                          AND executed_at IS NULL
+                          AND reviewed_at < NOW() - INTERVAL '5 minutes'
+                    """)
+                    stranded = cur.fetchall()
+
                 conn.commit()
+
                 if expired_proposals > 0:
                     logger.info(f"Expired {expired_proposals} stale annealing proposals")
+
+                # Re-dispatch stranded proposals outside the cursor
+                if stranded:
+                    for (proposal_id,) in stranded:
+                        try:
+                            exec_job_id = queue.enqueue(
+                                job_type="proposal_execution",
+                                job_data={
+                                    "proposal_id": proposal_id,
+                                    "triggered_by": "scheduler_recovery",
+                                },
+                            )
+                            queue.update_job(exec_job_id, {
+                                "status": "approved",
+                                "approved_by": "scheduler_recovery",
+                            })
+                            queue.execute_job_async(exec_job_id)
+                            logger.info(
+                                f"Re-dispatched stranded proposal {proposal_id} "
+                                f"as job {exec_job_id}"
+                            )
+                        except Exception as e:
+                            logger.error(
+                                f"Failed to re-dispatch proposal {proposal_id}: {e}"
+                            )
             finally:
                 queue._return_connection(conn)
         except Exception as e:
-            logger.debug(f"Proposal expiry check skipped: {e}")
+            logger.warning(f"Proposal lifecycle cleanup skipped: {e}")
 
     async def manual_cleanup(self) -> Dict[str, int]:
         """
--- a/api/app/workers/annealing_worker.py
+++ b/api/app/workers/annealing_worker.py
@@ -154,7 +154,8 @@ def _auto_approve_and_dispatch(
                         UPDATE kg_api.annealing_proposals
                         SET status = 'approved',
                             reviewed_by = 'annealing_worker',
-                            reviewed_at = %s
+                            reviewed_at = %s,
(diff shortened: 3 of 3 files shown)


aaronsb:
## Code Review: Autonomous Annealing Proposal Execution

**PR size:** ~260 lines added across 7 files. Standard review scope.

---

### 1. Correctness: Auto-Approve Path vs. Human Review Path

The autonomous path in `_auto_approve_and_dispatch()` (`api/app/workers/annealing_worker.py:134-197`) correctly mirrors the human review endpoint (`api/app/routes/ontology.py:373-474`). Both follow the same sequence:

1. `UPDATE ... SET status = 'approved'` with reviewer metadata
2. `job_queue.enqueue(job_type="proposal_execution", ...)`
3. `job_queue.update_job(exec_job_id, {"status": "approved", ...})`
4. `job_queue.execute_job_async(exec_job_id)`

This is good alignment. The execution pipeline (`proposal_execution_worker.py` -> `_claim_proposal` with `WHERE status = 'approved'` -> `ProposalExecutor`) is shared and unchanged.

**One minor omission:** The human review path stores `reviewer_notes` (ontology.py:416). The autonomous path does not set `reviewer_notes`. Not a bug -- there are no notes to store -- but worth noting for audit trail completeness. A value like `'auto-approved (autonomous mode)'` in `reviewer_notes` would make the audit trail more self-documenting when reviewing proposals after the fact. Low priority.

---

### 2. Race Conditions

**Well handled.** The atomic `UPDATE ... WHERE id = %s AND status = 'pending' RETURNING id` (annealing_worker.py:153-159) prevents double-dispatch even if a human reviews via API concurrently with the autonomous worker. The `_claim_proposal` in `proposal_execution_worker.py:176-201` provides a second guard at execution time (`WHERE status = 'approved'`). This is defense-in-depth -- two atomic guards in the pipeline.

**One subtle timing concern:** Between the `INSERT` in `_store_proposal` (annealing_manager.py:389-403) and the `UPDATE ... SET status = 'approved'` (annealing_worker.py:153-159), the proposal is briefly `pending`. In autonomous mode, this window is milliseconds. However, if the job scheduler's proposal expiry task (job_scheduler.py:200-219) runs during a very long annealing cycle (many candidates, slow LLM), it could theoretically expire a proposal that the autonomous worker is about to approve. In practice: the TTL is 7 days (from `expires_at DEFAULT NOW() + INTERVAL '7 days'` in migration 046), so this is not a real concern. Just documenting that the invariant holds because of the generous TTL.

---

### 3. Error Handling: Partial Dispatch Failures

`_auto_approve_and_dispatch` (annealing_worker.py:148-195) wraps each proposal in its own try/except and continues to the next on failure. This is correct -- a single proposal failure should not prevent the remaining proposals from executing.

**However**, when `job_queue.enqueue` or `job_queue.execute_job_async` fails after the proposal has already been marked `approved` in the database (annealing_worker.py:171-182), the proposal is left in `approved` status with no execution job dispatched. The `proposal_execution_worker._claim_proposal` won't run because no job was enqueued for it.

This is partially mitigated by the existing `process_approved_jobs` recovery in `job_scheduler.py:97-131`, which picks up stuck approved jobs. But that mechanism finds stuck *jobs*, not stuck *proposals*. The proposal sits at `approved` with no corresponding job in the queue.

**Suggestion:** This is an edge case (enqueue failure after DB update), and the 7-day TTL expiry provides a safety net. But for operational visibility, logging the specific failure as a warning (already done at line 189-192) is sufficient. If this becomes a recurring issue, a periodic sweep for `approved` proposals with no corresponding execution job would close the gap.

---

### 4. Migration Safety (053)

```sql
UPDATE kg_api.annealing_options
SET value = 'autonomous', description = '...'
WHERE key = 'automation_level' AND value = 'hitl';
```

The `AND value = 'hitl'` guard is idempotent and safe. If someone already changed the value to something else (e.g., `'autonomous'` manually, or a custom value), the UPDATE is a no-op and the verification block still passes (it only checks that the key exists, not the specific value).

The verification block:
```sql
IF v_level IS NULL THEN
    RAISE EXCEPTION 'Migration failed: automation_level option not found';
END IF;
```

This correctly validates that the `automation_level` key exists (which migration 049 guarantees). It does *not* assert the value is `autonomous`, so re-running the migration after a manual override to `hitl` won't fail. Good.

**One consideration for existing installs:** The migration changes the default for *existing* installs from `hitl` to `autonomous`. This means after `./operator.sh upgrade`, annealing proposals will start auto-executing. The migration comment explains the rationale well. The documentation in `SCHEDULED-JOBS.md` line 117 provides the revert path: `UPDATE kg_api.annealing_options SET value = 'hitl' WHERE key = 'automation_level';`. This is transparent behavior.

---

### 5. Scheduler Proposal Expiry (job_scheduler.py:200-219)

**Connection pool usage:** The code uses `queue._get_connection()` / `queue._return_connection()` (private methods of JobQueue). This is the same pattern used elsewhere in `job_scheduler.py` -- Tasks 1-3 use the queue's public API (`list_jobs`, `update_job`, `delete_job`), but Task 4 needs raw SQL because `JobQueue` has no proposal-aware methods.

Accessing private methods (`_get_connection`, `_return_connection`) is a code smell but pragmatic here -- adding a public method to JobQueue for a single use case would be over-engineering. The `try/finally` ensures the connection is returned even on error.

**The `except` handler:** Using `logger.debug` (line 219) for the expiry skip is appropriate -- this task is additive (cleanup), and failure should not propagate. However, `logger.debug` might be too quiet for production. If the `annealing_proposals` table doesn't exist yet (fresh install before migration 046), this will silently fail at debug level. Consider `logger.info` or `logger.warning` on first failure.

---

### 6. Documentation Accuracy (SCHEDULED-JOBS.md)

The mermaid sequence diagram (lines 122-167) accurately reflects the code flow:

- Trigger -> Launcher -> atomic epoch claim -> enqueue -> Worker -> Manager -> score/centroid/edge/candidate/LLM loop -> proposals
- Autonomous branch: Worker -> approve -> enqueue execution -> dispatch
- HITL branch: proposals await review

The diagram shows `Queue->>Queue: execute_job_async -> ProposalExecutor` which correctly represents the in-process async dispatch. The decision criteria table matches the code defaults (`protection_score < 0.15`, `degree >= 10`). The configure-via SQL statement is correct.

**Minor doc issue:** The manual trigger endpoint docstring at `ontology.py:487` still says "ADR-200 Phase 3b" and "generates proposals for human review" -- this should be updated to reflect autonomous mode. Not in the diff for this PR (pre-existing), but worth a follow-up.

---

### 7. SOLID and Code Quality

**Single Responsibility:** Clean separation. `AnnealingManager` handles scoring/candidates/proposals. `annealing_worker` handles job orchestration and the autonomous dispatch decision. `ProposalExecutor` handles execution. The autonomous logic is added at the correct layer (worker, not manager).

**Open/Closed:** The `automation_level` parameter flows through `job_data` as a string, making it easy to add new modes (e.g., the `aitl` mentioned in migration 049's description) without modifying existing code paths. The `if/elif` pattern is acceptable for 2 modes.

**Variable rename in annealing_manager.py:** The rename from `proposal` to `proposal_id` (lines 138-155) is a clarity improvement. `_store_proposal` always returned `Optional[int]` (the ID), so the old name `proposal` was misleading. Good cleanup.

**Manual trigger gap:** The API endpoint `trigger_annealing_cycle` (ontology.py:477-515) does not pass `automation_level` in `job_data`. The worker defaults to `autonomous` via `job_data.get("automation_level", "autonomous")` (annealing_worker.py:92). This means manual API triggers always run in autonomous mode regardless of the database setting. The launcher path correctly reads `automation_level` from `annealing_options` and passes it through. This inconsistency means: cron/post-ingestion triggers respect the DB config, but manual API calls do not. Consider reading the DB setting in the route handler or passing it through, so that a user who sets `hitl` mode gets consistent behavior regardless of trigger source.

---

### Summary

This is a well-structured change. The execution path alignment with the human review endpoint is correct, the atomic UPDATE prevents double-dispatch, the migration is idempotent and safe, and the documentation is accurate. The separation of concerns is clean.

**Items to consider (none blocking):**

1. **Manual API trigger does not respect DB `automation_level` setting** -- it defaults to `autonomous` regardless. This is the most actionable finding.
2. **Approved-but-no-job edge case** -- if enqueue fails after proposal approval, the proposal is stranded. Mitigated by TTL expiry but not by automatic recovery.
3. **`reviewer_notes` not set in autonomous path** -- minor audit trail gap.
4. **`logger.debug` for expiry errors** -- may be too quiet for detecting table-missing scenarios.

PR #318, opened 2026-02-11
feat(fuse): config-driven daemon mode with systemd-first default

### LINKED ISSUE #317: feat(fuse): config-driven daemon mode with systemd-first default
## Summary

`kg-fuse mount` should default to user-space systemd service management instead of always double-forking a daemon process. The daemon mode is config-driven with smart detection.

## Config

New field in `fuse.json`: `daemon_mode` — either `"systemd"` or `"daemon"`.

### New installs (`kg-fuse init`)
- Auto-detect systemd availability (`systemctl` binary + `/run/systemd/system`)
- If systemd available → set `daemon_mode: "systemd"` (default)
- If not → set `daemon_mode: "daemon"`

### `kg-fuse mount`
- **systemd mode**: install unit if needed, `systemctl --user start kg-fuse`
- **daemon mode**: existing double-fork behavior
- If config says `daemon` but systemd is detected → print info: "systemd is available for use (see `kg-fuse init`)"
- If config says `systemd` but systemd not available → error, suggest switching to `daemon`

### User overrides
- If user explicitly chooses `daemon` when systemd is available → print recommendation message but respect the choice, save config
- Config is the single source of truth for mode

### `kg-fuse reset` (new command)
- **systemd mode**: `systemctl --user restart kg-fuse`
- **daemon mode**: kill existing PID, start new daemon

### `kg-fuse unmount`
- **systemd mode**: `systemctl --user stop kg-fuse`
- **daemon mode**: existing SIGTERM + fusermount behavior

## Non-goals
- No changes to `--foreground` mode (always runs directly for debugging)
- No removal of daemon mode (still needed for non-systemd systems)

aaronsb:
## Summary

Closes #317

- `kg-fuse mount` now delegates to `systemctl --user start kg-fuse` when systemd is available, instead of always double-forking a daemon process
- New `daemon_mode` config field in `fuse.json` (`"systemd"` or `"daemon"`) — auto-detected on first use, persisted for future runs
- Systemd unit updated to `Type=simple` with `ExecStart=kg-fuse mount --foreground` (proper lifecycle management, restart-on-failure, journal integration)
- New `kg-fuse reset` command (systemctl restart or kill+remount depending on mode)
- `client-manager.sh` updated to delegate all FUSE lifecycle to `kg-fuse` CLI commands instead of reimplementing its own parallel management
- 21 new tests added (241 total, all passing)

## Files changed

| File | Change |
|------|--------|
| `fuse/kg_fuse/config.py` | `daemon_mode` field + `set_daemon_mode()` |
| `fuse/kg_fuse/safety.py` | `Type=simple` unit, `systemd_start/stop/restart()` |
| `fuse/kg_fuse/cli.py` | `_resolve_daemon_mode()`, routed mount/unmount/init/reset/status |
| `fuse/kg_fuse/main.py` | Added `reset` subcommand |
| `client-manager.sh` | Delegate to `kg-fuse mount/unmount/init` instead of raw fusermount/nohup/pgrep |
| `fuse/tests/test_*.py` | 21 new tests covering daemon_mode, systemd lifecycle, config persistence |

## Test plan

- [x] `make test-fuse` — 241 tests pass
- [ ] Manual: `kg-fuse init ~/Knowledge` on systemd system → sets `daemon_mode: "systemd"`
- [ ] Manual: `kg-fuse mount` → `systemctl --user start kg-fuse`
- [ ] Manual: `kg-fuse status` → shows systemd service state
- [ ] Manual: `kg-fuse reset` → restarts via systemctl
- [ ] Manual: Edit fuse.json to `daemon_mode: "daemon"` → mount forks + prints systemd hint

### DIFF
--- a/client-manager.sh
+++ b/client-manager.sh
@@ -494,6 +494,8 @@ linux_stop_fuse() {
             fusermount3 -u "$mount_dir" 2>/dev/null || true
         elif command -v fusermount &>/dev/null; then
             fusermount -u "$mount_dir" 2>/dev/null || true
+        else
+            umount "$mount_dir" 2>/dev/null || true
         fi
         return
     fi
--- a/fuse/kg_fuse/cli.py
+++ b/fuse/kg_fuse/cli.py
@@ -1,7 +1,7 @@
 """
 Subcommand implementations for kg-fuse CLI.
 
-Commands: init, mount, unmount, status, config, repair, update.
+Commands: init, mount, unmount, status, config, reset, repair, update.
 """
 
 import json
@@ -117,6 +117,28 @@ def _resolve_daemon_mode(config: FuseConfig) -> str:
     set_daemon_mode(mode)
     config.daemon_mode = mode
     return mode
+
+
+def _offer_rc_autostart() -> None:
+    """Offer to add kg-fuse mount to shell RC file for login autostart."""
+    rc_installed = False
+    shell_info = detect_shell()
+    if shell_info:
+        shell_name, rc_path = shell_info
+        if _prompt_yn(f"Add auto-mount to {rc_path}?"):
+            mount_line = "command -v kg-fuse >/dev/null && kg-fuse mount"
+            ok, msg = add_to_rc(rc_path, mount_line)
+            print(f"  {msg}")
+            if ok:
+                rc_installed = True
+
+    if rc_installed:
+        print(f"\n{_green('Ready!')} Mount with:")
+        print(f"  {_bold('kg-fuse mount')}")
+        print(f"\nWill auto-mount on shell login.")
+    else:
+        print(f"\n{_green('Ready!')} Mount with:")
+        print(f"  {_bold('kg-fuse mount')}")
 
 
 def _test_api(api_url: str) -> tuple[bool, str]:
@@ -433,7 +455,9 @@ def cmd_init(args: Namespace) -> None:
 
             if _systemd_unit_enabled():
                 print(f"  Systemd unit already enabled — updating unit file.")
-                install_systemd_unit(kg_fuse_path, enable=True)
+                ok, msg = install_systemd_unit(kg_fuse_path, enable=True)
+                if not ok:
+                    print(f"  {_red(msg)}")
             else:
                 ok, msg = install_systemd_unit(kg_fuse_path, enable=True)
                 if ok:
@@ -459,48 +483,10 @@ def cmd_init(args: Namespace) -> None:
             print(f"  You can switch later by setting \"daemon_mode\": \"systemd\" in")
             print(f"  {get_fuse_config_path()} or by running {_bold('kg-fuse init')} again.\n")
 
-            # Offer RC file autostart as fallback
-            rc_installed = False
-            shell_info = detect_shell()
-            if shell_info:
-                shell_name, rc_path = shell_info
-                if _prompt_yn(f"Add auto-mount to {rc_path}?"):
-                    mount_line = "command -v kg-fuse >/dev/null && kg-fuse mount"
-                    ok, msg = add_to_rc(rc_path, mount_line)
-                    print(f"  {msg}")
-                    if ok:
-                        rc_installed = True
-
-            if rc_installed:
-                print(f"\n{_green('Ready!')} Mount with:")
-                print(f"  {_bold('kg-fuse mount')}")
-                print(f"\nWill auto-mount on shell login.")
-            else:
-                print(f"\n{_green('Ready!')} Mount with:")
-                print(f"  {_bold('kg-fuse mount')}")
+            _offer_rc_autostart()
     else:
         set_daemon_mode("daemon")
-
-        # No systemd — offer RC file autostart
-        rc_installed = False
-        shell_info = detect_shell()
-        if shell_info:
-            shell_name, rc_path = shell_info
-            if _prompt_yn(f"Add auto-mount to {rc_path}?"):
-                mount_line = "command -v kg-fuse >/dev/null && kg-fuse mount"
-                ok, msg = add_to_rc(rc_path, mount_line)
-                print(f"  {msg}")
-                if ok:
-                    rc_installed = True
-
-        if rc_installed:
-            print(f"\n{_green('Ready!')} Mount with:")
-            print(f"  {_bold('kg-fuse mount')}")
-            print(f"\nWill auto-mount on shell login.")
-        else:
-            print(f"\n{_green('Ready!')} Mount with:")
-            print(f"  {_bold('kg-fuse mount')}")
-            print(f"  {_dim('kg-fuse mount                # mounts all configured')}")
+        _offer_rc_autostart()
     print()
 
 
@@ -534,6 +520,11 @@ def cmd_mount(args: Namespace) -> None:
                 print(f"Switch to daemon mode:")
                 print(f"  Edit {get_fuse_config_path()} and set \"daemon_mode\": \"daemon\"")
                 sys.exit(1)
+
+            if mountpoint:
+                print(f"{_yellow('Note:')} systemd mode starts all configured mounts via the service unit.")
+                print(f"  Specific mountpoint '{mountpoint}' argument is ignored.")
+                print(f"  To mount a single path, use: kg-fuse mount --foreground {mountpoint}\n")
 
             # Ensure unit is installed and up-to-date
             unit_path = get_systemd_unit_path()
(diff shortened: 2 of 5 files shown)


aaronsb:
## Code Review: feat(fuse): config-driven daemon mode with systemd-first default

**PR Size:** 589 additions / 225 deletions across 8 files -- standard-to-thorough review tier.

---

### Overall Assessment

This is a well-structured change that consolidates daemon lifecycle management behind the `kg-fuse` CLI, eliminating the parallel management logic that `client-manager.sh` previously maintained. The `daemon_mode` config field is a clean abstraction that unifies the systemd-vs-fork decision into a single, persisted setting. The `Type=simple` + `--foreground` approach is the correct systemd pattern for a FUSE driver.

Test coverage is solid: 21 new tests covering config persistence, daemon mode resolution, systemd unit generation, and lifecycle operations.

---

### Findings

#### [Major] `set_daemon_mode()` has no input validation

**Location:** `fuse/kg_fuse/config.py:350-354`

**Problem:** `set_daemon_mode()` accepts any string and writes it to `fuse.json`. A typo like `set_daemon_mode("systemD")` or `set_daemon_mode("fork")` would silently persist an unrecognized value. Then `_resolve_daemon_mode()` would re-auto-detect on every invocation because the value doesn't match `("systemd", "daemon")`, causing a write-on-every-read loop.

**Why it matters:** The auto-detect fallback path writes to disk on each call. An invalid persisted value creates a subtle write amplification bug -- every `kg-fuse status`, `kg-fuse mount`, etc. would re-detect and re-write.

**Suggestion:** Add validation in `set_daemon_mode()`:

```python
VALID_DAEMON_MODES = ("systemd", "daemon")

def set_daemon_mode(mode: str) -> None:
    """Set daemon_mode in fuse.json. Creates the file if needed."""
    if mode not in VALID_DAEMON_MODES:
        raise ValueError(f"Invalid daemon_mode {mode!r}, expected one of {VALID_DAEMON_MODES}")
    fuse_data = read_fuse_config() or {}
    fuse_data["daemon_mode"] = mode
    write_fuse_config(fuse_data)
```

And add a test for the rejection case.

---

#### [Major] `cmd_mount` in systemd mode ignores `--mountpoint` argument

**Location:** `fuse/kg_fuse/cli.py:528-551`

**Problem:** When `daemon_mode == "systemd"` and `foreground is False`, the function calls `systemd_start()` and returns immediately, without regard for whether the user passed a specific `mountpoint` argument. Running `kg-fuse mount /mnt/specific` in systemd mode would start the entire systemd service (which mounts ALL configured mounts), ignoring the user's intent to mount only one path.

**Why it matters:** This is a silent behavior change from the user's perspective. They asked to mount one path but got all of them. There is no warning or feedback about the discrepancy.

**Suggestion:** When `mountpoint` is specified in systemd mode, either:
1. Fall through to the daemon path for single-mount operations, or
2. Print a clear message explaining that systemd manages all mounts together, e.g.:
```
"Note: in systemd mode, all configured mounts are managed together. Starting service..."
```

---

#### [Minor] `cmd_init` discards `install_systemd_unit` return value on the "already enabled" path

**Location:** `fuse/kg_fuse/cli.py:434-436`

**Problem:** When `_systemd_unit_enabled()` is true, the code calls `install_systemd_unit(kg_fuse_path, enable=True)` to update the unit file, but discards both `ok` and `msg`. On the `else` branch (line 438), the return value is correctly checked.

```python
if _systemd_unit_enabled():
    print(f"  Systemd unit already enabled - updating unit file.")
    install_systemd_unit(kg_fuse_path, enable=True)  # <-- return value discarded
else:
    ok, msg = install_systemd_unit(kg_fuse_path, enable=True)
    if ok:
        print(f"  {_green(msg)}")
    else:
        print(f"  {_red(msg)}")
```

**Why it matters:** If `daemon-reload` or `enable` fails during the update, the user gets no feedback. They see "already enabled - updating" but never learn the update failed.

**Suggestion:** Handle the return value consistently on both branches. Either always print the result, or at minimum log a warning on failure.

---

#### [Minor] `cmd_unmount` in systemd mode always does fusermount cleanup

**Location:** `fuse/kg_fuse/cli.py:706-716`

**Problem:** After `systemd_stop()` succeeds, the code unconditionally calls `fusermount_unmount()` and `clear_pid()` on every configured mount. If systemd cleanly stopped the process (which itself calls `pyfuse3.close(unmount=True)`), these fusermount calls will fail with "not mounted" errors for every mount. The output of those failures is suppressed, but the `clear_pid()` calls remove PID files that may not exist, which is benign but noisy in debug logging.

**Why it matters:** Cosmetic -- the safety belt is reasonable. But it may produce confusing output or log noise in the common case (systemd stops cleanly).

**Suggestion:** Only run the fusermount cleanup if `systemd_stop()` fails, or check mount status before cleanup:

```python
if daemon_mode == "systemd" and not mountpoint and has_systemd():
    ok, msg = systemd_stop()
    ...
    # Only clean up if service stop reported failure
    if not ok:
        for mount_path in config.mounts:
            fusermount_unmount(mount_path)
            clear_pid(mount_path)
    return
```

---

#### [Minor] Duplicated RC-file autostart block in `cmd_init`

**Location:** `fuse/kg_fuse/cli.py:462-503`

**Problem:** The RC-file autostart logic (detect shell, prompt, add_to_rc, print "Ready!") is duplicated nearly verbatim between the "systemd declined" branch (lines 462-480) and the "no systemd" branch (lines 481-503). The only difference is the trailing hint line on the no-systemd path.

**Why it matters:** Violates DRY. If the RC autostart flow changes (different prompt text, new shell support), it needs updating in two places.

**Suggestion:** Extract a helper like `_offer_rc_autostart() -> bool` that encapsulates the detect/prompt/add/print cycle. Both branches would call it.

---

#### [Minor] `cmd_reset` daemon-mode path does not check auth credentials

**Location:** `fuse/kg_fuse/cli.py:921-960`

**Problem:** `cmd_reset` in daemon mode calls `_start_mount()` which eventually needs valid `client_id` and `client_secret`. But `cmd_reset` loads config via `load_config()` without CLI credential overrides, and never checks if credentials are present before attempting re-mount. If credentials are missing (e.g., after an `kg oauth create` rotation), all re-mount attempts will fail with confusing errors from deep in the mount stack.

**Why it matters:** `cmd_mount` has an explicit credential check at the top (lines 515-521). `cmd_reset` bypasses that.

**Suggestion:** Either reuse the credential validation from `cmd_mount`, or add a similar check at the top of `cmd_reset`.

---

#### [Nit] Module docstring in `cli.py` is stale

**Location:** `fuse/kg_fuse/cli.py:1-5`

**Problem:** The docstring reads "Commands: init, mount, unmount, status, config, repair, update" but `reset` is now a command too.

**Suggestion:** Add `reset` to the list.

---

#### [Nit] `client-manager.sh` fallback drops `umount` fallback

**Location:** `client-manager.sh`, `linux_stop_fuse()` around line 488

**Problem:** The old code had a `umount "$mount_dir"` fallback when neither `fusermount` nor `fusermount3` was available. The new fallback path (when `kg-fuse` is not installed) drops this. On minimal systems without fusermount, the old code at least tried `umount`; now it silently does nothing.

**Why it matters:** Edge case. On minimal containers or chroots without fusermount installed, the fallback would silently fail. Low impact since such systems are unlikely to run FUSE at all.

**Suggestion:** Consider restoring the `umount` fallback in the `kg-fuse not found` path, or document that fusermount is now a hard requirement for the fallback path.

---

### What looks good

- **Type=simple + --foreground** is the correct systemd pattern. The old `Type=forking` approach was fragile because systemd's forking detection has timing issues with FUSE mounts.
- **Auto-detect with persist** (`_resolve_daemon_mode`) is a smart UX choice -- users don't need to configure this unless they want to override it.
- **Separation of install vs enable** in `install_systemd_unit(enable=bool)` is a clean API improvement over the old `--now` approach.
- **client-manager.sh delegation** is the right call -- having two parallel lifecycle managers was a maintenance burden and a source of mode-mismatch bugs.
- **Test coverage** is well-targeted: unit generation content, lifecycle happy/sad paths, config persistence, and resolution logic all have tests.
- **Error handling** on the systemd subprocess calls consistently catches both `TimeoutExpired` and `FileNotFoundError`.

---

### Test coverage gaps to consider

1. **`cmd_reset` daemon mode** -- no test for the kill+refork path. The systemd path in `cmd_reset` is indirectly tested, but the daemon fallback path that calls `_stop_mount` + `_start_mount` is untested.
2. **`cmd_mount` systemd routing** -- no test verifying that `cmd_mount` delegates to `systemd_start()` when `daemon_mode == "systemd"`.
3. **`cmd_unmount` systemd routing** -- no test for the systemd stop + fusermount cleanup path.
4. **Invalid `daemon_mode` values** -- what happens if fuse.json contains `"daemon_mode": "garbage"`? Currently auto-detects on every call (see Major finding above).

---

PR #333, opened 2026-02-11
ux: improve MCP tool descriptions, label resolution, and queue defaults

### LINKED ISSUE #326: ux: connect tool should indicate structural fallback paths
## Source

Agent UX report from live MCP session (2026-02-10, 316 concepts / 743 relationships).

## Problem

When `connect` finds no semantic paths between concepts, it returns structural Source→Ontology→Source paths without indicating they're fallback results. An agent spent 20+ minutes and 8 redundant `connect` calls not understanding why it kept getting the same structural paths, thinking manually-created edges weren't being stored.

The agent also discovered that manually-created edges (via graph edit) appear in `concept details` relationships and are findable via Cypher, but `connect` doesn't traverse them because they lack traversal weight in a young graph.

## Recommendations

1. Add metadata to connect response indicating path type: `"path_type": "semantic"` vs `"path_type": "structural_fallback"`
2. When no semantic paths found, include a message: "No semantic paths found within max_hops; showing structural fallback"
3. Consider an `include_all_edges: true` parameter that forces the pathfinder to consider all edge types equally (including manually created edges with no traversal history)

## Impact

This was the #1 friction point in the session. A single metadata field would have saved significant wasted tool calls and agent confusion.

### LINKED ISSUE #327: ux: lower label resolution threshold or add fuzzy matching
## Source

Agent UX report from live MCP session (2026-02-10).

## Problem

Graph edit's semantic label resolution (`from_label`/`to_label`) uses an 85% similarity threshold. Labels that feel unambiguous to an agent — "Multipolar Trap", "Operational Moat" — fail to resolve, requiring fallback to hash IDs. This breaks the creation flow every time.

## Current behavior

- Label resolution at 85% threshold → hard failure: "No concept found matching label"
- Agent must go back to search, find the concept ID, copy the hash, switch to `from_concept_id`

## Recommendations (pick one or combine)

1. **Lower threshold** to 75% — catches more unambiguous matches
2. **Fuzzy "did you mean?"** — on failure, return top match with similarity: `"Did you mean 'Multipolar Trap Dynamics' (72% match)?"`
3. **Top-1 with confirmation** — resolve to top match and include what it matched in the response so the agent can verify

## Context

The agent notes: "I use labels for creation and exploration, and the system returns the resolved IDs in responses so I can use them for subsequent precise operations. This already works — I just need label resolution to be more forgiving."

### LINKED ISSUE #328: ux: reframe program tool as primary reasoning tool in MCP descriptions
## Source

Agent UX report from live MCP session (2026-02-10).

## Problem

The agent made ~8 redundant `connect` calls before the user pointed it to the `program` tool. Once discovered, it became the most productive tool in the session. But the MCP tool descriptions position `search`/`connect`/`related` as primary and `program` reads like an advanced/admin feature.

## Current state

PR #303 (Phase 3b) already revised agent guidance to present three tiers and add program cross-references. But the agent still didn't discover it organically — suggesting the guidance isn't prominent enough or the framing still doesn't convey when to switch.

## Recommendation

Make the cognitive mode distinction explicit in tool descriptions:

> "Use search/connect/related for quick associative exploration (System 1). Use program for deliberate multi-step reasoning (System 2) — when reflexive tools aren't giving you what you need."

The agent's productive loop was: search → details → connect → **program/Cypher when connect isn't enough** → graph edit → program to verify. The anti-pattern was repeating connect hoping for different results.

## Key quote

> "The most important thing I learned is that this platform has two cognitive modes, and I spent too long in the wrong one."

### LINKED ISSUE #329: ux: default continue_on_error=true for graph edit queue
## Source

Agent UX report from live MCP session (2026-02-10).

## Problem

Queue of 8 edge creations stopped at operation 4 when one label resolution failed. Operations 5-8 never executed. Agent had to re-submit remaining operations in a second call and lost track of what was actually created.

## Current behavior

- `continue_on_error` defaults to `false`
- On first error, remaining operations are skipped
- Response only includes results up to the failure

## Recommendation

Either:
1. **Default `continue_on_error: true`** for queue operations — partial success is more useful than total abort for batch graph edits
2. Or at minimum, **return all results** (including which succeeded before the error) even when stopping, so the agent knows what state the graph is in

### LINKED ISSUE #330: ux: add AGE Cypher syntax examples to program tool description
## Source

Agent UX report from live MCP session (2026-02-10).

## Problem

Agent needed 3 attempts to write valid Cypher for the program tool:

1. \`\$W_IDS\` reference failed ("parameters argument is missing")
2. Pipe-separated relationship types \`[r:PREVENTS|CONTRASTS_WITH]\` failed ("syntax error at |")
3. \`WHERE type(r) IN ['PREVENTS', 'CONTRASTS_WITH']\` worked

Apache AGE's openCypher subset differs from Neo4j's Cypher in ways that aren't discoverable without trial and error. The `program/syntax` resource exists but the agent didn't read it initially.

## Recommendation

Add a few high-value syntax examples directly in the program tool description (not just in the resource):

- `WHERE type(r) IN ['TYPE_A', 'TYPE_B']` instead of `[r:TYPE_A|TYPE_B]`
- `WHERE c.concept_id IN $W_IDS` for working graph references
- Common AGE-specific patterns that differ from Neo4j

The resource is the right place for the full reference, but the most common gotchas should be in the tool description where agents see them immediately.

aaronsb:
## Summary

Addresses five agent UX friction points identified in a live MCP session report (2026-02-10):

- **#328** — Reframe program tool as primary reasoning escalation path with System 1/System 2 framing
- **#330** — Add AGE Cypher syntax gotchas directly in program tool description (WHERE type(r) vs inline `[r:TYPE|TYPE]`)
- **#326** — Add actionable guidance when connect finds no paths (suggestions to escalate to program, increase max_hops)
- **#329** — Default `continue_on_error=true` for graph edit queue (partial success > total abort for batch ops)
- **#327** — Lower label resolution threshold from 85% to 75% with "Did you mean?" suggestions for near-misses (60-75%)

## Changes

- `cli/src/mcp-server.ts` — tool descriptions for search, concept, graph, program; queue default flip
- `cli/src/mcp/formatters/concept.ts` — connect no-path message with escalation hints
- `cli/src/mcp/graph-operations.ts` — `resolveConceptByLabel()` threshold lowered, fuzzy suggestions added, `executeQueue` default signature aligned
- `cli/tests/mcp/graph-operations.test.ts` — 8 unit tests for label resolution + queue defaults
- `docs/reference/mcp/` — auto-generated docs updated

## Test plan

- [x] `npm run build` — TypeScript compiles clean
- [x] `npm test` — 46/46 tests pass (8 new)
- [x] Manual: graph queue with bad label continues past errors by default
- [x] Manual: fuzzy label resolution accepts 75%+ matches, suggests 60-75% near-misses
- [x] Manual: connect with no paths shows guidance message

Closes #326, closes #327, closes #328, closes #329, closes #330

### DIFF
--- a/cli/tests/mcp/graph-operations.test.ts
+++ b/cli/tests/mcp/graph-operations.test.ts
@@ -1,0 +1,172 @@
+/**
+ * GraphOperationExecutor unit tests
+ *
+ * Tests label resolution thresholds and queue execution defaults.
+ */
+
+import { GraphOperationExecutor } from '../../src/mcp/graph-operations';
+import type { KnowledgeGraphClient } from '../../src/api/client';
+import type { SearchResponse } from '../../src/types';
+
+/** Build a minimal mock client with only the methods under test. */
+function mockClient(overrides: Partial<KnowledgeGraphClient> = {}): KnowledgeGraphClient {
+  return {
+    searchConcepts: jest.fn(),
+    ...overrides,
+  } as unknown as KnowledgeGraphClient;
+}
+
+/** Build a SearchResponse with the given results. */
+function searchResponse(
+  results: Array<{ concept_id: string; label: string; score: number }>
+): SearchResponse {
+  return {
+    query: 'test',
+    count: results.length,
+    results: results.map(r => ({
+      ...r,
+      documents: [],
+      evidence_count: 0,
+    })),
+  };
+}
+
+describe('GraphOperationExecutor', () => {
+  describe('resolveConceptByLabel', () => {
+    it('should resolve when score >= 0.75', async () => {
+      const client = mockClient({
+        searchConcepts: jest.fn().mockResolvedValue(
+          searchResponse([{ concept_id: 'c_abc', label: 'CAP Theorem', score: 0.82 }])
+        ),
+      });
+      const executor = new GraphOperationExecutor(client);
+
+      const result = await executor.resolveConceptByLabel('CAP Theorem');
+
+      expect(result).toBe('c_abc');
+      expect(client.searchConcepts).toHaveBeenCalledWith(
+        expect.objectContaining({ min_similarity: 0.6, limit: 3 })
+      );
+    });
+
+    it('should resolve at exactly 0.75 threshold', async () => {
+      const client = mockClient({
+        searchConcepts: jest.fn().mockResolvedValue(
+          searchResponse([{ concept_id: 'c_edge', label: 'Boundary Concept', score: 0.75 }])
+        ),
+      });
+      const executor = new GraphOperationExecutor(client);
+
+      const result = await executor.resolveConceptByLabel('Boundary Concept');
+
+      expect(result).toBe('c_edge');
+    });
+
+    it('should throw "did you mean?" when score is between 0.60 and 0.75', async () => {
+      const client = mockClient({
+        searchConcepts: jest.fn().mockResolvedValue(
+          searchResponse([
+            { concept_id: 'c_near', label: 'Multipolar Trap Dynamics', score: 0.72 },
+            { concept_id: 'c_alt', label: 'Multipolar Systems', score: 0.65 },
+          ])
+        ),
+      });
+      const executor = new GraphOperationExecutor(client);
+
+      await expect(executor.resolveConceptByLabel('Multipolar Trap'))
+        .rejects.toThrow('Did you mean');
+
+      await expect(executor.resolveConceptByLabel('Multipolar Trap'))
+        .rejects.toThrow('c_near');
+    });
+
+    it('should include up to 3 suggestions in "did you mean?" error', async () => {
+      const client = mockClient({
+        searchConcepts: jest.fn().mockResolvedValue(
+          searchResponse([
+            { concept_id: 'c_1', label: 'Alpha', score: 0.72 },
+            { concept_id: 'c_2', label: 'Beta', score: 0.68 },
+            { concept_id: 'c_3', label: 'Gamma', score: 0.62 },
+          ])
+        ),
+      });
+      const executor = new GraphOperationExecutor(client);
+
+      await expect(executor.resolveConceptByLabel('test'))
+        .rejects.toThrow(/Alpha.*Beta.*Gamma/);
+    });
+
+    it('should throw "no concept found" when no results returned', async () => {
+      const client = mockClient({
+        searchConcepts: jest.fn().mockResolvedValue(searchResponse([])),
+      });
+      const executor = new GraphOperationExecutor(client);
+
+      await expect(executor.resolveConceptByLabel('Nonexistent'))
+        .rejects.toThrow('No concept found matching label');
+    });
+
+    it('should include similarity percentage in suggestions', async () => {
+      const client = mockClient({
+        searchConcepts: jest.fn().mockResolvedValue(
+          searchResponse([{ concept_id: 'c_x', label: 'Operational Moat', score: 0.689 }])
+        ),
+      });
+      const executor = new GraphOperationExecutor(client);
+
+      await expect(executor.resolveConceptByLabel('Operational Moat'))
+        .rejects.toThrow('68.9%');
+    });
+  });
+
+  describe('executeQueue', () => {
+    it('should default to continue_on_error=true', async () => {
+      const client = mockClient({
+        searchConcepts: jest.fn().mockResolvedValue(searchResponse([])),
+        createConcept: jest.fn()
+          .mockResolvedValueOnce({ concept_id: 'c_1', label: 'A', created: true })
+          .mockRejectedValueOnce(new Error('duplicate'))
+          .mockResolvedValueOnce({ concept_id: 'c_3', label: 'C', created: true }),
+      } as unknown as Partial<KnowledgeGraphClient>);
+      const executor = new GraphOperationExecutor(client);
+
+      const result = await executor.executeQueue([
+        { op: 'create', entity: 'concept', label: 'A', ontology: 'test' },
+        { op: 'create', entity: 'concept', label: 'B', ontology: 'test' },
+        { op: 'create', entity: 'concept', label: 'C', ontology: 'test' },
+      ]);
+
+      // With default continue_on_error=true, all 3 operations should execute
+      expect(result.results).toHaveLength(3);
+      expect(result.successCount).toBe(2);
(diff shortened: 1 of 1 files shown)


aaronsb:
## Code Review -- PR #333

**Scope**: 8 files, 71 additions / 25 deletions. Five MCP UX improvements (#326-#330).

---

### Finding 1: `executeQueue` default mismatch (potential bug)

**Location**: `cli/src/mcp/graph-operations.ts:585` and `cli/src/mcp-server.ts:2581`

The MCP schema and call site now default `continue_on_error` to `true`:

```ts
// mcp-server.ts:2581
const continueOnError = toolArgs.continue_on_error !== false;
```

But the underlying `executeQueue` method signature still declares the old default:

```ts
// graph-operations.ts:585
async executeQueue(
    operations: QueueOperation[],
    continueOnError: boolean = false  // <-- still says false
): Promise<QueueExecutionResult> {
```

This is not a runtime bug today because the MCP handler always passes the resolved boolean explicitly. However, it creates a documentation/intent mismatch: any future caller invoking `executeQueue()` without passing the second argument will get `false` (stop on first error), contradicting the new documented behavior. The function's own default should agree with the public contract, or at least carry a JSDoc note explaining the discrepancy.

**Suggestion**: Update `executeQueue`'s parameter default to `true` or add a comment explaining that the MCP layer owns the default policy.

---

### Finding 2: `resolveConceptByLabel` -- gap between `min_similarity` and acceptance threshold

**Location**: `cli/src/mcp/graph-operations.ts:262-286`

The search fetches results with `min_similarity: 0.6`, but then the acceptance gate is `score >= 0.75`. Results in the 0.60-0.749 range are fetched from the API just to become "did you mean?" suggestions that throw an error.

This is actually a well-designed deliberate pattern: the wider fetch window feeds the suggestion mechanism. Worth noting that it works correctly. The only risk is if the API's `min_similarity` parameter uses a different similarity metric than `score` in the response. If those are always the same metric (and from reading the code, they are), this is sound.

One edge case to consider: if the search returns exactly one result and its score is, say, 0.72, the `suggestions` array will contain only that one item. The error message reads "Did you mean: ..." with a single suggestion, which is fine UX. No issue here, just confirming it was considered.

**Assessment**: Logic is correct. No action needed.

---

### Finding 3: Threshold consistency across code paths

**Location**: `cli/src/mcp/graph-operations.ts` vs `cli/src/lib/validation.ts:418-421`

The MCP `resolveConceptByLabel()` now uses 0.75 acceptance / 0.6 floor. Meanwhile, the CLI's `validation.ts` has its own concept resolution at line 418 using `min_similarity: 0.7` with no acceptance gate (any result from the API is accepted). These are different code paths serving different callers (MCP vs CLI), so they are not required to match. But the divergence means:

- MCP: accepts at >= 75%, suggests at 60-74%, rejects below 60%
- CLI: accepts at >= 70%, no suggestions, rejects below 70%

If this divergence is intentional (MCP agents benefit from stricter matching + suggestions, CLI users are more precise), that is fine. If it drifted accidentally, consider aligning them. Worth a brief note in the code.

**Suggestion**: Add a short comment in `resolveConceptByLabel` noting why MCP uses a different threshold than the CLI validation path, to prevent future confusion.

---

### Finding 4: Tool description quality -- strong overall

The additions are well-crafted. Specific positives:

- The "escalation" framing (search/connect as System 1, program as System 2) gives agents a clear decision heuristic without being overly prescriptive.
- The AGE Cypher gotchas in the program tool description are genuinely useful. The `WHERE type(r)` vs inline `[r:TYPE]` syntax difference is a real pain point.
- The connect no-path suggestions in `formatConnectionPaths()` are actionable and ordered from least to most effort.
- The "Do not repeat connect hoping for different results" line is direct and prevents a real failure mode.

One minor observation: the escalation hint appears in three places (search description, concept description, concept formatter output). The repetition is likely intentional since agents may only see one of these depending on which tool they called. This is reasonable for discoverability.

---

### Finding 5: Documentation says 75% but description also mentions "near-misses (60-75%)"

**Location**: `cli/src/mcp-server.ts:920-921` (graph tool description)

```
- Resolution uses vector similarity (75% threshold) to find matching concepts
- Near-misses (60-75%) return "Did you mean?" suggestions with concept IDs
```

This is accurate and consistent with the code. The two-tier model is clearly communicated. No issue.

---

### Finding 6: No new tests for the changed behavior

The PR body shows 38/38 tests pass, and the manual test plan covers the three behavioral changes. However, `resolveConceptByLabel()` has no unit tests in the repository (confirmed via grep). The three behavioral changes -- threshold lowering, "did you mean?" suggestions, and `continue_on_error` default flip -- are all testable with mocked API responses.

This is not a merge blocker (the function existed without tests before this PR), but if the label resolution logic is important enough to tune thresholds for, it is important enough to pin with tests. Suggested test cases:

1. Score >= 0.75 returns the concept ID directly
2. Score in 0.60-0.74 throws with "Did you mean" containing up to 3 suggestions
3. No results throws with "No concept found"
4. `executeQueue` with default `continueOnError` continues past errors

**Suggestion**: Consider adding these as a follow-up, since the PR is already passing the existing suite.

---

### Finding 7: Minor -- `from_similarity` non-null assertion

**Location**: `cli/src/mcp/formatters/concept.ts:163-164`

```ts
output += `From Match: ${(result.from_similarity! * 100).toFixed(1)}%\n`;
output += `To Match: ${(result.to_similarity! * 100).toFixed(1)}%\n`;
```

These non-null assertions pre-date this PR, but the new suggestion block below them runs when `count === 0`. If the API returns `from_similarity: null` when no paths are found (possible depending on the API contract), these lines would produce `NaN%`. Not introduced by this PR, but worth noting since the formatter was touched.

---

### Summary

| Category | Assessment |
|----------|-----------|
| Correctness | Sound. Label resolution logic, threshold tiers, and queue default all work as described. |
| Consistency | Description text matches code behavior. One cross-path threshold divergence with validation.ts worth documenting. |
| Quality | Tool descriptions are clear, specific, and well-structured. Escalation framing is a good pattern. |
| Risk | `executeQueue` default mismatch (Finding 1) is the main item to address. Threshold lowering from 85% to 75% is reasonable for agent UX. |
| Missing | Unit tests for `resolveConceptByLabel` would be valuable as a follow-up. |

**Verdict**: One item to fix (Finding 1 -- `executeQueue` default alignment), the rest are suggestions. No merge blockers otherwise.

PR #371, opened 2026-05-19
feat(web): incremental graph physics — calm node growth on poke

aaronsb:
## Problem

Any time the graph was "poked" — expand a node, follow a concept, toggle a filter — the CPU and GPU force-sim hooks re-seeded **every** node from a fresh sphere and reset alpha to full (1.0). The entire graph exploded and re-settled on every change.

## Fix

`seedWithCarryover` (new, in `positions.ts`): keeps every node that survived the change at its settled — and drag-adjusted — position, and only sphere-seeds genuinely new nodes so they ease in toward the existing layout. `carriedCount` lets the sim hook pick:

- **nothing carried** → genuinely fresh graph → full `alphaInitial` (1.0)
- **anything carried** → incremental poke → gentle `alphaIncremental` (0.3)

The outgoing layout is snapshotted straight from `positionsRef` at re-seed time (it still holds the previous generation there, index-aligned with a `priorIdsRef`), so there is **zero per-frame cost**. Scene's old unconditional `reheat()`-on-length effect — the thing that slammed alpha to full on every poke — is removed; the hooks now own re-seed + the demand-mode frame kick.

## Scope

Both **Force Graph** and **Document Explorer** benefit — they share the `Scene → useSim` engine, and 2D vs 3D is just the `dimensions` param to the same CPU/GPU hooks. There is no separate D3 physics canvas.

The manual **Reheat** button still does a deliberate full re-layout (`alphaInitial`), unchanged.

## Also in this PR

- Unit test for `seedWithCarryover` (fresh / full carry / partial+reorder / torn-snapshot guard) — 4 tests, passing.
- Reworded three stale `transformForD3` / "2D explorer" comments in `ForceGraph.tsx` (no D3 explorer exists; the rationale is kept, the dead reference dropped).
- Fixed two pre-existing unused-var lint errors in `ForceGraph.tsx` (`onNodeClick`, `mergeRawGraphData`).

## Verification

- `tsc -b`: changed files add **zero** type errors (2 errors remain in `BlockPalette.tsx` / `dataTransformer.ts` — pre-existing on `main`, out of scope).
- `eslint`: all changed + new files clean.
- `vitest`: `positions.test.ts` 4/4 pass.
- **Visual**: confirmed by the author — poking the graph (right-click → add adjacent) now grows calmly instead of exploding.

## Follow-ups (not in this PR)

- GPU hook still fully rebuilds `GPUComputationRenderer` on edge-only changes (calm now, but rebuild cost unchanged). In-place neighbor-texture update is the natural next step if perf matters.
- New nodes spawn from the central sphere region and fly in. If spawning near the connecting node is preferred, that's a separate change.

### DIFF
--- a/web/src/explorers/ForceGraph/ForceGraph.tsx
+++ b/web/src/explorers/ForceGraph/ForceGraph.tsx
@@ -38,7 +38,7 @@ import { useThemeStore } from '../../store/themeStore';
 /** ForceGraph — r3f Canvas + scene composition.  @verified c17bbeb9 */
 export const ForceGraph: React.FC<
   ExplorerProps<ForceGraphData, ForceGraphSettings>
-> = ({ data, settings, onSettingsChange, onNodeClick, onSendToReports, className }) => {
+> = ({ data, settings, onSettingsChange, onSendToReports, className }) => {
   const [selectedId, setSelectedId] = useState<string | null>(null);
   const [hoveredId, setHoveredId] = useState<string | null>(null);
   const [hiddenIds] = useState<Set<string>>(() => new Set());
@@ -72,8 +72,6 @@ export const ForceGraph: React.FC<
   const [focusedNode, setFocusedNode] = useState<string | null>(null);
   // Open node info panels — one per selected node, dismissable.
   const [activeNodeInfos, setActiveNodeInfos] = useState<NodeInfoData[]>([]);
-
-  const mergeRawGraphData = useGraphStore((s) => s.mergeRawGraphData);
   const visibleEdgeCategories = useGraphStore((s) => s.filters.visibleEdgeCategories);
   // Universal filters live in the store so every explorer sees the same
   // filtered data. Reading them via individual selectors keeps zustand's
@@ -179,10 +177,10 @@ export const ForceGraph: React.FC<
 
   const vocabStore = useVocabularyStore();
 
-  // Resolve an edge's category with a 3-level fallback (matches the chain
-  // used by the 2D explorer's transformForD3): vocabulary lookup →
-  // API-provided category → fallback. Keeping the chain in sync means the
-  // shared visibleEdgeCategories store stays consistent across explorers.
+  // Resolve an edge's category with a 3-level fallback: vocabulary lookup
+  // → API-provided category → fallback. Document Explorer resolves edge
+  // categories the same way; keeping the chain identical means the shared
+  // visibleEdgeCategories store stays consistent across both explorers.
   const edgeCategory = useCallback(
     (e: ForceGraphData['edges'][number]): string => {
       let cat = vocabStore.getCategory(e.type);
@@ -255,7 +253,7 @@ export const ForceGraph: React.FC<
     if (edgeColorBy === 'confidence') {
       return es.map((e) => {
         // weight ∈ [0,1] (set from API confidence). Hue 0=red, 120=green
-        // matches the d3 2D explorer's scale so cross-view comparison reads.
+        // — low confidence reads hot, high confidence cool.
         const w = typeof e.weight === 'number' ? e.weight : 0.5;
         const hue = Math.max(0, Math.min(1, w)) * 120;
         return `hsl(${hue}, 70%, 50%)`;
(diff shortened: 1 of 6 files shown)


aaronsb:
## Code Review — incremental graph physics

**Verdict: clean. No blockers, no should-fix. Two nits below.** Reviewed the carry-over snapshot timing in both hooks, the index-alignment invariant across all paths, the Scene effect removal, and test coverage.

### What this accomplishes

Replaces "re-seed every node from a fresh sphere + alpha=1.0 on any node-set change" with `seedWithCarryover`: surviving nodes keep their settled (and drag-adjusted) positions, only genuinely new nodes sphere-seed, and `carriedCount` drives a gentle `alphaIncremental` (0.3) vs full `alphaInitial` (1.0). The unconditional Scene reheat-on-length effect — the actual cause of the whole-graph "sproing" — is gone; the hooks now own re-seed plus the demand-mode `invalidate()` kick.

### Correctness — index-alignment invariant holds

- **CPU** (`useForceSim.ts:136-156`): the `useMemo` reads `positionsRef`/`priorIdsRef` (outgoing generation, index-aligned) and writes both before returning. `useFrame` (`:192`, `:359`-equivalent) only mutates `positionsRef` in place at length `3*N` of the generation that set it up and never touches `priorIdsRef`. Read-then-write is atomic from React's view. Invariant preserved.
- **GPU** (`useGpuForceSim.ts:131-143`): same read-then-reassign ordering, in a post-commit effect that runs before the next frame, so `positionsRef` still holds the previous generation's last readback (`:359-367`). Cleanup (`:293-300`) intentionally does **not** touch `priorIdsRef` — comment (`:101-102`) and behavior agree; cleanup runs before the next effect, so clearing it there would drop the snapshot. Correct.
- **`nodeCount===0` path** (`useGpuForceSim.ts:113-118`): clears `positionsRef` and `priorIdsRef` together — no torn state. CPU has no early return but reaches the equivalent state naturally (`nodes.map` → `[]`, `havePrior` false).
- **StrictMode**: first mount has `priorIdsRef.current === []` → `havePrior` false → `alphaInitial`. Dev double-mount is a fresh instance with fresh refs → same result. No spurious incremental reheat on initial load.
- **Torn-snapshot guard**: `seedWithCarryover` (`positions.ts:60-63`) gates on `priorPositions.length >= priorIds.length * 3` before any indexed read — no OOB. Test covers it.
- `invalidate` confirmed in scope in the CPU hook (`useForceSim.ts:108`).

### Scene effect removal — nothing lost

Old `useEffect(() => sim.reheat(), [nodes.length])` fired only on count change. The replacement `useEffect(invalidate, [nodes])` in both hooks fires on identity change — a strict superset that also covers same-count filter swaps the old code missed. Manual Reheat button still calls `sim.reheat()` with `alphaInitial` — unchanged. The detailed comment left at `Scene.tsx:179-185` is good practice for a non-obvious deletion.

### Nits (non-blocking)

**1. Empty-set carry-over loss — confirm it's intended UX.** Filter → empty → unfilter loses the snapshot (`priorIdsRef` reset to `[]` on the empty generation), so the next non-empty render gets a full `alphaInitial` reheat instead of gentle. This is almost certainly the desired behavior ("nothing on screen → nothing to remember"), but it's an implicit design choice worth a one-line code comment near the GPU `nodeCount===0` reset (`useGpuForceSim.ts:115`) so it reads as deliberate rather than incidental.

**2. Test coverage gap: zero-survivors-with-non-empty-prior.** `positions.test.ts` covers fresh / full carry / partial+reorder / torn snapshot, but not "non-empty prior, zero surviving ids" (full filter swap). That path yields `carriedCount=0` → `alphaInitial`, which is exactly the gentle-vs-full discriminator the sim hooks depend on. One extra case locks the alpha-selection contract against regression.

### Notes (no action)

- `useMemo` writing refs + calling `setAlphaDisplay` during render (`useForceSim.ts:136-156`) is a pre-existing render-purity concern, not introduced here. It's also load-bearing: the carry-over snapshot needs `positionsRef` to still hold the outgoing generation, which moving to an effect would break. Leave as-is.
- `ForceGraph.tsx` comment rewordings and the two unused-var lint fixes (`onNodeClick`, `mergeRawGraphData`) are accurate and in scope.

Nice work — the snapshot timing is the subtle part and it's handled correctly in both the render-phase (CPU) and post-commit (GPU) cases, with the asymmetry documented.

---

PR #374, opened 2026-05-19
feat(web): startup auto-zoom + Force Graph / Document Explorer parity

aaronsb:
## Summary

Startup auto-zoom for both graph explorers, converged visual controls, and a reusable PCA "tangent-orient" camera action — plus two bug fixes found along the way.

## What's in it

**Startup framing (the original ask: "I have to zoom every time")**
- Camera auto-frames on **first load only** (0→N transition; Add Adjacent / filter / follow never re-arm — preserves the PR #371 anti-"sproing" behaviour for the camera).
- Waits ~2s for physics to settle, then eases in (tweened).

**PCA tangent-orient action (`pcaFrame` → `orientView` → `useOrientAndFrame`)**
- Rotates the camera to face the graph's broad face (covariance eigen-decomposition; look along the minor axis). Pure math is unit-tested (30 scene tests).
- **Depth-anchored framing + viewport-aware floor + `pullback` knob**: closeness is the cluster's near face, with a graph-scaled floor so flat/tiny graphs can't collapse the camera into a node. `pullback` is the single tuning knob; per-explorer (Force Graph default, Document Explorer 0.45 — larger glyphs).
- World-Y up (no roll-snap / tilted orbit pole).
- **Double-click a node** → same whole-graph frame, re-aimed so the node is centred/nearest with the bulk swung behind it. Whole-graph (not local-neighbourhood) so successive clicks don't creep closer.
- Lives in the shared `Scene` → Force Graph **and** Document Explorer both get it.

**Explorer convergence**
- Shared `VisualControls` component; Document Explorer gains type-differentiated platonic shapes + lighting parity. Anti-divergence.

**Bug fixes**
- Edges/arrows no longer frustum-culled when the camera wheels in close (stale origin bounding sphere on CPU-animated batched geometry → `frustumCulled={false}`).

## Follow-ups (not in this PR)

- Issue #375 — visual-narration camera spline-path hop (the retained 1-hop adjacency in `useOrientAndFrame` is its substrate).
- Live verification still owed: tween-cancel after 2D↔3D switch; Document Explorer pullback value; double-click consistency.

## Testing

- 30 ForceGraph scene unit tests green (`pcaFrame`, `orientView`, plus existing positions/shapes/facetedMaterial).
- `tsc` + `eslint` clean except 2 **pre-existing** errors (`BlockPalette.tsx`, `dataTransformer.ts`) unrelated to this branch.
- Integration paths (tween/cancel, dblclick, per-explorer wiring) are glue — not unit-tested anywhere in the codebase; verified by live use.

### DIFF
--- a/web/src/explorers/DocumentExplorer/DocumentExplorer.tsx
+++ b/web/src/explorers/DocumentExplorer/DocumentExplorer.tsx
@@ -501,6 +501,7 @@ export const DocumentExplorer: React.FC<
               ? (settings?.projection ?? '3D') === '3D'
               : (settings?.visual?.lighting ?? 'flat') === 'lit'
           }
+          orientPullback={0.45}
           enableDrag
           enableZoom={settings?.interaction?.enableZoom !== false}
           enablePan={settings?.interaction?.enablePan !== false}
--- a/web/src/explorers/DocumentExplorer/DocumentExplorer.tsx
+++ b/web/src/explorers/DocumentExplorer/DocumentExplorer.tsx
@@ -29,6 +29,7 @@ import type {
 import { useThemeStore } from '../../store/themeStore';
 import { StatsPanel, PanelStack } from '../common';
 import { Scene } from '../ForceGraph/scene/Scene';
+import { shapeGeometryByClass, type ShapeName } from '../ForceGraph/scene/shapes';
 import { DIM_MODEL } from '../ForceGraph/scene/dimModel';
 import type { EngineNode, EngineEdge } from '../ForceGraph/types';
 import type { ForceSimHandle } from '../ForceGraph/scene/useForceSim';
@@ -48,10 +49,14 @@ const COLORS: Record<DocNodeType, string> = {
   'extended-concept': '#6366f1',
 };
 
-const NODE_CLASS_BY_TYPE: Record<DocNodeType, 'document' | 'concept'> = {
-  'document':         'document',
-  'query-concept':    'concept',
-  'extended-concept': 'concept',
+// Document Explorer has *named* node types, so shape is a genuine second
+// axis here (unlike Force Graph, where it currently mirrors ontology).
+// A stable explicit map — NOT shapeFor()'s hash — so adding a future
+// type can't silently re-bucket the existing three.
+const TYPE_SHAPE: Record<DocNodeType, ShapeName> = {
+  'document':         'dodecahedron',
+  'query-concept':    'octahedron',
+  'extended-concept': 'tetrahedron',
 };
 
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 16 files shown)


aaronsb:
## Code Review — startup auto-zoom + explorer parity

**What this changes**: Adds a pure PCA "tangent-orient" camera action (`pcaFrame` + `orientView`, 30 passing scene tests), a `useOrientAndFrame` tween hook, first-load auto-zoom (`useFitCamera`), double-click focus in `Nodes.tsx`, and `frustumCulled={false}` on Edges/Arrows. The pure-math layering is clean, the Jacobi eigensolver is well-chosen for degenerate clouds, and the lifecycle commentary (StrictMode geometry-leak asymmetry, demand-loop pumping) is unusually careful. Three substantive issues below before merge.

---

### BLOCKING — `fill` has opposite semantics at its two call sites

**Location**: `web/src/explorers/ForceGraph/scene/useOrientAndFrame.ts:234` (ortho) and `:248-250` (perspective), vs `orientView.ts:163`.

**Problem**: `orientedPerspectiveView` was refactored so `fill` is a *minimum-distance floor*: `dist = Math.max(depthAnchored, exactFit * fill)`. `DEFAULT_FILL = 0.2` is sized for that role. But `frameAlongCurrentDir` (new in this PR — `useOrientAndFrame.ts` does not exist on `main`, so this path is in scope) uses the same `fill` as a raw multiplier:

- Perspective `:248-250`: `dist = max(radius/sin(vFov/2), radius/sin(hFov/2)) * fill`. `radius/sin(vFov/2)` is the exact sphere-fit distance; `× 0.2` puts the camera at 1/5 of that — **inside the cluster**.
- Ortho `:234`: `zoom = minPx / (2 * radius * fill)` → `5×` the exact-fit zoom — far too close.

**Why it matters**: `frameAlongCurrentDir` is the fallback for (a) every 2D-projection first-load and (b) any 3D graph the near-spherical guard catches (`vMax/vMin < 1.5` — easy to hit on a balanced force layout). On those, the auto-zoom plants the camera through the graph. The pure `orientView` tests don't exercise this path, so the suite stays green while the runtime path is wrong.

**Suggested change**: Treat `fill` as a floor here too — fit at the exact distance/zoom and only clamp closer-than-floor — or introduce a distinct option (e.g. `frameFill` with overscan semantics, default ≥ 1.0) and pass it explicitly. Either way the two call sites must not share one value with inverted meaning. Worth a quick manual check on a 2D first-load to confirm the fix.

---

### SHOULD-FIX — first-load orient is not cancelled by interaction during the 2s settle

**Location**: `web/src/explorers/ForceGraph/scene/useFitCamera.ts:67-88`.

**Problem**: `useOrientAndFrame` correctly cancels its tween on the OrbitControls `start` event, but the armed one-shot in `useFitCamera` only watches `nodes.length`. If the user pans/zooms/drags within `SETTLE_MS` (2s), the queued `orient()` still fires at t=2s and yanks the camera off where they put it.

**Why it matters**: 2s is long enough that an eager user interacts first; the camera then jumps. This is the same "sproing" class the header explicitly says it's avoiding for incremental changes — it just isn't guarded for the first-load window.

**Suggested change**: Subscribe to the same controls `start` event (or a canvas `pointerdown`) while armed; on first interaction set `firedRef.current = true` / `armedRef.current = false` to suppress the one-shot. The user keeps the manual Fit escape hatch the header already documents.

---

### NIT — double-click on an already-selected node leaves it deselected

**Location**: `web/src/explorers/ForceGraph/scene/Nodes.tsx:328-342`.

**Problem**: The comment at :336 says "the first click already set it", but if the node was already `selectedId`, the first click of the pair runs `onSelect(null)` (toggle off). Focus still works (it uses the node id, not selection), but the node ends up deselected after a focus double-click — mildly surprising and contradicts the comment.

**Suggested change**: When `onNodeDoubleClick` is wired, defer the single-click toggle into a `DOUBLE_CLICK_MS` timeout that the double-click branch clears; or only toggle-off on the resolved single click. Low severity — behaviour is coherent, just not what the comment claims.

### NIT — non-monotonic clock for click timing

**Location**: `web/src/explorers/ForceGraph/scene/Nodes.tsx:326` (`Date.now()`).

The tween uses `performance.now()`; double-click detection uses `Date.now()`, which can jump backward (NTP/clock adjust) and is lower-resolution. Use `performance.now()` for consistency and a monotonic 300ms window.

---

### Acknowledged as intentional (not flagged)

`adjacency` retained-but-unused (#375 seam), `frustumCulled={false}` on Edges/Arrows (correct trade for CPU-animated single-draw-call batches — the boundingSphere reasoning checks out), the loose `controls` structural cast, and hooks not unit-tested while pure math is. All documented and defensible. The deliberate non-adoption of `pose.up` (keeping world-Y up, `:180-189`) is a good call and well-reasoned.

**Assessment**: Architecture and pure-math layering are solid; tests cover the right (non-obvious) surface. The `fill` divergence is a real runtime defect on a path users hit and should block; the settle-window cancel is a clear UX fix; the two nits are polish.

---

PR #389, opened 2026-05-19
fix(providers): forced tool_use + catalog-driven model resolution

aaronsb:
## Summary

- **Anthropic concept extraction** now uses forced `tool_use` against a strict schema (`EXTRACTION_TOOL_SCHEMA`) — parsed dict comes from `tool_use.input`, no JSON string scraping. `max_tokens` raised from 4096 → 16384 with explicit `stop_reason == \"max_tokens\"` detection so silent truncation can't masquerade as malformed JSON.
- **No hardcoded model literals** anywhere in `ai_providers.py` / `vision_providers.py`. New `_resolve_provider_model_id(provider, category, override, env_var)` does override → catalog → env → `ValueError`. Applied across Anthropic / OpenAI / Ollama / OpenRouter constructors and every inline `translate_to_prose` / `describe_image` site. Catalog (ADR-800/801) is the single source of truth.
- **Vision path** is now catalog-resolved on both `vision_providers.AnthropicVisionProvider` and the legacy `AIProvider.describe_image` (which previously ignored all configuration and pinned `claude-3-5-sonnet-20241022` — retired Oct 28, 2025). Opus 4.7 sampling-param removal handled on every Anthropic call. Truncation now detected on Anthropic + OpenAI vision describe_image too.

## Why this was needed

The chunk-extraction symptom was the user-visible bug, but it was sitting on three layers of latent rot:

1. `messages.create()` with no structured-output enforcement — `max_tokens=4096` truncated dense extractions and the heuristic fence-stripper accepted the partial output downstream.
2. Hardcoded fallback `claude-3-5-sonnet-20241022` (retired Oct 28, 2025) in two vision paths, plus `claude-3-haiku-20240307` (retired Apr 19, 2026) in `translate_to_prose`, plus `claude-sonnet-4-20250514` (Sonnet 4.0 — deprecated June 2026) as the constructor default.
3. Configuration didn't flow into the workers for the legacy inline `describe_image` path — `vision_model_name` was silently ignored.

## Verified

Re-ran the failing ingest after the fix:

```
23:57:17 | httpx     | POST https://api.anthropic.com/v1/messages \"HTTP/1.1 200 OK\"
23:57:17 | ingestion | ✓ Extracted 20 concepts, 24 instances, 20 relationships
```

20-concept payload over a 720-word chunk would have easily exceeded the prior 4096 cap. Downstream label matching fires cleanly off the parsed dict.

## Test plan

- [x] Unit: \`tests/unit/lib/test_vision_provider_catalog.py\` (10/10) — updated for new resolver contract
- [x] Unit: \`tests/test_mock_provider.py\` (10/10) + \`tests/unit/lib/test_provider_config.py\` (19/19)
- [x] Broader \`tests/unit/\` sweep: 465 pass, 8 pre-existing \`test_program_executor.py\` failures confirmed unrelated (reproduced on \`main\` via \`git stash\`)
- [x] Live ingest against \`api.anthropic.com\` — 720-word chunk, 20 concepts / 24 instances / 20 relationships
- [ ] Image ingest path (catalog-resolved vision model) — not exercised in this verification run; spot-check before merging if there's an image doc handy

## Notes / follow-ups (not in this PR)

- \`api/app/lib/ai_providers.py\` is now 2418 lines (was 2203). Was already over the 800-line priority threshold before this change. Worth splitting into \`providers/{anthropic,openai,ollama,openrouter,llamacpp,local_embedding}.py\` as a follow-up.
- The shared \`EXTRACTION_PROMPT_TEMPLATE\` in \`llm_extractor.py\` still ends with \"Return your response as a JSON object\" — now redundant on the Anthropic path (tool schema is the contract) but shared across providers. Per-provider prompt customization is a separate change.

### DIFF
--- a/api/app/lib/ai_providers.py
+++ b/api/app/lib/ai_providers.py
@@ -20,6 +20,83 @@ logger = logging.getLogger(__name__)
 
 
 # Default prompt for image description (ADR-033 Phase 1)
+# Anthropic structured-output contract for concept extraction.
+#
+# Why: messages.create() with no enforcement returns prose that has to be JSON-scraped,
+# and silent max_tokens truncation looks like malformed JSON. Forced tool use returns
+# a parsed dict in tool_use.input and lets us detect truncation explicitly via stop_reason.
+EXTRACTION_TOOL_NAME = "record_extracted_concepts"
+
+EXTRACTION_TOOL_SCHEMA: Dict[str, Any] = {
+    "name": EXTRACTION_TOOL_NAME,
+    "description": "Record concepts, supporting instance quotes, and concept-to-concept relationships extracted from the provided text.",
+    "input_schema": {
+        "type": "object",
+        "additionalProperties": False,
+        "properties": {
+            "concepts": {
+                "type": "array",
+                "items": {
+                    "type": "object",
+                    "additionalProperties": False,
+                    "properties": {
+                        "concept_id": {"type": "string"},
+                        "label": {"type": "string"},
+                        "description": {"type": "string"},
+                        "search_terms": {
+                            "type": "array",
+                            "items": {"type": "string"},
+                        },
+                    },
+                    "required": ["concept_id", "label", "description", "search_terms"],
+                },
+            },
+            "instances": {
+                "type": "array",
+                "items": {
+                    "type": "object",
+                    "additionalProperties": False,
+                    "properties": {
+                        "concept_id": {"type": "string"},
+                        "quote": {"type": "string"},
+                    },
+                    "required": ["concept_id", "quote"],
+                },
+            },
+            "relationships": {
+                "type": "array",
+                "items": {
+                    "type": "object",
+                    "additionalProperties": False,
+                    "properties": {
+                        "from_concept_id": {"type": "string"},
+                        "to_concept_id": {"type": "string"},
+                        "relationship_type": {"type": "string"},
+                        "direction_semantics": {
+                            "type": "string",
+                            "enum": ["outward", "inward", "bidirectional"],
+                        },
+                        "confidence": {
+                            "type": "number",
+                            "minimum": 0.0,
+                            "maximum": 1.0,
+                        },
+                    },
+                    "required": [
+                        "from_concept_id",
+                        "to_concept_id",
+                        "relationship_type",
+                        "direction_semantics",
+                        "confidence",
+                    ],
+                },
+            },
+        },
+        "required": ["concepts", "instances", "relationships"],
+    },
+}
+
+
 IMAGE_DESCRIPTION_PROMPT = """Analyze this image for knowledge extraction. Provide a detailed description:
 
 **Text Content:** Transcribe ALL visible text exactly as written (titles, headings, bullets, labels, annotations).
@@ -96,6 +173,68 @@ def _load_api_key(provider: str, explicit_key: Optional[str] = None, env_var: Op
 
     logger.debug(f"No API key found for {provider}")
     return None
+
+
+def _resolve_provider_model_id(
+    provider: str,
+    category: str,
+    override: Optional[str] = None,
+    env_var: Optional[str] = None,
+) -> str:
+    """Resolve a model id from override → catalog → env → error.
+
+    No hardcoded literals — per ADR-800/801, the provider model catalog is
+    the source of truth. If a deployment has not yet run the provider's
+    'Get models' admin action and no env override is set, this raises with
+    an actionable message rather than silently picking a stale literal that
+    may have been retired.
+
+    For category='vision', selection is by the per-row supports_vision flag
+    (consistent with vision_providers._resolve_vision_model). For other
+    categories ('extraction', 'embedding', …), selection is by the catalog's
+    category column.
+    """
+    if override:
+        return override
+
+    try:
+        from .model_catalog import list_catalog
+        from .age_client import AGEClient
+
+        client = AGEClient()
+        conn = client.pool.getconn()
+        try:
+            rows = list_catalog(conn, provider=provider, enabled_only=True)
+            if category == "vision":
+                rows = [r for r in rows if r.get("supports_vision")]
+            else:
+                rows = [r for r in rows if r.get("category") == category]
+            if rows:
+                for r in rows:
+                    if r.get("is_default"):
+                        return r["model_id"]
+                return rows[0]["model_id"]
+        finally:
+            client.pool.putconn(conn)
+    except Exception:
+        # Degrade to env / error — catalog/DB unavailability must not silently
+        # bind a retired model.
+        logger.debug(
+            f"{provider} catalog lookup failed for category='{category}'",
+            exc_info=True,
+        )
+
+    if env_var:
+        env_value = os.getenv(env_var)
+        if env_value:
+            return env_value
+
+    raise ValueError(
+        f"No {provider} model resolved for category='{category}'. "
+        "Either populate the provider model catalog via the admin 'Get models' "
+        f"action (ADR-800/801)"
+        + (f", or set the {env_var} environment variable." if env_var else ".")
+    )
 
 
 def _list_models_from_catalog(provider: str) -> Optional[Dict[str, List[str]]]:
@@ -254,9 +393,15 @@ class OpenAIProvider(AIProvider):
         )
         logger.info(f"OpenAI client configured with max_retries={max_retries}")
 
-        # Configurable models with defaults
-        self.extraction_model = extraction_model or os.getenv("OPENAI_EXTRACTION_MODEL", "gpt-4o")
-        self.embedding_model = embedding_model or os.getenv("OPENAI_EMBEDDING_MODEL", "text-embedding-3-small")
+        # Catalog-driven model resolution (no hardcoded literals — ADR-800/801).
+        self.extraction_model = _resolve_provider_model_id(
+            "openai", "extraction",
+            override=extraction_model, env_var="OPENAI_EXTRACTION_MODEL",
+        )
+        self.embedding_model = _resolve_provider_model_id(
+            "openai", "embedding",
+            override=embedding_model, env_var="OPENAI_EMBEDDING_MODEL",
+        )
 
         # Optional separate embedding provider (e.g., LocalEmbeddingProvider)
         self.embedding_provider = embedding_provider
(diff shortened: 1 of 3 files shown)


aaronsb (commented):
## Code Review

Substantive review focused on correctness of the forced-tool-use path, the new resolver contract, and the touchpoints the PR changed. Verified end-to-end checks (live Anthropic run, 39 + 465 unit pass) cover happy-path behavior, but several latent correctness and operational issues remain.

---

### Finding 1 — Translation cost regression (silent 12–60x bill increase)
**Severity: HIGH / Blocking-ish — Confidence: high**

`AnthropicProvider.translate_to_prose` (ai_providers.py:988-990), `OpenAIProvider.translate_to_prose` (~line 480), and `OpenRouterProvider.translate_to_prose` (lines 1354-1356) all now resolve via the `extraction` catalog category with an optional `*_TRANSLATION_MODEL` env override.

Previously:
- Anthropic: hardcoded `claude-3-haiku-20240307` (~$0.25/M in)
- OpenAI: hardcoded `gpt-4o-mini`
- OpenRouter: hardcoded `openai/gpt-4o-mini`

After this PR, with no env override, translation runs on the *extraction default* — Sonnet 4.x ($3/M) or Opus ($15/M). Translation fires on every code block / diagram during ingest, so this is potentially a 12–60x cost increase that won't surface in tests.

**Suggestion**: Add a `translation` (or `small`/`cheap`) catalog category and resolve against it. At minimum, document the change loudly in the PR description and add a `WARN` log on first translation call noting which model was selected and its tier. The docstring at ai_providers.py:977-983 mentions ANTHROPIC_TRANSLATION_MODEL as a tuning knob but doesn't flag the cost shift.

---

### Finding 2 — `supports_tool_use` not enforced on extraction resolver
**Severity: HIGH / Should-fix — Confidence: medium-high**

`_resolve_provider_model_id("anthropic", "extraction", …)` (ai_providers.py:178-237) returns the default-or-first row whose `category == "extraction"`. The new `extract_concepts` flow (line 922-923) hard-requires `tool_use` support via `tools=[EXTRACTION_TOOL_SCHEMA]` + forced `tool_choice`.

If an operator marks a non-tool-capable model (e.g. a legacy Claude 2 row, or a hypothetical fine-tune) as the extraction default, the Anthropic API returns a generic 400 and the carefully-crafted `tool_use`-block-missing error path at lines 946-952 never fires. The catalog already has a `supports_tool_use` column (model_catalog.py:51) — filter by it for the extraction category, or document the constraint and add a startup-time assertion.

**Suggestion**:
```python
if category == "extraction":
    rows = [r for r in rows if r.get("category") == category
            and r.get("supports_tool_use", True)]
```
(default-True preserves behavior for catalogs that haven't backfilled the column.)

---

### Finding 3 — Legacy `OpenAIProvider.describe_image` truncation-guard inconsistency
**Severity: MEDIUM / Should-fix — Confidence: high**

The PR description states *"Truncation now detected on Anthropic + OpenAI vision describe_image too."* True for `vision_providers.OpenAIVisionProvider.describe_image` (vision_providers.py:317-323), but **not** for the legacy `AIProvider.describe_image` on OpenAI (ai_providers.py:543-593):

- `max_tokens=2000` (line 576) — vs 8192 in the catalog-driven sibling
- No `finish_reason == "length"` check
- No `temperature` skip for any future OpenAI model that rejects sampling params

Same gap exists on `OpenRouterProvider.describe_image` (lines 1381-1429): `max_tokens=2000`, no truncation guard, no catalog-resolved vision model (it just reuses `self.extraction_model`). That last one may be out of scope, but the legacy OpenAI path is in scope because the PR claims to fix it.

**Suggestion**: Mirror the truncation guard from `vision_providers.OpenAIVisionProvider.describe_image:317-323` into `ai_providers.OpenAIProvider.describe_image`, and bump `max_tokens` to 8192 for parity.

---

### Finding 4 — Bare `OpenAIProvider()` embedding-fallback regression
**Severity: MEDIUM / Should-fix — Confidence: high**

`AnthropicProvider.__init__` (ai_providers.py:886-894) and `OllamaProvider.__init__` (ai_providers.py:1597-1611) construct `OpenAIProvider()` purely to get an embedding implementation. But `OpenAIProvider.__init__` now resolves *both* extraction and embedding models (lines 396-404), and either failing raises `ValueError`.

Pre-PR: `embedding_model=os.getenv(\"OPENAI_EMBEDDING_MODEL\", \"text-embedding-3-small\")` was passed, and `extraction_model` defaulted to the `\"gpt-4o\"` literal — construction succeeded.

Post-PR: an Anthropic-extraction + OpenAI-embedding deployment whose OpenAI catalog has an `embedding` row but no `extraction` row (and no `OPENAI_EXTRACTION_MODEL` env) will fail to construct, and the misleading except clause re-raises *\"Anthropic requires an embedding provider\"* — masking the real cause.

**Suggestion**: Either (a) make `OpenAIProvider`'s extraction resolution lazy (only resolve when `extract_concepts` is called), or (b) accept a sentinel like `embedding_only=True` that skips extraction resolution, or (c) catch the inner `ValueError` more narrowly and re-raise with the original message.

---

### Finding 5 — Connection-pool churn in `_resolve_provider_model_id`
**Severity: MEDIUM / Should-fix — Confidence: high**

Every call to `_resolve_provider_model_id` constructs a fresh `AGEClient()` (ai_providers.py:204), which builds a brand-new `ThreadedConnectionPool(1, 20, …)` (age_client/base.py:69) and runs `LOAD 'age'; SET search_path …` on first checkout. After the function returns, the pool falls out of scope unreferenced — psycopg2's `ThreadedConnectionPool` does **not** auto-close on `__del__`, so pooled sockets either leak until GC or until process exit.

Because `get_provider()` is uncached (queries.py:329, 572, 1509; ingest.py:205; documents.py:161 all call it per-request) and `OpenAIProvider` resolves twice per construction, a single request can build 2–4 pools. `translate_to_prose` / `describe_image` add another pool per call. The PR's `_resolve_vision_model` has the same pattern (vision_providers.py:48-58).

This was a latent issue before — the PR amplifies it by adding more call sites — so it's reasonable to defer the full fix, but it should be tracked. Quick mitigations:
- Use a single short-lived `psycopg2.connect()` per resolve (no pool), or
- Cache an `AGEClient` singleton at module level for catalog lookups, or
- Accept a `conn` parameter so callers that already hold a connection don't churn.

---

### Finding 6 — `EXTRACTION_TOOL_SCHEMA` permissiveness
**Severity: LOW–MEDIUM / Nit — Confidence: medium**

ai_providers.py:30-97:

- `relationship_type` has no `enum` — the model can invent strings. If the downstream graph layer constrains this (e.g. only `IMPLIES | SUPPORTS | CONTRADICTS`), the schema should mirror that constraint so the contract is enforced at the LLM boundary, not at insert time.
- `search_terms.items` has no `minLength` and the array has no `minItems` — a concept can come back with `search_terms: []`, which probably defeats the label-matching step downstream. Worth checking what `llm_extractor`'s label matcher does with an empty list.
- The post-parse code (lines 954-957) only `setdefault`s top-level keys — it doesn't validate that `instances[i].concept_id` and `relationships[i].{from,to}_concept_id` reference an actual `concept_id` in the `concepts` array. JSON-Schema can't cross-reference, but the natural place to add this kind of structural validation is right after `result = dict(tool_block.input)`.

**Suggestion**: Tighten `relationship_type` with an enum if one exists downstream; add `minItems: 1` to `search_terms`; add a small post-parse validator that drops or warns on dangling concept_id references.

---

### Finding 7 — Stale comments / docstrings referencing the old fallback semantics
**Severity: LOW / Nit — Confidence: high**

- `vision_providers.py:30-35` — module-level comment still says *\"hardcoded literal (last resort, warned)\"*. The resolver now raises.
- `vision_providers.py:60-67` — `_catalog_vision_models` docstring says *\"degrades to env/literal\"* and *\"warned literal fallback is the user-visible signal\"*. Both stale; it now degrades to env/error.
- `vision_providers.py:558` — `OllamaVisionProvider.__init__` Args docstring: *\"defaults to VISION_MODEL env or granite-vision-3.3:2b\"*. The literal default is gone.
- `ai_providers.py:1034` — `AnthropicProvider.describe_image` docstring says *\"(catalog → VISION_MODEL env → literal)\"*. No literal anymore.
- `ai_providers.py:22` — comment *\"Default prompt for image description (ADR-033 Phase 1)\"* is now orphaned above the `EXTRACTION_TOOL_NAME` block (the prompt it described is moved down to line 100).

---

### Finding 8 — Exception chaining lost on Anthropic error paths
**Severity: LOW / Nit — Confidence: high**

`raise Exception(f\"Anthropic concept extraction failed: {e}\")` at ai_providers.py:968-969 (and similarly 1026-1027, 1111-1112) drops the original traceback. Use `raise Exception(...) from e` so the carefully-crafted `stop_reason` / missing-tool-use diagnostics survive when the error gets wrapped further upstream by the worker.

```python
except Exception as e:
    raise Exception(f\"Anthropic concept extraction failed: {e}\") from e
```

---

### Finding 9 — Opus-4.7 prefix check fragility
**Severity: LOW / Nit — Confidence: medium**

Four scattered checks `if not model.startswith(\"claude-opus-4-7\"):` (ai_providers.py:928, 1000, 1078; vision_providers.py:460). Correct for current direct-API SKUs (no Bedrock support in the repo, so `us.anthropic.claude-opus-4-7-…` isn't a concern today). But if the sampling-param restriction extends to Opus 5 / a future Sonnet 4.7 / Haiku 4.x, four sites need to change in lockstep.

**Suggestion**: Extract a small helper:
```python
def _anthropic_drops_sampling_params(model_id: str) -> bool:
    \"\"\"True if model rejects temperature/top_p/top_k (Opus 4.7 family).\"\"\"
    return model_id.startswith(\"claude-opus-4-7\")
```
Cheaper to extend, and the intent of the check is self-documenting at each call site.

---

### What's solid

- Forced `tool_use` pattern is clean: explicit `tool_choice`, parsed dict via `tool_block.input`, no fence-stripping. The `tool_use` block discovery loop (lines 941-945) correctly defends against multiple content blocks and uses `getattr` to tolerate SDK shape variance.
- Explicit `stop_reason == \"max_tokens\"` check before the tool-block search is the right ordering — truncation gets a specific error rather than a generic \"no tool_use\" one.
- The resolver contract (override → catalog → env → raise) is consistent across all four call categories. The `try/except: pass-to-env` around catalog access correctly degrades for a down DB rather than failing closed.
- Cross-module coupling is clean: confirmed `ai_providers.py` no longer imports from `vision_providers.py` (the prior `_resolve_vision_model` import is gone — only a docstring reference at line 193 remains, which is fine).
- The dead `_extract_json` removal on `AnthropicProvider` is correct; the helper still exists on OpenAI/OpenRouter/Ollama because those paths still need fence-stripping.
- Test changes correctly mirror the new \"no literal fallback\" contract; the `caplog` warning assertion was correctly replaced with `pytest.raises(ValueError)`.

---

### Verdict

**Needs fixes** before merge. Findings 1 (translation cost) and 2 (`supports_tool_use`) are the substantive blockers; 3 (legacy OpenAI describe_image) and 4 (bare `OpenAIProvider()` regression) are correctness gaps the PR description claimed to close but didn't. Findings 5–9 are non-blocking but cheap to address in the same PR.

PR #404, opened 2026-05-23
fix: annealing/ingestion race — layered defense (closes #402)

### LINKED ISSUE #402: Race condition: annealing dissolves ontologies with in-flight ingestion jobs, causing silent data loss
## Problem

The annealing worker and the ingestion job queue do not coordinate. When ingestion jobs are queued against a target ontology, the annealing manager can dissolve / merge / decompose that ontology before those jobs run. The ingestion worker then dequeues a job whose target no longer exists and the job fails silently — from the operator's perspective, content was submitted and accepted, but never landed in the graph.

**Severity: high.** This is operator-visible data loss: the system reports the ingestion was accepted, but the content is never written and the failure is not surfaced.

### Observed failure timeline

```
t0  Operator submits ingest jobs → queued, target_ontology = X
t1  Annealing cycle runs, judges X low-protection
t2  Annealing dissolves X (or merges into Y)
t3  Ingestion worker dequeues job → target ontology X does not exist
t4  Job fails silently
```

Reproduced during batch-ingest testing.

### Scope note

This issue is **only** about the race condition and its three layered defects. The broader annealing redesign (closed-vocabulary, scorer rework, control-surface) is being captured in a separate ADR — do not pull those design choices in here. The 3 failed proposals currently visible in `kg_api.annealing_proposals` (IDs 35, 36, 37) are evidence of the *other* design issue and are **not** in scope here.

---

## Defect A — Annealing does not check for in-flight work against its target

The annealing manager's candidate selection is graph-state-aware but **queue-state-unaware**. Before proposing demotion / merge / decompose / dissolve for an ontology, it must consult the job queue.

**Acceptance criteria:**
- [ ] Before an annealing cycle proposes any mutation (demote, merge, decompose, dissolve) targeting ontology `X`, the candidate-selection step queries the job queue for jobs whose payload targets `X` in any non-terminal state (queued / pending / processing / awaiting-approval).
- [ ] If one or more such jobs exist, that candidate is vetoed for this cycle (skipped, not deferred-then-executed) and the veto is logged with: ontology id, count of blocking jobs, blocking job ids.
- [ ] The veto is observable: it appears in annealing cycle output / logs and is countable for diagnostics (no silent skip).
- [ ] A test exercises the race directly: queue an ingestion job targeting ontology `X`, trigger an annealing cycle that would otherwise select `X`, assert no proposal is created and the veto is logged.

**Relevant code:**
- `api/app/services/annealing_manager.py` (cycle orchestration, candidate selection)
- `api/app/services/proposal_executor.py` (executes approved proposals)
- `api/app/workers/annealing_worker.py`
- `api/app/lib/annealing_evaluator.py`

---

## Defect B — Ingestion silently fails when the target ontology is missing

Two sub-fixes; both are required.

### B1 — Loud failure on missing target

**Acceptance criteria:**
- [ ] When the ingestion worker dequeues a job whose target ontology does not exist, the job transitions to `status='failed'` (not `dropped`, not `skipped`, not a no-op success).
- [ ] The failure carries a structured, operator-readable error message that names the ontology and the cause, e.g. `\"target ontology X was removed between queue and execution\"`.
- [ ] The failed job is visible in the job-list API response with the error attached (same surface as other failed jobs).
- [ ] A test exercises this path: queue an ingestion job, delete the target ontology, run the worker, assert the job's terminal state is `failed` with the expected error message and that the API surfaces it.

### B2 — Idempotent (re)creation of the target ontology

A queued ingestion is operator intent; a background reorganization should not silently void that intent.

**Acceptance criteria:**
- [ ] When the ingestion worker dequeues a job whose target ontology does not exist, it (re)creates the ontology and proceeds — unless that ontology has been explicitly flagged by an operator as frozen / deleted / do-not-recreate.
- [ ] The flag that prevents recreation is a deliberate, operator-set marker (not the absence of a row). The exact column / mechanism is left to the implementer; the requirement is that there be a clear positive signal an operator placed there.
- [ ] If recreation is blocked by the flag, the job lands in `failed` per B1 with an error that distinguishes \"ontology was deliberately removed\" from \"ontology vanished mid-flight.\"
- [ ] Recreation events are auditable (logged with: job id, ontology id, who/what queued the job).
- [ ] A test covers both branches: (a) missing ontology + no freeze flag → recreated and job runs; (b) missing ontology + freeze flag set → job fails with the freeze-distinct error.

**Relevant code:**
- `api/app/routes/ingest.py`
- `api/app/workers/ingestion_worker.py`
- `api/app/services/job_queue.py`
- `api/app/services/job_scheduler.py`

---

## Defect C — Default annealing cadence is too aggressive

The shipping defaults in `kg_api.annealing_options` emit cycles too frequently relative to typical operator activity, which widens the race window and increases the chance of evaluating ontologies that haven't accumulated enough signal yet. This is independent of the larger control-surface redesign and should ship as a defaults change.

**Acceptance criteria:**
- [ ] The shipping defaults for `kg_api.annealing_options` raise the floor on cycle eligibility along at least two axes: minimum ontology age before it is eligible for evaluation, and minimum activity (e.g. instance / edge / event count, or time-since-last-write) before it is eligible.
- [ ] Specific numeric defaults are chosen, justified in the migration or PR description, and applied via a migration (not only documented).
- [ ] Existing deployments are migrated: the new floors apply to fresh installs and are surfaced as upgrade defaults; an operator who has tuned these values does not have their tuning silently overwritten.
- [ ] After the change, an out-of-the-box install does not evaluate brand-new or near-empty ontologies in its first cycles. A test or documented manual check demonstrates this.

**Relevant code / schema:**
- `kg_api.annealing_options` (existing migrations: `046_annealing_proposals.sql`, `047_annealing_scheduled_job.sql`, `053_annealing_autonomous_default.sql`)
- `api/app/services/annealing_manager.py` (consumes these options)

---

## Out of scope

- Closed-vocabulary / scorer redesign (separate ADR in progress).
- The 3 failed proposals (IDs 35, 36, 37) — those are evidence of the design issue, not this race.
- Any UI work beyond surfacing the failed-job state that already exists in the job-list API.

## Definition of done

All three defects above have their acceptance criteria met, with tests, and an operator running a batch ingest concurrent with annealing observes either successful ingestion or a clearly-failed job — never a silent drop.

aaronsb:
Closes #402.

The annealing manager and ingestion queue did not coordinate. When annealing dissolved an ontology that still had ingest jobs queued against it, the worker dequeued with a missing target and silently recreated the ontology under a fresh id — operator saw "submitted," content never landed where they asked.

Four atomic commits, one per defect in #402, layered prevent → shrink → recover; then three follow-up commits responding to the review on this PR.

## Summary

| Layer | Commit | Defect |
|---|---|---|
| Prevent | `4282768f` | **A** — Annealing vetoes demotion candidates with in-flight ingestion. Queue-aware: queries `kg_api.jobs` for non-terminal ingestion jobs against each demotion candidate; vetoes structurally, logs the count + job_ids, counts vetoes in the cycle result. Promotions are not vetoed (they don't modify the source ontology). |
| Shrink window | `99cadd46` | **C** — Per-ontology cadence floors. Migration 065 adds `min_ontology_age_epochs=3` and `min_ontology_concept_count=5`. Both demotion and promotion candidate selection consult them. Migration uses `INSERT … ON CONFLICT DO NOTHING` so operator-tuned rows stay put; existing keys are deliberately not changed. |
| Recover (loud) | `f73737e5` | **B1** — Worker fails loudly on missing target. Both ingest routes stamp `ontology_existed_at_submit` on the job; a new `_validate_target_ontology` helper raises with a structured, distinct error string when the operator's intent has been invalidated. Job-queue exception handler maps to `status='failed'` with the message on the `error` column; the job-list API already surfaces it. |
| Recover (intent) | `7fbe2498` | **B2** — Tombstone-aware idempotent recreate. Migration 066 adds `kg_api.ontology_tombstones`. Operator-delete writes a tombstone; annealing dissolve does not. Worker consults it: missing + tombstoned → fail; missing + no tombstone → recreate with audit log (operator intent overrides background reorganization). |

## Review response (PR-404, 5 findings)

Code review surfaced 5 findings; advisor reconciliation found that finding #4 as written would have introduced a worse bug (orphan content into a graph being deleted) unless paired with a worker-side change, and that finding #2 had a secondary footgun (vetoed-at-execute proposals permanently failed instead of soft-skipped). Three follow-up commits address all 5 plus both advisor flags.

| Commit | Findings addressed |
|---|---|
| `8f6f0586` (worker) | **#3** restore VANISHED raise; **#4** worker checks tombstone *unconditionally* (the precondition for safely reordering tombstone writes upstream); **#5** tombstone-read fallback reassessed — `existed_at_submit=True` + read-failure now falls through to VANISHED, no longer silently recreates |
| `c0cc5f2b` (routes) | **#1** dissolve writes tombstone (same data-loss class as delete); **#4** tombstone is the *first* graph-mutating step of delete; **#6** advisor-flagged — `POST /ontology/` clears any prior tombstone so operator-recreate doesn't leave a "create succeeds, ingest fails forever" trap |
| `2a99e555` (executor) | **#2** `execute_demotion` re-checks the queue veto before calling `dissolve_ontology` (closes the cycle-to-execute gap that's wide in human-approval mode); advisor-flagged footgun — vetoed proposals return `retry_later=True` and the worker reverts status to `'approved'` (soft skip) instead of marking `'failed'` (permanent dead-end) |

A residual TOCTOU exists between the executor's veto SELECT and the dissolve commit — closing it fully needs an advisory lock on the ontology name in both job-enqueue and dissolve. Documented inline; the worker's `existed_at_submit=True` + missing → VANISHED raise is the downstream backstop for that residual window.

## Definition of done

> An operator running a batch ingest concurrent with annealing observes either successful ingestion or a clearly-failed job — never a silent drop.

- A + executor re-check block the common race upstream at two points (proposal creation and proposal execution).
- C shrinks the residual window for ontologies still settling.
- B1 + B2 + worker-unconditional-tombstone turn the surviving cases into observable, structurally distinct failures (vanished / tombstoned / frozen) or honored recreations.
- Dissolve route + create-clear-tombstone close the operator-initiated paths that B2 left exposed.

## Test plan

- [x] `pytest tests/unit/` + `tests/api/test_ingest.py` + `tests/api/test_ontology_routes.py` — 568 pass (full suite passes after review-response commits; new tests cover one observable per finding)
- [x] Migration 065 applied successfully on dev
- [x] Migration 066 applied successfully on dev
- [ ] Manual smoke: queue an ingest job, force an annealing cycle, observe veto log + no proposal for that ontology
- [ ] Manual smoke: delete an ontology via API, attempt re-ingest, observe `status=failed` with tombstone error
- [ ] Manual smoke: approve a demotion proposal, enqueue an ingest for the same ontology, run the execution worker; verify the proposal status reverts to `'approved'` (not `'failed'`) and dissolve did not run

## Files

- `api/app/services/annealing_manager.py` — veto logic + cadence floors
- `api/app/services/proposal_executor.py` — execute-time veto re-check + `retry_later`
- `api/app/launchers/annealing.py`, `api/app/workers/annealing_worker.py` — option plumbing
- `api/app/workers/ingestion_worker.py` — `_validate_target_ontology` (unconditional tombstone check, VANISHED restored) + `_ontology_tombstone`
- `api/app/workers/proposal_execution_worker.py` — retry_later soft-skip
- `api/app/routes/ingest.py` — stamp `ontology_existed_at_submit`
- `api/app/routes/ontology.py` — tombstone helpers (`_record_ontology_tombstone`, `_clear_ontology_tombstone`); delete/dissolve write tombstone before mutation; create clears tombstone
- `schema/migrations/065_annealing_cadence_floors.sql`
- `schema/migrations/066_ontology_tombstones.sql`
- `tests/unit/services/test_annealing_manager.py`, `tests/unit/services/test_proposal_executor.py`, `tests/unit/workers/test_ingestion_worker.py`, `tests/unit/workers/test_proposal_execution_worker.py`, `tests/api/test_ontology_routes.py`

### DIFF
--- a/api/app/launchers/annealing.py
+++ b/api/app/launchers/annealing.py
@@ -16,7 +16,9 @@ import logging
 
 logger = logging.getLogger(__name__)
 
-# Code defaults — overridden by kg_api.annealing_options rows
+# Code defaults — overridden by kg_api.annealing_options rows.
+# Per-ontology cadence floors (min_ontology_age_epochs, min_ontology_concept_count)
+# are added by migration 065 for #402 Defect C.
 DEFAULTS = {
     "epoch_interval": 5,
     "demotion_threshold": 0.15,
@@ -24,6 +26,8 @@ DEFAULTS = {
     "max_proposals": 5,
     "enabled": True,
     "automation_level": "autonomous",
+    "min_ontology_age_epochs": 3,
+    "min_ontology_concept_count": 5,
 }
 
 
@@ -36,7 +40,13 @@ def _read_options(conn) -> Dict:
             for key, value in cur.fetchall():
                 if key in ("enabled", "derive_edges"):
                     options[key] = value.lower() in ("true", "1", "yes")
-                elif key in ("epoch_interval", "promotion_min_degree", "max_proposals"):
+                elif key in (
+                    "epoch_interval",
+                    "promotion_min_degree",
+                    "max_proposals",
+                    "min_ontology_age_epochs",
+                    "min_ontology_concept_count",
+                ):
                     options[key] = int(value)
                 elif key in ("demotion_threshold", "overlap_threshold", "specializes_threshold"):
                     options[key] = float(value)
@@ -144,6 +154,8 @@ class AnnealingLauncher(JobLauncher):
             "overlap_threshold": options.get("overlap_threshold", 0.1),
             "specializes_threshold": options.get("specializes_threshold", 0.3),
             "automation_level": options.get("automation_level", "autonomous"),
+            "min_ontology_age_epochs": options.get("min_ontology_age_epochs", 3),
+            "min_ontology_concept_count": options.get("min_ontology_concept_count", 5),
             "dry_run": False,
             "triggered_at_epoch": current_epoch,
             "description": f"Annealing cycle at epoch {current_epoch}",
--- a/api/app/routes/ingest.py
+++ b/api/app/routes/ingest.py
@@ -245,6 +245,13 @@ async def ingest_document(
                     content=duplicate_info
                 )
 
+        # #402 Defect B1: Record whether the target ontology existed at submit
+        # time. The worker uses this to distinguish "first ingest, create the
+        # ontology" from "ontology vanished between queue and execution"
+        # (annealing dissolve, manual delete) — silent recreate of a dissolved
+        # ontology would mask operator-visible data loss.
+        existed_at_submit = age_client.get_ontology_node(ontology) is not None
+
         # Prepare job data
         use_filename = filename or file.filename or "uploaded_document"
 
@@ -260,6 +267,7 @@ async def ingest_document(
             "content": base64.b64encode(content).decode('utf-8'),  # Base64 encode for JSON serialization
             "content_hash": content_hash,
             "ontology": ontology,
+            "ontology_existed_at_submit": existed_at_submit,
             "filename": use_filename,
             "user_id": current_user.id,  # Track job owner (user ID from kg_auth.users)
             "username": current_user.username,  # For Garage metadata (FUSE support)
@@ -389,6 +397,9 @@ async def ingest_text(
                     content=duplicate_info
                 )
 
+        # #402 Defect B1: see file-upload endpoint above for rationale.
+        existed_at_submit = age_client.get_ontology_node(ontology) is not None
+
         # Prepare job
         use_filename = filename or "text_input"
 
@@ -396,6 +407,7 @@ async def ingest_text(
             "content": base64.b64encode(content).decode('utf-8'),  # Base64 encode for JSON serialization
             "content_hash": content_hash,
             "ontology": ontology,
+            "ontology_existed_at_submit": existed_at_submit,
             "filename": use_filename,
             "user_id": current_user.id,  # Track job owner (user ID from kg_auth.users)
             "username": current_user.username,  # For Garage metadata (FUSE support)
--- a/api/app/routes/ontology.py
+++ b/api/app/routes/ontology.py
@@ -58,6 +58,91 @@ def get_age_client() -> AGEClient:
     return AGEClient()
 
 
+def _clear_ontology_tombstone(client, ontology_name: str) -> bool:
+    """Remove a tombstone for `ontology_name`, if present.
+
+    Called when an operator explicitly re-creates a previously removed
+    ontology — the recreate is positive operator intent that supersedes
+    the prior removal intent. Without this, the unconditional tombstone
+    check in the ingestion worker would fail every future ingest into
+    the recreated name (#402 PR-404 review, finding #6 / advisor flag).
+
+    Returns True if a tombstone was actually deleted, False if none
+    existed or the lookup failed. Errors are logged but not raised —
+    the create itself should still succeed.
+    """
+    try:
+        conn = client.pool.getconn()
+        try:
+            with conn.cursor() as cur:
+                cur.execute(
+                    "DELETE FROM kg_api.ontology_tombstones WHERE name = %s",
+                    (ontology_name,),
+                )
+                deleted = cur.rowcount > 0
+            conn.commit()
+            if deleted:
+                logger.info(
+                    f"Cleared tombstone for '{ontology_name}' on operator recreate"
+                )
+            return deleted
+        finally:
+            client.pool.putconn(conn)
+    except Exception as e:
+        logger.error(
+            f"Failed to clear tombstone for '{ontology_name}': {e}",
+            exc_info=True,
+        )
+        return False
+
+
+def _record_ontology_tombstone(client, ontology_name: str, removed_by: str, reason: str) -> None:
+    """Record a tombstone for an operator-initiated removal (#402 B2 + PR-404).
+
+    The tombstone is the positive operator-intent signal that distinguishes
+    "deliberately removed by an operator" from "dissolved by background
+    annealing" — the ingestion worker checks for it unconditionally (before
+    reading the graph node) so a tombstone written before the graph delete
+    fails the in-flight ingest with ONTOLOGY_TOMBSTONED_ERROR rather than
+    letting it write orphan content into a graph the operator is removing.
+
+    Annealing's own dissolve_ontology path does NOT call this; that's the
+    asymmetry B2 relies on (annealing dissolve is recoverable via recreate,
+    operator delete/dissolve is not).
+
+    Errors are logged but not raised — the tombstone is defense-in-depth on
+    top of the queue-aware veto (Defect A) and the proposal-executor re-check.
+    """
+    try:
+        conn = client.pool.getconn()
+        try:
+            with conn.cursor() as cur:
+                cur.execute(
+                    """
+                    INSERT INTO kg_api.ontology_tombstones
+                        (name, removed_by, reason)
+                    VALUES (%s, %s, %s)
+                    ON CONFLICT (name) DO UPDATE SET
+                        removed_at = NOW(),
+                        removed_by = EXCLUDED.removed_by,
+                        reason     = EXCLUDED.reason
+                    """,
+                    (ontology_name, removed_by, reason),
+                )
+            conn.commit()
+            logger.info(
+                f"Recorded tombstone for ontology '{ontology_name}' "
+                f"(by={removed_by}, reason={reason})"
+            )
+        finally:
+            client.pool.putconn(conn)
+    except Exception as e:
+        logger.error(
+            f"Failed to record tombstone for '{ontology_name}': {e}",
+            exc_info=True,
+        )
+
+
 @router.post("/", response_model=OntologyNodeResponse, status_code=201)
 async def create_ontology(
     request: OntologyCreateRequest,
@@ -109,6 +194,11 @@ async def create_ontology(
                 status_code=409,
                 detail=f"Ontology '{request.name}' already exists (has source data)"
             )
+
+        # #402 PR-404 review (finding #6 / advisor): operator-initiated
+        # recreate is positive intent that supersedes a prior tombstone.
+        # Clear it so subsequent ingests don't fail TOMBSTONED forever.
+        _clear_ontology_tombstone(client, request.name)
 
         # Get creation epoch
         creation_epoch = client.get_current_epoch()
@@ -1005,6 +1095,13 @@ async def delete_ontology(
 
     Example:
         DELETE /ontology/Test%20Ontology?force=true
+
+    Note (PR-404 review): a tombstone is recorded before any graph
+    mutation. With force=true on a name that never existed, this leaves
+    a tombstone that blocks future ingests into that name until the
+    operator either re-creates the ontology (clears the tombstone) or
+    deletes the row from kg_api.ontology_tombstones directly. The
+    error string operators see explains both recovery paths.
     """
     from ..services.job_queue import get_job_queue
 
@@ -1024,6 +1121,18 @@ async def delete_ontology(
                     status_code=404,
                     detail=f"Ontology '{ontology_name}' not found"
                 )
+
+        # #402 B2 + PR-404 review: write the tombstone *before* any graph
+        # mutation. The worker checks tombstones unconditionally, so any
+        # ingest dequeuing during the delete sees the operator's intent
+        # and fails ONTOLOGY_TOMBSTONED rather than racing into a graph
+        # that's about to be removed.
+        _record_ontology_tombstone(
+            client,
+            ontology_name,
+            removed_by=getattr(current_user, "username", None) or "unknown",
+            reason="operator-initiated delete via API",
+        )
 
         # ADR-057/ADR-081: Clean up ALL Garage objects before deleting sources
         # This includes: images, source documents, and projections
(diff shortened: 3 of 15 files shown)


aaronsb:
## Review: layered defense (#402)

**What this changes**: queue-aware veto (A), per-ontology cadence floors (C), and tombstone-aware loud-fail/recreate (B1+B2). Defense is well-layered and the commit-by-commit decomposition tells the story cleanly. Test coverage on the new logic looks solid.

The review below focuses on findings that change risk under load.

---

### 1. Operator-initiated dissolve bypasses the tombstone (load-bearing gap)

**Location**: `api/app/routes/ontology.py:1740-1794` (`POST /ontology/{name}/dissolve`)

The commit message for B2 frames the distinction as "operator-delete writes a tombstone; annealing dissolve does not." But the operator-facing `dissolve_ontology` route is *also* operator intent and is *also* a missing-target producer, and it does not write a tombstone. Sequence:

1. Operator A queues ingest job against `X`.
2. Operator B calls `POST /ontology/X/dissolve` (intentional, deliberate).
3. Worker dequeues A's job, target missing, no tombstone → **recreate path runs** with audit log line. Content lands under a fresh ontology_id named `X` that Operator B explicitly intended to retire.

This is the same silent-recreate failure mode #402 calls out, just triggered via the dissolve endpoint instead of the dissolve algorithm. Recommend writing a tombstone in this route too (with `reason="operator-initiated dissolve via API"` for distinguishability). The A veto does *not* cover this path — A only consults `kg_api.jobs` from within the annealing manager.

---

### 2. Tombstone write and graph delete are not atomic

**Location**: `api/app/routes/ontology.py:1147-1184`

`delete_ontology_node` runs first (its own implicit transaction); the tombstone INSERT runs afterward on a separately checked-out connection. Window:

1. T0: `delete_ontology_node` commits — ontology gone from graph.
2. T1: worker on another process dequeues, calls `_validate_target_ontology`, queries `ontology_tombstones`, finds none.
3. T2: worker takes the recreate path.
4. T3: route writes tombstone.

This is narrow but real. Two structural fixes worth considering: (a) wrap the graph delete and tombstone insert in a single connection/transaction so both succeed or both fail; (b) reorder — write the tombstone *first* (it costs nothing if the delete fails, and a stranded tombstone is recoverable by removing the row, whereas a stranded "ontology gone, no tombstone" window is the exact race that needs closing).

Option (b) is the simpler fix.

### 3. Proposal executor does not re-check the queue veto

**Location**: `api/app/services/proposal_executor.py:166-201`

The A veto blocks *proposal creation*, but `execute_demotion` only re-validates "ontology still exists" and "not pinned/frozen." Approved proposals that pre-date a newly-enqueued ingest job will still dissolve. With operator-approved (non-autonomous) cycles this can sit pending for arbitrary time.

Recommend `execute_demotion` re-run the same `_get_inflight_ingestion_targets({name})` check before calling `dissolve_ontology` and return `{success: False, error: "in-flight ingestion blocking demotion"}` if any rows surface. In autonomous mode the window is short; in human-approval mode it's wide-open.

### 4. `ONTOLOGY_VANISHED_MID_FLIGHT_ERROR` is dead code after B2

**Location**: `api/app/workers/ingestion_worker.py:30-37, 95-147`

After B2, `_validate_target_ontology` raises only `ONTOLOGY_TOMBSTONED_ERROR` or `ONTOLOGY_FROZEN_ERROR`. The "vanished" string is defined and the comment claims it surfaces on "legacy/pre-B2 jobs where `ontology_existed_at_submit` was True but the tombstone path was not checked" — but I don't see a code path that emits it. The `existed_at_submit` parameter survived only as an audit-log field.

Two acceptable resolutions: delete the constant + the parameter and document that B2 supersedes B1's loud-fail semantics; or restore B1's check for `existed_at_submit=True + missing + no tombstone` and raise `ONTOLOGY_VANISHED_MID_FLIGHT_ERROR` instead of recreating in that specific subset (preserving B1's loud-fail for the population where the queue-veto failed but the operator never deliberately deleted). The latter is more conservative — the recreate path is *only* triggered when both queue-veto (A) and tombstone-check (B2) failed to fire, which is the worst signal-to-noise regime to silently recover from.

### 5. Tombstone-read failure tolerance (judgment call, not a defect)

**Location**: `api/app/workers/ingestion_worker.py:81-89`

The "log loudly, fall back to recreate" stance is defensible *given* the queue-veto is the upstream layer. If finding #3 lands and the veto becomes truly upstream-effective, this is fine as-is. If #3 does not land, this fallback is the third silent-recreate route in the system, after #1 and the executor-bypass in #3.

### 6. Tested defaults match migration defaults

- A: queue-veto SELECT uses `ANY(%s)` over status + ontology lists, single statement, atomic snapshot of the jobs table. Good. Note `status = 'pending'` is on the non-terminal list — matches the initial INSERT in `job_queue.enqueue`.
- C: launcher defaults (3 / 5) match migration 065 seed values and worker `job_data.get(..., 3/5)` fallbacks — three places stay in sync.
- B1: route default `True` for `ontology_existed_at_submit` matches worker default `True` — pre-B1 jobs default to the safer behavior, as advertised.

### 7. Migration idempotency

- 065: `INSERT … ON CONFLICT (key) DO NOTHING` + `INSERT INTO schema_migrations … ON CONFLICT (version) DO NOTHING`. Re-runnable. Reversible by hand (`DELETE FROM annealing_options WHERE key IN (...)`).
- 066: `CREATE TABLE IF NOT EXISTS` + same migrations guard. Re-runnable. Drop is straightforward.

### 8. Test file size

`tests/unit/services/test_annealing_manager.py` is 784 lines but exclusively test functions across nine logical groupings. Test files trade breadth for cohesion — flagging only because you asked. No action recommended.

---

### Assessment

Defense-in-depth structure is correct, the commit decomposition is exemplary, and the per-defect tests carry their weight. Three findings would change behavior under load:

- **#1 (operator dissolve bypasses tombstone)** — same data-loss class as #402, different trigger. Should ship with this PR.
- **#3 (executor doesn't re-check veto)** — leaves a wide-open window in human-approval mode. Should ship with this PR.
- **#4 (dead error string)** — either resolution is fine; I'd prefer restoring the B1 raise so the recreate path narrows to "veto-missed AND no operator delete intent AND existed at submit."

Findings #2 (tombstone race window) and #5 (read-failure fallback) are smaller and depend on whether #3 lands as proposed.

---

aaronsb:
## Second-pass review — fresh eyes on the response

The three response commits address the original findings cleanly and the rationales in the commit messages are tight. The advisor-flagged fixes (worker-unconditional check, clear-on-create, retry_later soft-skip) are exactly the right shape. Two new findings surface from this pass, one of them parallel to #6.

---

### Finding A (request changes): rename does not clear tombstones — same shape as #6

`api/app/routes/ontology.py:1287-1362` (`rename_ontology`) does not interact with `kg_api.ontology_tombstones`. With the unconditional tombstone check now in the worker and clear-on-create only in `POST /ontology/`, a routine operator recovery flow dead-ends:

1. `DELETE /ontology/X` writes tombstone(X)
2. Operator decides to repurpose Y as X: `POST /ontology/Y/rename {new_name: X}`
3. Rename succeeds at the graph level; tombstone(X) still present
4. Any future ingest into X fails `TOMBSTONED` permanently

This is the same class of trap that finding #6 closed for `POST /ontology/`. The underlying invariant the response is reaching for — *any operator action that successfully establishes an ontology under name N supersedes tombstone(N)* — is enforced at only one of two entry points.

**Suggested fix**: after `client.rename_ontology_node(...)` succeeds (the Source-rename path already committed by then), call `_clear_ontology_tombstone(client, request.new_name)`. Mirror the create-route comment so future readers see the invariant.

---

### Finding B (request changes): dissolve leaves a stale tombstone on partial / TOCTOU failure

`api/app/routes/ontology.py:1846-1868`: the route now pre-flights existence + lifecycle, writes the tombstone, then calls `client.dissolve_ontology`. `dissolve_ontology` in `api/app/lib/age_client/ontology_scoring.py:527` has additional failure modes the route pre-flight cannot cover:

- source-listing query fails (DB error at line 581)
- `reassign_sources` returns success=False (line 598)
- TOCTOU on lifecycle between route pre-flight and `dissolve_ontology`'s own pre-flight

In any of these, the tombstone has committed and the ontology node still exists. The delete route's docstring acknowledges the analogous "force=true on a nonexistent name" case (lines 1099-1104). The dissolve route's docstring (lines 1817-1825) does not — and dissolve failure is closer to "didn't actually happen" than "operator chose to remove non-existent thing," so leaning on the same recovery path is harder to justify.

**Two options:**
1. Catch dissolve failure and call `_clear_ontology_tombstone` on the rollback path. Cleanest semantically (failed dissolve ≠ operator removal intent). Subtlety: if `reassign_sources` was partially successful before failing, the ontology is in a half-state and clearing the tombstone is arguably also wrong — but it's no worse than the current "block forever" outcome.
2. Add an explicit docstring note matching delete's pattern and document the manual recovery (delete the row from `kg_api.ontology_tombstones`).

Option 1 is preferable; option 2 is acceptable if option 1 is judged to add too much branching.

---

### Acknowledged: what the response got right

- The reasoning in `8f6f0586`'s commit message for moving the tombstone check before `get_ontology_node` is exactly the load-bearing argument. The "tombstone present + node still present" test (`test_tombstone_present_with_live_node_still_raises`) pins the behavior the asymmetric route/worker ordering depends on. Good test surface.
- Restoring the `VANISHED` raise + the new split on tombstone-read failure (`existed_at_submit=True → VANISHED`, `False → create`) is the correct read of the original review's intent — the response improved on what I originally asked for.
- `retry_later` + worker reverting the claim to `'approved'` instead of `'failed'` is the right shape. The scheduler in `api/app/services/job_scheduler.py:272-310` re-dispatches stranded approved proposals with a 5-minute `reviewed_at` floor on a default 1-hour cleanup cadence; combined with `_claim_proposal`'s `WHERE status='approved'` atomic guard, this gives bounded retry without thrash and without double-execution. Mention this in the PR body's "retry semantics" if helpful.
- The recreate-then-ingest sequence is clean: `POST /ontology/X` clears tombstone(X) before `create_ontology_node`, and `DELETE` already calls `queue.delete_jobs_by_ontology` so there's no pre-delete-queued ingest to dequeue against the recreated namespace.
- The TOCTOU residual between the executor's veto SELECT and the dissolve commit is honestly documented at lines 225-232 of `proposal_executor.py`, and the worker's `VANISHED` raise is the right downstream backstop. The advisory-lock path you've flagged for future work is the correct direction.

---

### Test coverage note (informational, not a finding)

`test_demotion_vetoed_by_inflight_ingestion` overrides `mock_cursor.fetchall` on the shared cursor fixture to return job IDs. The assertion shape is sound (retry_later True, dissolve not called, job IDs in error), but the mock accepts any SQL — a regression that filtered by `status='completed'` or queried the wrong table would still pass. The behavior under test is right; the SQL surface is unverified. Documented inline in this comment for the record; not requesting a change since the SELECT is a short mirror of `AnnealingManager._get_inflight_ingestion_targets` whose own tests cover the SQL shape.

`TestDeleteOntologyTombstone`'s side-effect ordering assertion (asserts `delete_ontology_node.call_count == 0` inside the tombstone-record side effect) is a robust pattern — couldn't pass under reverse ordering at the Python call level.

---

### Summary

Two real changes requested (rename + dissolve stale-tombstone), parallel shape to fixes the response already accepted. Everything else lands cleanly. The reasoning in the commit messages and inline comments is consistent and load-bearing — the asymmetric route/worker design is well-defended.

---
*AI-assisted second-pass review via Claude Code*

PR #408, opened 2026-05-24
fix(#405): honest CPU fallback for local embeddings + ADR-101 ROCm strategy

### LINKED ISSUE #405: AMD ROCm init path silently broken: API image has CUDA-only PyTorch
**Symptom**

`./operator.sh init` "Hardware: AMD ROCm" path produces a usable-looking config (profile activates with `device=cuda`), but the API container fails at startup with:

```
Failed to load embedding model nomic-ai/nomic-embed-text-v1.5:
Found no NVIDIA driver on your system.
```

Ingestion then fails with `Embedding model not loaded. Call load_model() first.` for every concept. The startup log claims "Falling back to API-based embeddings" — that fallback never actually engages; the local provider chain still tries the unloaded manager and raises.

**Root cause**

The `kg-api` Docker image ships **CUDA-only PyTorch**:

```
$ docker exec kg-api-dev python -c "import torch; print(torch.__version__, torch.version.hip)"
2.12.0+cu130 None
```

There is no ROCm/HIP support compiled in. For AMD users with ROCm, the wheel would need to be `torch==X+rocmY` instead of `+cu130`. The current image cannot use ROCm devices regardless of host configuration.

**Why it looks fixed but isn't**

The branch `fix/catalog-refresh-embedding-coupling` (commit `a5a5efbc`) does map the wizard's GPU choice to a PyTorch device string, but it maps both `nvidia` and `amd*` to `cuda`. ROCm uses the `cuda` device name only when PyTorch is built with HIP — so on the current CUDA-built image that mapping still fails for AMD users.

**Reproduction**

1. Fresh `./operator.sh init`
2. Choose Hardware → AMD ROCm
3. Complete init, observe the profile shows `(cuda)`
4. Submit any ingest job — fails with the trace above

**Fixes to consider**

- Build a separate `kg-api-rocm` image with ROCm PyTorch wheels; pick image by host hardware at install time
- Or: detect at startup that requested device is unusable and **actually** fall back to CPU rather than emit a misleading log line
- Or: in the operator config flow, map AMD ROCm → `cpu` until a ROCm image is published, with a clear note explaining why

**Workaround**

```
docker exec kg-operator python /workspace/operator/configure.py embedding --device cpu 1
docker restart kg-api-dev
```

CPU embedding works fine for nomic-embed-text-v1.5 (~137M params).

**Note on the misleading fallback log**

The "Falling back to API-based embeddings" line at startup is wrong when the active profile is `local`: it implies recovery happened, but the local-embedding code path then crashes on every request. Separate issue — possibly worth its own ticket — but worth flagging here since it makes the AMD ROCm failure mode harder to diagnose.

aaronsb:
Closes #405. End-to-end verified on AMD 7900 XTX (gfx1100) + iGPU 9950X3D (gfx1036), Arch ROCm 7.2.3 host.

## Summary

Issue #405 reported that `./operator.sh init` "AMD ROCm" path produced a usable-looking config but every ingest then failed with `Embedding model not loaded. Call load_model() first.` The startup log claimed "Falling back to API-based embeddings" — a lie; no fallback actually occurred. The root cause is image packaging: published `kg-api:latest` ships PyPI's default torch wheel (CUDA runtime bundled, no ROCm). This PR ships both halves of the fix.

## What's in this PR

**ADR-101 (Accepted)** — `docs/architecture/infrastructure/ADR-101-rocm-image-variant-and-install-time-selection.md`. Frames the root cause, proposes three ROCm variant tags (`rocm60`, `rocm61`, `rocm72-host`) with install-time selection via `KG_API_IMAGE_TAG` substitution in `docker-compose.ghcr.yml`. NVIDIA stays on `:latest` (default PyPI torch wheel carries the CUDA runtime — works on NVIDIA and CPU hosts unchanged).

**Honest CPU fallback** — `api/app/lib/embedding_model_manager.py`. When `load_model()` fails on the configured device (CUDA-built torch on AMD host, missing /dev/kfd for ROCm, MPS on non-Apple, etc.), retries once with `device='cpu'` rather than leaving the platform silent-broken. Also closes a global-pollution bug where a failed init left `_model_manager` pointing at a half-built manager (the actual reason concepts saw "model not loaded" instead of "not initialized"). Hot-reload deliberately does NOT carry the fallback — operator just made an explicit just-typed device choice, atomic-swap preserves the previous working model on failure, silently downgrading would be the worse failure.

**Honest startup logs** — `api/app/main.py`. No more "Falling back to API-based embeddings" when the active profile is still `local`. Mirror fix at line 218 for visual-embedding init. On both-paths-failure, the message surfaces what actually happened with concrete remediation paths.

**Published ROCm image** — `ghcr.io/aaronsb/knowledge-graph-system/kg-api:rocm72-host`. Built from `api/Dockerfile.rocm-host` on AMD's official `rocm/pytorch:rocm7.2.3_ubuntu24.04_py3.12_pytorch_release_2.9.1` base image. Single-arch linux/amd64 (ROCm has no production arm64 path). `publish.sh images-rocm` is the new target group; rocm60/rocm61 variants stay gated behind `--force` until someone runs them on real hardware (#409).

**Operator wiring with single-source-of-truth helper** — `operator/lib/image-tag.sh` is the one place `GPU_MODE → KG_API_IMAGE_TAG` is defined. `operator.sh:load_config`, `operator/lib/common.sh:load_operator_config`, and `operator/lib/guided-init.sh` all source it. The earlier inline-three-times shape drifted between commits, the helper ends that. `docker/docker-compose.ghcr.yml` parameterized to substitute the tag. `operator/lib/guided-init.sh` persists the tag in `.operator.conf`, warns about `ROCR_VISIBLE_DEVICES=0` on hosts with both a discrete dGPU and an iGPU, and reframes wizard option [3] as "preview" (it now resolves to `:latest` + CPU fallback because rocm60/rocm61 are deferred — previously it pointed at an unpublished image). `operator.sh:get_compose_cmd` gained the missing `amd-host` GPU overlay case — without it, no `/dev/kfd` device mounts, no `privileged: true`, ROCm runtime had nothing to talk to inside the container.

## Verified on real hardware (7900 XTX)

| Check | Result |
|---|---|
| Container image pulled | `ghcr.io/.../kg-api:rocm72-host` |
| `/dev/kfd` + `/dev/dri` in container | Mounted |
| `rocminfo` inside container | gfx1100 (7900 XTX) + gfx1036 (iGPU) |
| `torch.cuda.is_available()` | True |
| `torch.version.hip` | 7.2.53211-c2d9476115 |
| Text embedding device | `nomic-ai/nomic-embed-text-v1.5 (768 dims, cuda)` |
| Vision embedding device | `nomic-ai/nomic-embed-vision-v1.5 on cuda` |
| GPU under load | 7900 XTX 12% busy, iGPU 1% idle — work routes to discrete |
| VRAM detected | 24268 MB free |

Diagnosis also incidentally verified the CPU fallback fires correctly on the ROCm container when the GPU overlay wasn't yet applied — proof the two halves of the ADR work together as designed.

## Commits

1. `fac3daa3` — ADR-101 draft → Accepted
2. `49fc13ba` — Real CPU fallback in `init_embedding_model_manager`; global-pollution fix; honest main.py log
3. `fd2e72cf` — Review follow-ups: extract `_build_with_cpu_fallback` helper, case-insensitive `'cpu'` retry guard, real-torch integration test, attempt-then-success log ordering, ADR status promoted to Accepted, async-reload asymmetry comment
4. `9c7db593` — `publish.sh images-rocm` target + `kg-api:rocm72-host` published; operator/lib/common.sh KG_API_IMAGE_TAG export; ghcr compose parameterization; guided-init persistence + multi-GPU note; Dockerfile.rocm-host base bumped to ROCm 7.2.3
5. `c448f098` — Drift fix between operator.sh (top-level) and operator/lib/common.sh: KG_API_IMAGE_TAG export, missing `amd-host` overlay case
6. `2583051c` — Final review pass: collapse GPU_MODE→TAG triplication into `operator/lib/image-tag.sh`, fix wizard option [3] footgun (was pointing at unpublished `rocm60` image), honest visual-embedding log at main.py:218, doc-surface drift in ADR-101 + publish.sh comment

## Test plan

- [x] `pytest tests/unit/ tests/api/` — 921 passed, 15 skipped (no regressions)
- [x] `tests/unit/lib/test_embedding_model_manager_fallback.py` — 8 cases (happy path, CUDA-fails / CPU-rescues, both-fail terminal state with cleared global, CPU-as-configured no-retry, case-insensitive `'cpu'` parametrized)
- [x] `tests/unit/lib/test_embedding_model_manager_real_torch.py` — exercises real sentence-transformers loader (no `load_model` mock) on no-CUDA host, proves the catchable-exception contract
- [x] End-to-end ROCm verification on 7900 XTX — model loads on ROCm device, 24GB VRAM detected, embeddings generate on discrete GPU
- [x] ADR-101 lints clean (`docs/scripts/adr lint`)
- [x] Shared `derive_kg_api_image_tag` smoke-tested across all three call sites

## Tracking issues for deferred follow-ups

- #409 — Build & publish ROCm 6.x variants (rocm60, rocm61)
- #410 — CPU fallback for visual embedding generator (parity with text)
- #411 — Operator: `--image-source` flag silently dropped by guided-init.sh
- #412 — Operator: persisted `KG_API_IMAGE_TAG` goes stale on manual `GPU_MODE` edits

## Housekeeping

Stale branch `fix/catalog-refresh-embedding-coupling` was deleted local + remote during this work (commits already on main via direct push, never PR'd).


aaronsb:
## Review — post-`c448f098` final state

Focused on `fd2e72cf`, `9c7db593`, `c448f098`. End-to-end verification on the 7900 XTX confirms the `amd-host` half lands clean; the `amd` (wheel-based) half ships a footgun that should be addressed before merge.

---

### Must-fix / discuss

**1. Wizard offers an image that doesn't exist in GHCR.**

`gh api .../kg-api/versions` shows only `latest`, `rocm72-host`, and versioned aliases published. `kg-api:rocm60` and `kg-api:rocm61` are **not in the registry** — `cmd_images_rocm` gates them behind `--force` and they've never been built.

But the wizard at `operator/lib/guided-init.sh:187-193` offers **[3] "Linux with AMD GPU (ROCm wheels)"** as a peer option, with description "For systems with ROCm 6.x installed." A user on Ubuntu 22.04 + ROCm 6.x picks the option that names their hardware → `GPU_MODE=amd` → both `operator.sh:75` and `operator/lib/common.sh:47` derive `KG_API_IMAGE_TAG="${ROCM_VERSION:-rocm60}"` → `docker compose pull` fails with `manifest unknown` for `kg-api:rocm60`.

This is the exact "looks fixed, silently broken at install" failure mode #405 was filed to eliminate — except now it bites the wheel-based AMD users instead of the host-mode ones. The PR claims to close #405 for AMD users; today it only closes it for `amd-host` users.

Pick one (don't ship as-is):
- Gate option 3 in the wizard with `[EXPERIMENTAL — requires \`--image-source local\` to build]` until rocm60 ships
- Move option 3 behind a "Show advanced/unverified options" prompt
- Have `operator.sh` print a clear pre-pull warning when `KG_API_IMAGE_TAG` resolves to a variant that won't be in GHCR
- Build + push `rocm60`/`rocm61` unverified (contradicts your own ` cmd_images_rocm` "refuse to ship untested" stance — probably wrong, listed for completeness)

**2. Doc/code drift on the publish-flow surface.**

- `ADR-101.md:138-141` documents the command as `./publish.sh images api-rocm --variants rocm60`.
- `scripts/publish.sh:767` header comment also says `--variants rocm60,rocm61`.
- The actual implemented surface is `./publish.sh images-rocm rocm60 rocm61` — a separate subcommand with positional args, no `--variants` flag.

Two stale references to a non-existent flag, both in load-bearing maintainer-facing docs. Cheap to fix; should align with what `cmd_images_rocm` actually parses.

**3. Triplication of the `GPU_MODE` → `KG_API_IMAGE_TAG` derivation.**

`operator.sh:73-80`, `operator/lib/common.sh:39-54`, and `operator/lib/guided-init.sh:316-318` all carry the same case statement. The fact that `c448f098` was necessary at all — because `operator.sh`'s copy fell out of sync with `common.sh`'s — is empirical evidence the duplication is unsafe.

You asked whether the two-copy version is intentional separation by entry-point. The answer would be yes for *two*. For *three*, the next ROCm version bump means three edits and a high probability of repeating exactly this drift. Suggest a tiny shared file — `operator/lib/image-tag.sh` exporting a `derive_kg_api_image_tag()` function — sourced by all three. The triplication is now the load-bearing problem, not the original two-copy split.

---

### Should-fix

**4. The asymmetry comment defends a contract no test locks.**

The fix to extract `_build_with_cpu_fallback` + the long comment in `reload_embedding_model_manager` are sound, but `tests/unit/lib/test_embedding_model_manager_fallback.py` only exercises the init path. The reload path's claims — (a) does NOT fall back, (b) atomic-swap preserves the previous working manager on failure — are asserted by code reading only.

Add a test: existing `_model_manager` set → call `reload_embedding_model_manager` with a `load_model` that raises → assert `_model_manager` is still the original instance and a `RuntimeError` propagates. Otherwise a future refactor of `reload_*` breaks the contract silently and the comment becomes a lie.

**5. Visual embedding init log still uses the dishonest pattern this PR explicitly killed for text.**

`api/app/main.py:218`:
```python
logger.info("   Visual embedding features may be limited")
```

Same shape as the `"Falling back to API-based embeddings"` line you just rewrote at line 199. When `init_visual_embedding_generator` raises and an image-using profile is active, "may be limited" is just as misleading as the old text-path message. The full fallback is correctly deferred per ADR-101, but the log line is a one-line edit (downgrade severity, surface what actually happens — ingestion of image content will fail until the profile is changed or the device is fixed) and it's in scope for this PR. Skipping it leaves a half-honest startup log.

**6. No tracking issues for the deferred work.**

PR description lists three follow-ups (rocm60/rocm61 publish, visual fallback, `--image-source` parity); GitHub issue search returns only #405. "Deferred to follow-up PR" without an issue is the kind of promise that decays. Open three before merge.

**7. `.operator.conf` `KG_API_IMAGE_TAG` becomes stale if a user edits `GPU_MODE` by hand.**

`guided-init.sh` persists the derived tag at init time. Both loaders short-circuit on `if [ -z "$KG_API_IMAGE_TAG" ]`. So a user who later edits `GPU_MODE=cpu` → `GPU_MODE=amd-host` in `.operator.conf` (which the auto-generated header comment at `guided-init.sh:313` actively suggests they do) will keep the old `KG_API_IMAGE_TAG=latest` and silently land on the CUDA-bundled image — the same shape of bug `c448f098` just fixed.

Two viable shapes:
- Persist only the user-overridable bits (`GPU_MODE`, `DEV_MODE`, `IMAGE_SOURCE`) and always re-derive `KG_API_IMAGE_TAG` on load. Honors edits.
- Keep persisting, but document inline in the generated file: `# DO NOT edit GPU_MODE without also updating KG_API_IMAGE_TAG`. Worse — relies on the user noticing.

Related: the `.operator.conf` auto-generated comment (`./operator.sh config --dev true --gpu nvidia|amd|mac|cpu`) references a `config` subcommand that doesn't exist in `operator.sh`. Another doc/code drift, low severity.

---

### Nits

**8. `ROCM_VERSION` override is undocumented in the persisted config.**

`common.sh:47` honors `ROCM_VERSION` as a shell-level override (`amd → ${ROCM_VERSION:-rocm60}`). Nobody writes it; the header comment in `.operator.conf` doesn't mention it; the wizard doesn't ask. Either drop the override path (currently dead weight outside test/dev) or document it as a comment line in the generated file.

**9. `guided-init.sh:317` doesn't honor `ROCM_VERSION` (only the other two copies do).**

Consistent across the three copies for `amd` mode in `common.sh` and `operator.sh`, but `guided-init.sh` ignores it. Either way, fixing this is moot if you collapse the triplication (item 3).

---

### Praise

- The `_build_with_cpu_fallback` extraction is cleanly keyword-only-args and module-scope; the asymmetry comment on `reload_*` is exactly the kind of long-form-rationale-in-code that earns its keep three months from now when someone asks "why don't we fall back here too?"
- The case-insensitive `cpu` retry guard with the parametrized test is a small fix that prevents real double-log noise.
- `test_embedding_model_manager_real_torch.py` is the right shape — proves the catchable-exception contract the mocked tests can only assume. The skip-if-CUDA guard makes the assertions honest.
- The `rocm71-host → rocm72-host` rename is sound (the old tag was never published; `gh api` confirms) and the "naming convention" addition to ADR-101 — future ROCm 7.3 lands as a new tag rather than overloading — is the right immutability stance.
- End-to-end verification on real hardware (7900 XTX gfx1100 + iGPU 9950X3D gfx1036), with the CPU fallback also firing correctly *during* diagnosis, is the kind of evidence that turns "should work" into "verified."
- The `ROCR_VISIBLE_DEVICES=0` hint at yellow-info-warning severity is the right level — discrete + iGPU coexistence on Ryzen 7000+ is a real foot-gun and runtime ordering isn't guaranteed across kernel/ROCm versions, even if your hardware happened to enumerate the discrete card first.

---

### Empirical checks done in this review

- `gh api .../kg-api/versions` → confirmed `rocm60`/`rocm61` not in registry
- `requires-python` floor → only `fuse/pyproject.toml` pins (`>=3.11`); API has no constraint. py3.13→py3.12 base bump safe.
- `grep -rn "rocm71"` → no stale references survive the rename
- ADR-101's `configure.py embedding --device` reference → verified the flag exists in `operator/configure.py:774`
- `kg-api:latest` multi-arch (amd64+arm64) → Apple Silicon path through `mac → latest` is sound

Issue #417, opened 2026-05-26
Ingestion concurrency: per-provider semaphore (ADR-049) never wired; single-job chunks serial by design


aaronsb:
## Symptom

Ingestion of a large markdown file appears to dispatch chunks strictly serially. Observed live in `kg-api-dev` logs:

```
04:01:27 [Chunk 17] 848 words   → 04:01:45  18s extraction
04:01:47 [Chunk 18] 965 words   → 04:03:04  77s extraction
04:03:10 [Chunk 19] ...
```

No `429` errors, no retry-after warnings — just raw Anthropic Sonnet API latency on `max_tokens=16384`, one chunk at a time. The DB has `kg_api.ai_extraction_config.max_concurrent_requests = 8`, but raising it changes nothing.

## Investigation

Two distinct findings — they look like one bug but they are two.

### Finding 1 — Within-job serial chunks is **by design**

`api/app/workers/ingestion_worker.py:518`:

```python
for i, chunk in enumerate(chunks, 1):
    ...
    recent_concept_ids = process_chunk(
        chunk=chunk,
        ...
        recent_concept_ids=recent_concept_ids,   # ← sequential dependency
        ...
    )
```

The loop is plain `for`. Each `process_chunk` returns `recent_concept_ids` that feed into chunk N+1's call so the LLM can deduplicate against concepts that chunk N just created. Parallelizing naively breaks dedup quality (chunk N+1 extracts duplicates of chunk N's concepts because it never saw them).

This is consistent with `1ce64eee feat(ADR-031): Add service token authorization and worker concurrency`, which placed the `ThreadPoolExecutor` at the **job-queue level** (multiple ingestions in parallel), not at the chunk level (one ingestion's chunks in parallel). The design assumption is: scale via more jobs, not bigger jobs.

That assumption breaks when the user has one large document. They feel `N × ~30-80s` wall time with nothing they can tune.

### Finding 2 — Per-provider semaphore (ADR-049) is never wired

`api/app/lib/rate_limiter.py:199` defines `get_provider_semaphore(provider_name)`. ADR-049 §2 says it gates concurrent API calls per provider — "OpenAI 8, Anthropic 4, Ollama 1".

Pickaxe across all branches finds **zero callers**, ever:

```bash
git log --all -S "get_provider_semaphore(" --pickaxe-regex
# (empty)
```

[...]

Issue #252, opened 2026-01-31
feat(ontology): per-source affinity routing during demotion execution


aaronsb:
## Context

ADR-200 Phase 4 sends all sources from a demoted ontology to a single absorption target. The ADR describes a more granular approach where each source is routed to its individual best-affinity target.

## What

Currently `dissolve_ontology()` moves all sources to one target. The ADR (Section 8) specifies:

| Signal | Reassignment |
|--------|-------------|
| Concept edges primarily into Ontology A | Reassign to A |
| Concept bridges A and B roughly equally | Assign to whichever has higher grounding confidence |
| Concept is orphaned (weak edges everywhere) | Return to primordial pool |

Implement per-source routing:
- For each source in the dying ontology, compute which other ontology its concepts have the strongest affinity to
- Route each source individually to its best match
- Sources with no clear affinity go to primordial pool
- This replaces the single-target `dissolve_ontology()` call with a per-source `reassign_sources()` loop

## Trade-offs

- **Performance:** N affinity queries vs 1 bulk move. Mitigated by using Phase 5 materialized edges as a first-pass filter.
- **Complexity:** More graph traversal per demotion. Background worker absorbs the latency.
- **Correctness:** Better ontology coherence post-demotion since sources land where they actually belong.

## References

- ADR-200 Section 8: Demotion
- `api/app/services/proposal_executor.py` — `execute_demotion()`
- `api/app/lib/age_client.py` — `reassign_sources()`, `get_cross_ontology_affinity()`

aaronsb:
## Status update post-merge of #416 (ADR-200/206 clarification)

The Framing A vs B decision is resolved by the merged ADR: per-source affinity routing is now native to `DISSOLVE` (ADR-206 line 126), with `force_primordial` as the defended override. No 8th verb.

**Current code state** (as of `ed7e830d`):

- `dissolve_ontology()` at `api/app/lib/age_client/ontology_scoring.py:527` is still single-target — accepts one `target_ontology` and routes ALL sources via `reassign_sources()` (line 596).
- `get_cross_ontology_affinity()` exists at line 180 of the same file but is only consulted from `proposal_executor._determine_absorption_target()` (line 361) as a fallback target picker, not for per-source routing.
- No `force_primordial` parameter anywhere.

**Plan:** This work folds into a single bundle PR alongside the 2-verb → 6-verb migration and the rest of the annealing-worker integration (per session decision — interconnectedness justifies one PR with multiple commits).

[...]

Issue #241, opened 2026-01-30
Graph integrity worker: detect and heal structural edge misuse


aaronsb:
## Context

During ADR-200 Phase 3a implementation, we discovered that the lifecycle scoring queries need to be edge-agnostic — they traverse `(c:Concept)-->(s:Source)-[:SCOPED_BY]->(o:Ontology)` rather than hard-coding structural edge names like `[:APPEARS]`.

This raised a question: what happens if an LLM generates a structural edge name (SCOPED_BY, APPEARS, EVIDENCED_BY, FROM_SOURCE) as a semantic relationship between Concept nodes? Apache AGE has no schema constraints preventing this. The lifecycle queries wouldn't break (they match on node types, not just edge labels), but it creates data integrity issues:

- VocabType pollution: structural names appearing as vocabulary types
- Confusing graph topology: edges with infrastructure meaning used semantically
- Grounding contamination: epistemic measurement mixing structural and semantic edges

## Proposal

A **graph integrity worker** that periodically detects and heals structural edge misuse.

### Detection

Flag edges where a structural edge name connects wrong node types:

| Edge Name | Expected Pattern | Illegal Pattern |
|-----------|-----------------|-----------------|
| SCOPED_BY | Source → Ontology | Concept → Concept |
| APPEARS | Concept → Source | Concept → Concept |
| EVIDENCED_BY | Concept → Instance | Concept → Concept |
| FROM_SOURCE | Instance → Source | Concept → Concept |

Also flag: structural edge names appearing as VocabType nodes.

### Healing

- Replace illegal structural edges with `SIMILAR_TO` at low confidence (0.3) — safest semantic default when original intent is uncertain
- Remove VocabType nodes for structural edge names
- Log all replacements as warnings

### Phasing

- **Now (Phase 3a)**: Document as design intent, edge-agnostic queries already defend against this
- **Future phase**: Implement as periodic worker alongside the breathing automation
- The manual `check_integrity` command could get detection (no healing) sooner

## Related

- ADR-200 Phase 3a (ontology breathing controls)
- PR #240 (edge-agnostic query refactor)
- ADR-052 (vocabulary consolidation)

aaronsb:
## Status update post-merge of #416 (ADR-200/206 clarification)

**Current code state** (as of `ed7e830d`):

[...]

PR #418, opened 2026-05-26
feat(adr-206): 6-verb annealing vocabulary end-to-end

### LINKED ISSUE #249: feat(ontology): Bezier curve ecological pressure for promotion/demotion thresholds
## Context

ADR-200 Phase 4 defers dynamic threshold adjustment based on ecological ratio. The infrastructure exists (ADR-046 Bezier curves, Phase 4's ecological snapshot), but the feedback loop that adjusts promotion/demotion pressure is not yet wired.

## What

- Use the ecological snapshot (`total_concepts`, `total_ontologies`, `avg_per_ontology`) to dynamically adjust `demotion_threshold` and `promotion_min_degree` per breathing cycle
- When primordial pool grows too large relative to named ontologies → increase promotion pressure (lower `promotion_min_degree`)
- When ontologies get too small → increase absorption pressure (raise `demotion_threshold`)
- Reuse Bezier curve infrastructure from ADR-046 vocabulary aggressiveness profiles

## Depends on

- ADR-200 Phase 4 (ecological snapshot computation)
- ADR-046 (Bezier curve infrastructure)

## References

- ADR-200 Section 9: Resolution Limit (Ecological Ratio)
- `api/app/services/breathing_manager.py` — `_get_ecological_snapshot()`
- `api/app/lib/ontology_scorer.py`

aaronsb:
Implements the ADR-206 closed annealing-action vocabulary end-to-end: schema → manager → executor → CLI → web UI. Closes #249 (both parts) and #252.

## What

Replaces the 2-verb `{promotion, demotion}` annealing model with the symmetric 6-verb closed vocabulary plus the Opus-tier `ADJUST_CONTROL` meta-action:

| Verb | Role |
|---|---|
| `CLEAVE` | Split a parent ontology around an anchor concept (target=new or target=existing) |
| `DISSOLVE` | Per-source affinity routing with `force_primordial` override (#252) |
| `MERGE` | Fold ≥2 donor ontologies into a target |
| `RENAME` | Rename + redescribe a named ontology |
| `NO_ACTION` | Record a deliberate no-op with reasoning |
| `ESCALATE` | Tier-1 admission of uncertainty (Phase 2 chain wiring deferred) |
| `ADJUST_CONTROL` | Bezier-derived self-regulation of Phase-3 control knobs |

The LLM drives the choice through a provider-neutral `call_with_tools` facade (OpenAI / Anthropic / OpenRouter / Ollama / llama.cpp). Legacy `promotion` / `demotion` rows still execute — they alias to `CLEAVE` / `DISSOLVE` on read.

## Commit sequence

1. `884bc9ce` schema 067 — proposal_type CHECK expanded to admit 7 verbs + deprecated 2; `proposal_kind` discriminator + `params` JSONB
2. `23c426a6` `ProposalType` enum + alias reader (`promotion` → `CLEAVE`+primordial; `demotion` → `DISSOLVE`)
3. `f797eed8` `AIProvider.call_with_tools` facade across 5 providers + tests
4. `c7b2b524` 7 tool schemas matching ADR-206 §1 parameter shapes
5. `b292b865` `AnnealingDecisionService` — builds context, drives LLM, parses tool call
6. `5b9bbf07` `AnnealingManager` switches to the decision service; deletes the old boolean evaluator
7. `7c7da6d8` Worker normalizes legacy verbs; executor implements all 6 + DISSOLVE per-source (#252)
8. `f6f1de0d` `pressure_recommendation` block in `_get_ecological_snapshot` (#249 Part 1)
9. `6eb13240` `ADJUST_CONTROL` auto-emission past deadband; executor enforces safety-rail allow-list (#249 Part 2)
10. `6f25aadd` CLI — verb glyphs + params highlights + Params section (`kg ontology proposals`)
11. `99fedb9c` Web UI — verb-aware proposal cards
12. `07263073` docs cross-references (MCP enum, SCHEDULED-JOBS diagram)
13. `7e58210d` schema 068 — `annealing_pressure_history`; manager hook + two read endpoints
14. `f046292f` Web UI — SVG Bezier curve panel + per-control recommendations

## Closes

- #249 (Part 1 read-out + Part 2 ADJUST_CONTROL self-regulation)
- #252 (per-source DISSOLVE routing + `force_primordial`)

## Out of scope (follow-ups)

- #241 — graph integrity worker (different worker, intentionally separate)
- Phase 2 escalation chain wiring (Opus / Human) — ESCALATE currently records to the audit ledger only
- Pressure-history trend chart — endpoint already serves the data, just no chart component yet
- File-size hygiene: `api/app/lib/ai_providers.py` (3003 lines) grew past the priority threshold after the facade landed; extraction into `api/app/lib/providers/` is a focused follow-up
- #417 — `get_provider_semaphore` (ADR-049) was never wired; surfaced during this work, filed separately

## Tests

- 258/258 unit tests green inside `kg-api`
- Migration 067 + 068 applied to dev DB and verified via inline probes
- CLI builds cleanly (tsc + doc gen)
- Web UI builds cleanly (vite)

## Test plan

- [ ] `docker exec kg-api pytest tests/unit/ -x -q` passes
- [ ] Run an annealing cycle in dev: verify 6-verb proposals land in `kg_api.annealing_proposals` with `params` populated
- [ ] CLI: `kg ontology proposals` renders verb glyphs + params highlights; `kg ontology proposal <id>` shows full Params section
- [ ] Web UI: ontology admin tab shows the new proposal cards + the Bezier pressure panel populates after a cycle runs
- [ ] Run a second cycle; verify ADJUST_CONTROL proposals emit only when |delta| ≥ deadband and don't double-emit for an in-flight key


api/app/routes/ontology.py
```diff
@@ -331,40 +337,56 @@ async def list_ontologies(
...
+
+    return AnnealingProposal.from_row({
+        "id": row["id"],
+        "proposal_type": row["proposal_type"],
+        "proposal_kind": row.get("proposal_kind", "ontology"),
+        "ontology_name": row["ontology_name"],
+        "anchor_concept_id": row.get("anchor_concept_id"),
+        "target_ontology": row.get("target_ontology"),
+        "reasoning": row["reasoning"],
+        "params": params_raw,
```
aaronsb, line 370: **Must-fix**: this reader expects `params` in the row dict, but the three SELECTs that feed it (list at ~423–433, get-by-id at ~462–471, review-UPDATE RETURNING at ~530–539) all omit `params` and `proposal_kind` from the column list. `from_row` runs `normalize_proposal_type(type, None)` and assigns `payload["params"] = merged_params or None`.

For v2 rows (CLEAVE/MERGE/RENAME/ESCALATE/ADJUST_CONTROL with structural params JSONB), the API response will always emit `params=None`. The CLI/web types declare these fields (`web/src/types/annealing.ts:28,34`, `cli/src/types/index.ts:589,596`) and the CLI formatter renders param highlights from them — those highlights will be silently empty in production.

The tests in `tests/api/test_ontology_routes.py` use only `SAMPLE_PROPOSAL_ROW` (proposal_type='demotion') and `SAMPLE_PROMOTION_ROW` (proposal_type='promotion'). For both, the alias delta synthesizes `params` from the alias table — masking the regression.

Fix: add `params, proposal_kind` to all three column lists. Then add a v2 test fixture (CLEAVE with `params={anchor_concept_id, target: {kind: 'new', new_name: '…'}}`) and assert the params flow through. Without that fixture this regression will recur.

api/app/models/ontology.py
```diff
@@ -13,6 +13,87 @@ class LifecycleState(str, Enum):
...
+
+# ADR-206 §Aliases — historical names → 6-verb vocabulary.
+#
+# Each entry yields the canonical ProposalType plus a params delta that
+# materializes the implicit context the old name carried. Callers merge
+# the delta into the row's params JSONB before treating the row as a
+# v2 proposal.
+_DEPRECATED_ALIASES: Dict[str, Tuple[ProposalType, Dict[str, Any]]] = {
+    # promotion: CLEAVE applied to the primordial pool.
+    "promotion": (ProposalType.CLEAVE, {"source_ontology": "primordial"}),
```
aaronsb: **Must-fix**: this delta unconditionally injects `source_ontology=primordial`, but in v1 the parent ontology for a promotion was the row's `ontology_name`, not primordial. Verified against `main:api/app/services/proposal_executor.execute_promotion`:

```python
parent_ontology = proposal.get("ontology_name")
...
source_ids = self.client.get_first_order_source_ids(anchor_id, parent_ontology)
```

After the alias normalization, a legacy `promotion` row stored with `ontology_name="Big Topic"` carries `params.source_ontology="primordial"`. The new `execute_cleave` reads `params.get("source_ontology") or proposal.get("ontology_name")` — the alias wins, `"primordial"` is used, and `get_first_order_source_ids(anchor_id, "primordial")` looks up sources in the wrong ontology. Zero sources reassigned; the ontology is created but empty.

For installs where legacy promotions came from non-primordial parents (which is the general case — promotions were the v1 mechanism for splitting *any* ontology around a high-degree concept), this silently breaks legacy in-flight execution — the exact scenario the PR description claims is supported.

Fix options:

1. Drop the delta entirely so `execute_cleave`'s `params.get("source_ontology") or proposal.get("ontology_name")` falls back to the typed column:
   ```python
   "promotion": (ProposalType.CLEAVE, {}),
   ```

2. Or change `normalize_proposal_type` to accept the full row (or pass `ontology_name` in), and have the promotion mapping inject `source_ontology` from `ontology_name` rather than a constant.

3. If primordial really is the intended default for *all* legacy promotions, document that assumption explicitly and add a test that exercises a legacy non-primordial promotion to confirm — but that contradicts what `execute_promotion` in `main` actually read.

api/app/services/annealing_manager.py
```diff
@@ -405,19 +479,44 @@ def _get_open_proposal_targets(self) -> Tuple[Set[str], Set[str]]:
...
+                               params
                         FROM kg_api.annealing_proposals
-                        WHERE status IN ('pending', 'approved', 'executing')
+                        WHERE proposal_type IN (
+                                'promotion', 'CLEAVE',
+                                'demotion', 'DISSOLVE',
+                                'MERGE', 'RENAME'
+                              )
+                          AND status IN ('pending', 'approved', 'executing')
                     """)
```
aaronsb, line 494: **Should-fix**: the IN list covers CLEAVE/DISSOLVE/MERGE/RENAME (plus the legacy strings), but not NO_ACTION or ESCALATE. An ESCALATE proposal for a high-degree concept records audit info but never creates an `ANCHORED_BY` edge — so `anchored_concept_ids` doesn't gate the concept either. Next cycle: same concept ranks high → another ESCALATE → another row.

Same failure mode for NO_ACTION on a demotion candidate (the ontology still ranks low next cycle → another NO_ACTION).

Fix: extend the IN list and add params-aware keying so NO_ACTION/ESCALATE on the same anchor/ontology block re-emission for some cooldown window — either until the row goes terminal, or for N epochs. The existing `failure_cooldown_epochs` knob would be a natural home.

api/app/services/proposal_executor.py
```diff
@@ -249,131 +365,659 @@ def execute_demotion(self, proposal: Dict[str, Any]) -> Dict[str, Any]:
...
+    # explicitly rather than "everything except this set" because adding a
+    # new safety rail without amending this list would silently widen
+    # autonomy — the system cannot widen its own autonomy is the
+    # self-regulation invariant.
+    _SAFETY_RAIL_KEYS = frozenset({
+        "automation_level",
+        "escalation_chain",
+        "opus_confidence",
+        "phone_a_friend_cost_budget",
+    })
```
aaronsb, line 925: **Should-fix**: the inline comment claims this is the self-regulation invariant — "the system cannot widen its own autonomy". But the implementation is a **deny-list**, not an **allow-list**:

```python
if control_key in self._SAFETY_RAIL_KEYS:
    return {success: False, ...}
# else: allowed
```

Adding a new safety-sensitive option to `kg_api.annealing_options` *without* amending `_SAFETY_RAIL_KEYS` would silently widen Opus's autonomy — exactly the failure mode this code warns against.

An allow-list would match the stated invariant:

```python
_TUNABLE_CONTROL_KEYS = frozenset({
    "failure_cooldown_epochs",
    "max_proposals_per_cycle",
    "min_activity_for_cycle",
})
if control_key not in self._TUNABLE_CONTROL_KEYS:
    return {
        "success": False,
        "error": f"control_key '{control_key}' is not Opus-tunable",
        "rejected_reason": "not_allow_listed",
    }
```

The three keys above are exactly what `_load_phase3_controls` and `_build_pressure_recommendation` already reference, so the allow-list has a natural single source of truth — adding a new tunable knob means editing one place, not three (annealing_options seed + pressure recommendation + this list).

api/app/services/proposal_executor.py
```diff
@@ -36,171 +45,290 @@
...
+            if not existing_node:
+                return {
+                    "success": False,
+                    "error": f"CLEAVE target '{existing_name}' does not exist",
+                }
+            target_name = existing_name
+
+        # ---- Cluster materialization ----
+        cluster_selection = params.get("cluster_selection", "first_order")
+        cluster_params = params.get("cluster_params") or {}
```
aaronsb: Nice-to-have: `cluster_params` is assigned here but never read again — `first_order` (the only implemented strategy) doesn't take strategy params, and the `embedding_radius` / `named_concepts` branches both early-return before any use. Either delete the assignment, or thread it through when the other strategies land.

tests/api/test_ontology_routes.py
```diff
@@ -1174,7 +1174,14 @@ def test_list_empty(self, api_client, auth_headers_user):
...
-        """List returns proposal data."""
+        """
+        List normalizes legacy proposal_type names on read (ADR-206).
+
+        Historical rows stored as `demotion`/`promotion` surface as their
+        canonical 6-verb vocabulary in API responses (DISSOLVE / CLEAVE).
+        The alias delta for `promotion` injects the implicit
+        source_ontology='primordial' into params.
+        """
         client = mock_proposals_cursor([SAMPLE_PROPOSAL_ROW, SAMPLE_PROMOTION_ROW])
```
aaronsb, line 1286: **Should-fix** (test gap): this fixture covers only v1 row shapes (`SAMPLE_PROPOSAL_ROW` = `'demotion'` row, `SAMPLE_PROMOTION_ROW` = `'promotion'` row). Both pass through the alias-table normalization, which synthesizes `params` from the delta — so the tests *don't* catch the bug where the SELECT in `list_proposals` drops the `params` column (see comment on `api/app/routes/ontology.py:_row_to_proposal`).

Adding a v2 fixture would lock the contract:

```python
SAMPLE_CLEAVE_ROW = {
    "id": 3,
    "proposal_type": "CLEAVE",
    "proposal_kind": "ontology",
    "params": {
        "anchor_concept_id": "c_x",
        "source_ontology": "big",
        "cluster_selection": "first_order",
        "target": {"kind": "new", "new_name": "new-domain"},
    },
    "ontology_name": "big",
    ...  # rest of fields
}
```

and asserting `data["proposals"][0]["params"]["target"]["new_name"] == "new-domain"` end-to-end.

schema/migrations/067_annealing_vocab_v2.sql
```diff
@@ -0,0 +1,211 @@
...
+     '10',
+     'Maximum number of proposals (any verb) emitted in a single annealing '
+     'cycle. Throttles decision volume under high ecological pressure. '
+     'Bezier-tunable via ADJUST_CONTROL (ADR-206 Phase 3).'),
+    ('min_activity_for_cycle',
+     '1',
+     'Minimum number of epoch increments since the last cycle before a new '
+     'cycle is eligible. Raises the no-op floor when the graph is quiet. '
+     'Bezier-tunable via ADJUST_CONTROL (ADR-206 Phase 3).')
+ON CONFLICT (key) DO NOTHING;
```
aaronsb, line 119: Cleanly written migration — the inline `DO $$ … END $$;` probes (CLEAVE-valid, ADJUST_CONTROL-valid, deprecated-still-valid, bogus-kind-rejected, seed-rows-present) cover the failure modes well. `ON CONFLICT DO NOTHING` keeps it idempotent.

Small thought: the verification doesn't explicitly probe the `params` column's existence. The probe at line 140 implicitly relies on it by inserting `'{"probe": true}'::jsonb`, which would fail if the column were missing — so the chain works in practice. Just calling it out in case you want belt-and-suspenders via `information_schema.columns`. Not blocking.

aaronsb (commented):
## Review summary — ADR-206 6-verb vocabulary end-to-end

Reviewed the full 14-commit bundle (6,737+/1,342−). Most of it is sound — schema → manager → executor → CLI/web wiring is coherent, SQL throughout is parameter-bound, the safety-rail rejection path is genuinely pre-DB, and the `call_with_tools` facade tests do exercise SDK call shape (not just result shape). Migration probes are inline-verified per project convention.

Two correctness bugs surfaced that I'd want addressed before merge. Both are localized; neither is structural. Several should-fix items can land as follow-ups.

### Must-fix (correctness blockers)

- **API SELECTs drop `params` and `proposal_kind`** — the three SELECTs in `api/app/routes/ontology.py` (list at lines 423–433, get-by-id at 462–471, review-UPDATE RETURNING at 530–539) all omit `params` and `proposal_kind` from the column list. `_row_to_proposal` then routes through `AnnealingProposal.from_row`, which expects `params` in the row dict. Result: every v2 proposal returned by the API loses its params block. Worker dispatch is fine (its own `_load_proposal` selects them), but the CLI and web verb-aware cards depend on `params` to render targets/donors/recommendations. The matching tests use only legacy `promotion`/`demotion` rows whose params get re-synthesized from the alias delta — that's why the regression isn't caught. See inline comment on `_row_to_proposal`.

- **Legacy `promotion` alias forces `source_ontology=primordial` regardless of actual parent** — `api/app/models/ontology.py` line 55. The v1 promotion flow stored the actual parent ontology in `ontology_name` (verified against `main:api/app/services/proposal_executor.execute_promotion`: `parent_ontology = proposal.get("ontology_name")`, then `get_first_order_source_ids(anchor_id, parent_ontology)`). The ADR-206 alias delta unconditionally injects `{source_ontology: "primordial"}`, which wins via `setdefault` for legacy rows (which have empty/NULL params). On execution, `execute_cleave` reads `params.get("source_ontology") or proposal.get("ontology_name")` — gets `"primordial"` first, looks up first-order sources in the primordial pool instead of the real parent, and reassigns zero sources. For any legacy promotion that was *not* from primordial, the cleave silently no-ops. See inline comment on the alias table.

### Should-fix (real issues, follow-up OK)

- **`_get_open_proposal_targets` doesn't include NO_ACTION/ESCALATE** — `api/app/services/annealing_manager.py` IN list at lines 485–489. An ESCALATE proposal for a high-degree concept records audit info but never creates an `ANCHORED_BY` edge, so `anchored_concept_ids` doesn't gate the concept either. Next cycle: same concept ranks high → another ESCALATE. Same pattern for NO_ACTION on a demotion candidate. Queue piles up duplicates that aren't blocked by either mechanism. Inline below.

- **Safety-rail rejection is a deny-list but the comment claims allow-list semantics** — `api/app/services/proposal_executor.py` `_SAFETY_RAIL_KEYS` (lines 907–918). The comment says "the system cannot widen its own autonomy is the self-regulation invariant" — but the implementation is a **deny-list**. Adding a new safety-sensitive option to `kg_api.annealing_options` *without* amending `_SAFETY_RAIL_KEYS` would silently let Opus tune it, exactly the failure mode the comment warns against. Inverting to an explicit allow-list (the three tunable keys the manager actually recommends) would match the stated invariant. Inline below.

- **Empty coverage for v2 rows through the API surface** — `tests/api/test_ontology_routes.py`. The `SAMPLE_PROPOSAL_ROW` / `SAMPLE_PROMOTION_ROW` fixtures are v1 shapes (no `params`, no `proposal_kind`). A test exercising a v2 CLEAVE/MERGE row end-to-end through `/ontology/proposals` would have caught the SELECT regression above. Inline below.

- **Three priority-band files now** — `api/app/lib/ai_providers.py` (3003 lines) is acknowledged in the PR body, but `api/app/services/proposal_executor.py` (1072) and `api/app/services/annealing_manager.py` (1156) also crossed the threshold. Both have natural split lines: executor → one module per verb (`cleave.py`, `dissolve.py`, …); manager → split orchestration + ecological-pressure logic (the route already imports the module-level `_PRESSURE_*` constants and `_pressure_curve_metadata`, so there's a natural seam). Not blocking but worth noting in the follow-up. Inline on annealing_manager below.

### Nice-to-have

- **Dead variable `cluster_params`** in `execute_cleave` — assigned, never read. Either wire it into a future cluster-strategy parameter pass or delete it.
- **String literals vs `ProposalType.X.value`** at writer sites in `annealing_manager._decide_and_store_*` and `_emit_control_proposals` — works because the CHECK constraint catches typos, but using the enum value would let mypy/IDE catch them earlier.

### Things I checked and they're fine

- All INSERT/UPDATE/SELECT in `annealing_manager._store_proposal`, `_emit_control_proposals`, `_record_pressure_snapshot`, `_get_open_*`, and the new `/annealing/pressure[/history]` endpoints use parameter binding. `Json()` wraps JSONB writes. No string interpolation into SQL.
- ADJUST_CONTROL safety-rail rejection is verified pre-DB (test `test_safety_rail_keys_rejected_without_db_touch` asserts `client.pool.getconn.assert_not_called()`); rejection is by exact-string `frozenset` membership — `automation_level_override` can't sneak past `automation_level`.
- Migrations 067/068 both validate via inline `DO $$ … END $$;` probes against constraints + seed rows. Idempotent UPs (`IF NOT EXISTS`, `ON CONFLICT DO NOTHING`).
- Provider `call_with_tools` tests assert on `provider.client.X.call_args.kwargs[…]` — exercising the SDK call shape per the focus item, not just the result shape.
- All seven verb executors handle missing/malformed params with `{success: False, error: …}` returns rather than raising. Verified for CLEAVE (missing anchor, missing target, target collisions), DISSOLVE (missing source_ontology, gone/pinned/frozen, in-flight veto), MERGE (<2 donors, frozen donor, target=donor collision), RENAME (missing names, primordial guard, pinned/frozen), NO_ACTION + ESCALATE (graceful, no graph mutation), ADJUST_CONTROL (missing key/value, unknown key, safety-rail rejection).
- Worker dispatch dictionary covers all 7 verbs; unknown verbs hit `Verb {…} not yet implemented` (not a raise).
- Single-threaded cycle (ADR-031) means `_get_open_*` + `_store_proposal` can't race against another cycle. The executor worker only transitions proposals through statuses that are all in the dedup IN-list, so no double-emit window.
- Every `proposal_type=` writer uses either a literal from the closed set or `decision.proposal_type` (sourced from the LLM tool_name, constrained by the closed tool catalog). No string concatenation from untrusted input.

### OK to merge as-is?

**No, but close.** The two must-fix items are both small focused changes — minutes to fix. Without them, v2 proposals will visibly lose their `params` in CLI/web rendering (cosmetically wrong, the data is still in the DB), and legacy non-primordial promotions will silently no-op on execution (correctness, but only triggers for installs with pre-ADR-206 in-flight `promotion` rows from non-primordial parents).

If this install has no legacy `promotion` rows pending, the second bug is theoretical; if it does, it's a real silent failure. Worth a quick `SELECT COUNT(*) FROM kg_api.annealing_proposals WHERE proposal_type='promotion' AND ontology_name != 'primordial' AND status IN ('pending','approved','executing')` against the dev DB before deciding fix-now vs ship-and-follow-up.

---

PR #426, opened 2026-05-28
fix(gpu): stop over-claiming VRAM on AMD ROCm (and CPU/CUDA)

aaronsb:
## Summary

Five compounding sources of unnecessary memory pressure on the embedding path, most visible on AMD GPUs with limited VRAM.

| # | Fix | File(s) |
|---|---|---|
| 1 | `--workers 2` → `--workers 1` (each uvicorn worker was a separate Python process loading its own copy of the model + ROCm runtime context) | `api/Dockerfile`, `api/Dockerfile.rocm-host`, `docker/docker-compose.dev.yml` |
| 2 | Vision model loads as `torch_dtype=torch.float16` on GPU (was fp32 — silent 2× weight blow-up) | `api/app/lib/visual_embeddings.py` |
| 3 | Drop `device_map=\"auto\"` for vision tower, use explicit `.to(self.device)` (accelerate sharding machinery is overhead for a sub-GB single tower) | `api/app/lib/visual_embeddings.py` |
| 4 | Gate `init_visual_embedding_generator()` on `_is_dispatch_leader` — only the leader runs ingestion, so non-leader workers were loading vision weights they would never use. Dispatch lock acquisition moved earlier in startup. | `api/app/main.py` |
| 5 | `PYTORCH_HIP_ALLOC_CONF=expandable_segments:True` on both ROCm compose variants — reduces reserved-but-unallocated VRAM under PyTorch's caching allocator | `docker/docker-compose.gpu-amd.yml`, `docker/docker-compose.gpu-amd-host.yml` |

Diagnosis caught during this work: the running API image on AMD-host can silently be a **CUDA-built torch** if a prior CPU build cached \`kg-api:latest\` (compose's local-build tag is variant-agnostic). Surfaced separately — not fixed in this PR. Worth a follow-up that mirrors the ADR-101 GHCR tag scheme for local builds.

## Test plan

- [ ] On an AMD-host machine, rebuild with \`./operator.sh stop && GPU_MODE=amd-host ./operator.sh start\` (forces fresh build picking up \`Dockerfile.rocm-host\`)
- [ ] Verify in container: \`docker exec kg-api-dev python -c \"import torch; print(torch.__version__, torch.version.hip, torch.cuda.is_available())\"\` — expect a \`+rocm\` torch with \`is_available()=True\`
- [ ] Check uvicorn startup log shows one worker, not two
- [ ] On non-leader worker scenario (if running with >1 worker via override), confirm log line \"Non-leader worker — skipping visual embedding generator\"
- [ ] Ingest a document with images — visual embeddings still get produced (on the leader)
- [ ] Search queries still work (text embedding still loaded on all workers)
- [ ] CPU variant: \`./operator.sh start\` (no GPU mode) — verify API still healthy on \`--workers 1\`

### DIFF
--- a/api/Dockerfile
+++ b/api/Dockerfile
@@ -112,5 +112,9 @@ EXPOSE 8000
 HEALTHCHECK --interval=30s --timeout=5s --start-period=10s --retries=3 \
   CMD curl -f http://localhost:8000/health || exit 1
 
-# Start FastAPI server with uvicorn (multi-worker in both dev and prod)
-CMD ["uvicorn", "api.app.main:app", "--host", "0.0.0.0", "--port", "8000", "--workers", "2"]
+# Start FastAPI server with uvicorn.
+# Single worker: each uvicorn worker is a separate Python process and would
+# load its own copy of any local embedding model (and ROCm/CUDA runtime
+# context on GPU variants) — 2 workers ≈ 2× memory for no async-throughput
+# gain on this workload. Ingestion runs in the queue, not the request path.
+CMD ["uvicorn", "api.app.main:app", "--host", "0.0.0.0", "--port", "8000", "--workers", "1"]
--- a/api/Dockerfile.rocm-host
+++ b/api/Dockerfile.rocm-host
@@ -51,4 +51,4 @@ EXPOSE 8000
 HEALTHCHECK --interval=30s --timeout=5s --start-period=10s --retries=3 \
   CMD curl -f http://localhost:8000/health || exit 1
 
-CMD ["uvicorn", "api.app.main:app", "--host", "0.0.0.0", "--port", "8000", "--workers", "2"]
+CMD ["uvicorn", "api.app.main:app", "--host", "0.0.0.0", "--port", "8000", "--workers", "1"]
--- a/api/app/lib/visual_embeddings.py
+++ b/api/app/lib/visual_embeddings.py
@@ -87,6 +87,7 @@ class VisualEmbeddingGenerator:
     def _load_transformers(self):
         """Load model via transformers AutoModel + AutoProcessor."""
         from transformers import AutoModel, AutoProcessor
+        import torch
 
         load_kwargs = {
             "trust_remote_code": self.trust_remote_code,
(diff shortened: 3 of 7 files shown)


aaronsb:
## Review: GPU memory over-claim fixes

### Blocking concern: fp16 weights + fp32 pixel_values dtype mismatch

`api/app/lib/visual_embeddings.py:104` loads model weights as `torch.float16` on cuda/mps, but `generate_embedding` (line ~206) and `generate_embeddings_batch` (line ~251) do:

```python
inputs = self.processor(images=image, return_tensors='pt').to(self.device)
```

`.to(self.device)` only moves to device — it does **not** change dtype. `pixel_values` from the processor is fp32. Feeding fp32 inputs into fp16 weights typically raises at the first conv/linear:

> `RuntimeError: Input type (torch.FloatTensor) and weight type (torch.HalfTensor) should be the same`

Whether this surfaces depends entirely on whether the Nomic Vision custom forward (`trust_remote_code=True`) explicitly casts `pixel_values` to model dtype, or wraps in `torch.autocast`. I cannot verify that from this repo alone — it lives in the remote model code.

Recommended fix in this PR:

```python
# generate_embedding
inputs = self.processor(images=image, return_tensors='pt').to(self.device)
if self.model.dtype == torch.float16:
    inputs = {k: (v.half() if v.is_floating_point() else v) for k, v in inputs.items()}
```

Or — cleaner — wrap the forward in `with torch.autocast(device_type=self.device, dtype=torch.float16):` and leave weights fp32 (you'd lose half the VRAM savings but eliminate dtype-mismatch risk). At minimum, the PR's test plan should add an explicit "ingest at least one image with GPU enabled and verify no dtype error" check — the existing item "Ingest a document with images" implies this but doesn't call out the fp16 path specifically.

### Correctness: dispatch lock move is safe, but the rationale comment is wrong by default

The lock move from `main.py:396` → `main.py:237` is structurally safe:

- The cold-start lock (`LOCK_ID=100_000_002`, line 98) is independent of the dispatch lock (`LOCK_ID=100_000_001`, line 55), and `_try_acquire_cold_start_lock` is called inside its own try-block on a separate pooled connection (line 295). Moving the dispatch acquisition does not affect cold-start.
- No code between the old position (line 396 in main) and the new position (line 237) reads `_is_dispatch_leader`. `init_embedding_model_manager()`, `validate_api_keys_at_startup()`, `get_embedding_worker()`, `initialize_builtin_embeddings()`, and the resume-interrupted-jobs block all run unconditionally on both leader and non-leader, both before and after this PR.
- No external module references `_is_dispatch_leader` (verified via `grep -rn` of `api/app/`).
- The old duplicate at line 396 is removed cleanly.

However: with `--workers 1` (the new default everywhere), there is exactly one process, it is always the leader, and the optimization is dormant. The visual-skip branch only fires under `--workers N>1`, which the PR has stopped configuring. The rationale comment at `main.py:230-235` (`"so non-leaders ... can skip GPU-heavy resources"`) is accurate but describes a path that is now reachable only under operator override. Worth a one-line note in the comment ("guards the operator-override case").

### Correctness: leader-gating leaks via lazy init on `/ingest/image/health`

`api/app/routes/ingest_image.py:325` calls `check_visual_embedding_health()` → `get_visual_embedding_generator()` (lib/visual_embeddings.py:404), which has a **lazy-load fallback** at lines 442-462: if the module-global `_embedding_generator` is `None` (which it will be on a non-leader after this PR), it reads the active profile and instantiates the model **inline on the HTTP request thread**.

With `--workers 1` this is irrelevant. Under any future override to `--workers >1`, a health-check request landing on a non-leader will defeat the optimization (loading the same ~350MB the PR is trying to skip) and add multi-second latency to that probe. Two options:

1. Cheap: at the call site, check `_is_dispatch_leader` and return a stub `{"status": "healthy", "device": "delegated-to-leader", ...}` for non-leaders.
2. Cleaner: have `get_visual_embedding_generator()` raise a known sentinel ("not initialized on this worker") and have the health route return 200 with `{"role": "non-leader"}`.

Not a blocker for the current PR scope, but flag for follow-up.

### Risk: dev compose comment is now slightly stale

`docker/docker-compose.dev.yml:46-49`:

> Hot reload (`--reload`) is incompatible with `--workers`, so restart the container after code changes

With `--workers 1`, uvicorn actually *would* accept `--reload`. The incompatibility-with-`--workers` rationale was a constraint of the old `--workers 2` setup, not the new `--workers 1`. The comment isn't wrong (`--workers 1 --reload` is still skipped here intentionally — bind-mounted volumes inside a container have caveats), but the *reason given* no longer holds. Tightening would help future-me:

> Single worker (matches production). We don't use `--reload` here because the bind-mounted source has flaky inotify behavior in some Docker setups; restart with `./operator.sh restart api` after code changes.

### Risk: `PYTORCH_HIP_ALLOC_CONF` on a CUDA-built torch — safe

PyTorch's CUDA build silently ignores `PYTORCH_HIP_ALLOC_CONF` (it only reads `PYTORCH_CUDA_ALLOC_CONF`). The var is set only in the two ROCm compose overlays (`gpu-amd.yml` and `gpu-amd-host.yml`), not in the base compose, so it cannot leak into the CPU-only or CUDA stacks. Even if the AMD-host stack ends up running a CUDA-built torch (the "silently cached `kg-api:latest`" follow-up flagged in the PR description), no harm — it's a no-op env var to that runtime. Good.

### Minor: small CPU-path behavior change not called out

The pre-PR CPU path passed `low_cpu_mem_usage=False` to `AutoModel.from_pretrained` (lines 110, 137 of the pre-PR file). The post-PR code drops it — `low_cpu_mem_usage` now defaults to whatever transformers' default is (recently True when accelerate is installed). For CPU loading, this means the model may be loaded via the meta-device path (incremental materialization), which is a minor semantics change for CPU users — they'll see slightly different memory behavior during load, though end state is the same. Not a bug, but the PR description claims point #3 is purely about GPU path; this is a CPU-path side effect worth noting in the commit message or restoring the `low_cpu_mem_usage=False` explicitly for the CPU branch.

### Minor: MPS + fp16

The `on_accelerator = self.device in ('cuda', 'mps')` branch enables fp16 on Apple Silicon too. MPS support for fp16 ViT ops has been bumpy historically (LayerNorm in older PyTorch, certain attention kernels). Most current PyTorch versions are fine, but dev Mac users may want a `KG_DISABLE_VISION_FP16` escape hatch if a regression turns up. Not blocking — just worth a follow-up if a Mac user reports image-ingest errors.

### Nits

- `main.py:230-235`: comment says "~350MB weights + accelerate/processor overhead" — accelerate overhead is now gone (per point #3 of the PR), so the phrasing is slightly stale. Minor.
- `visual_embeddings.py:135`: comment "matching the text loader pattern (see embedding_model_manager.py)" is structurally true for `.to(device)` but the text loader does **not** also pass `torch_dtype=float16` to the weights (it stays fp32 and post-casts the output). Slightly misleading framing — vision is more aggressive than text on dtype.

### Things I checked and found clean

- `_try_acquire_cold_start_lock` (LOCK_ID 100_000_002) is independent of dispatch lock (100_000_001), separate pooled connections, separate held-conn globals. No cross-contamination.
- No remaining references to `device_map`, `hf_device_map`, `dispatch_model`, or `accelerate`-style hooks anywhere in `api/app/`. Removing `device_map="auto"` doesn't strand any downstream consumer.
- Shutdown logic (`main.py:438+`) gates on `_is_dispatch_leader` the same way before and after — moving acquisition earlier doesn't change the released-at-process-exit semantics.
- The resume-interrupted-jobs block (`main.py:347-389`) runs unconditionally on both workers in both pre- and post-PR code — not changed by this PR, not made worse, but the underlying race is pre-existing.

aaronsb:
Addressed in 5f2f794:

- **Blocking #1 (dtype mismatch)**: capture `self._model_dtype` after load; new `_to_model_inputs` helper casts only floating-point tensors to model dtype on the way to the device, used by both `generate_embedding` and `generate_embeddings_batch`. fp16 weights + fp32 pixel_values would have raised on the first GPU image ingest.
- **Output contract**: cast `last_hidden_state` to fp32 before `.cpu().numpy()` so downstream consumers still see fp32 arrays.
- **Minor (`low_cpu_mem_usage=False`)**: restored on CPU path with a note pinning it against HF default drift.

Skipped:
- Latent issue (`get_visual_embedding_generator()` lazy-load at `routes/ingest_image.py:325`): real but only kicks in if someone overrides to `--workers >1` after this PR. Leaving for a follow-up rather than coupling it to this fix.
- Comment phrasings: reworked the `.to(device)` rationale comment in `_load_transformers` to call out that the dtype side diverges from the text loader (vision needs the cast, text doesn't).

Test plan addendum: GPU image ingestion is the load-bearing test for this PR now.

PR #427, opened 2026-05-28
fix(teardown): stop orphaning anonymous volumes on --full

aaronsb:
## Summary

`./operator.sh teardown --full` was leaving behind anonymous (SHA-named) docker volumes containing PG18 cluster data, so a fresh `start` would re-attach to old jobs/state. Found while debugging "previous jobs show up in a new instance."

Two root causes:

| # | Bug | Fix |
|---|---|---|
| 1 | `compose down` invoked with only base + dev overlays. With `GPU_MODE=amd-host` set, the actually-running stack used a GPU overlay this script never saw. Compose couldn't reach those services to detach their anonymous volumes, which orphaned as SHA leftovers when the next step force-removed the containers. | Source `common.sh` and use `run_compose` so teardown uses the **same overlay stack the platform was started with** (honors `GPU_MODE` / `DEV_MODE` / `IMAGE_SOURCE` from `.operator.conf`). |
| 2 | Manual volume sweep grepped only named patterns. SHA-named anonymous volumes can't match by name. | Second pass: list dangling volumes (`docker volume ls -qf dangling=true`) with their compose-project label or mountpoint, prompt before removal. Auto-deleting on a shared host would risk nuking unrelated projects (e.g. this developer's box has `sable_sable-pgdata` and `whisper-service_whisper-temp` sitting alongside ours). |

Also:
- Always pass `--remove-orphans` to `compose down` so future runs catch services that left the active overlay.
- Expand the named-volume grep to include the standalone/prod `kg_` prefix.
- `-y/--yes` propagates to the new dangling-volume prompt.

## Test plan

- [ ] On a host with `GPU_MODE=amd-host` and the platform up: `./operator.sh teardown --full`, confirm:
  - It picks up the gpu-amd-host overlay (visible from the compose down line)
  - The dangling-volume prompt appears if there are any leftovers
  - After accepting, `docker volume ls -qf dangling=true` is clean (or contains only volumes the user opted to keep)
- [ ] On a CPU-only host: same flow, no GPU overlay, still works
- [ ] With `-y/--yes`: prompt is skipped, dangling volumes removed automatically
- [ ] Decline the prompt: dangling volumes preserved, message points to manual cleanup
- [ ] Without `--full` (default): no dangling-volume pass runs (preservation default unchanged)

### DIFF
--- a/operator/lib/teardown.sh
+++ b/operator/lib/teardown.sh
@@ -71,7 +71,10 @@ while [[ $# -gt 0 ]]; do
             echo "  --remove-volumes    Also remove ALL volumes including caches"
             echo "  --include-operator  Also teardown the operator container (default: preserve)"
             echo "  --full              Complete teardown: images + volumes + operator"
-            echo "  -y, --yes           Skip confirmation prompt"
+            echo "  -y, --yes           Skip the initial confirmation AND the"
+            echo "                      dangling-volume prompt for kg-owned"
+            echo "                      volumes. Volumes that look unrelated"
+            echo "                      to this project are still skipped."
             echo "  -h, --help          Show this help"
             echo ""
             echo "Default Behavior:"
@@ -208,14 +211,48 @@ fi
 # left services unreferenced from compose's perspective — and their
 # anonymous volumes orphaned as SHA-named leftovers that the manual sweep
 # below can't grep for.
+COMPOSE_DOWN_RAN=false
 if [ -f "$SCRIPT_DIR/common.sh" ]; then
+    # Source under `set +e` so a malformed .operator.conf from an interrupted
+    # init doesn't abort the whole teardown before any cleanup runs. The
+    # fallback path below covers the case where sourcing or config load fails.
+    set +e
     # shellcheck source=operator/lib/common.sh
-    source "$SCRIPT_DIR/common.sh"
-    load_operator_config
-    COMPOSE_CMD=$(get_compose_cmd)
-    $COMPOSE_CMD down $DOWN_FLAGS 2>/dev/null || true
-else
-    # Fallback: best-effort base + dev overlay, no env file
+    source "$SCRIPT_DIR/common.sh" 2>/dev/null
+    src_rc=$?
+    if [ $src_rc -eq 0 ]; then
+        load_operator_config 2>/dev/null
+        cfg_rc=$?
+    else
+        cfg_rc=1
+    fi
+    set -e
+
+    if [ $src_rc -eq 0 ] && [ $cfg_rc -eq 0 ]; then
+        COMPOSE_CMD=$(get_compose_cmd 2>/dev/null)
+        # get_compose_cmd unconditionally appends `--env-file $ENV_FILE`. If
+        # the .env was already removed (e.g. a prior partial teardown without
+        # --keep-env), `docker compose` will refuse with non-zero exit and
+        # `|| true` would silently mask the failure — leaving services up and
+        # their anonymous volumes orphaned. Strip the missing env-file flag
+        # rather than fail the whole teardown.
+        if [ ! -f "$PROJECT_ROOT/.env" ]; then
+            COMPOSE_CMD=$(echo "$COMPOSE_CMD" | sed -E 's| --env-file [^ ]+||')
+            echo -e "  ${YELLOW}Note: .env missing, running compose without --env-file${NC}"
+        fi
+        if [ -n "$COMPOSE_CMD" ]; then
+            $COMPOSE_CMD down $DOWN_FLAGS 2>/dev/null || true
+            COMPOSE_DOWN_RAN=true
+        fi
+    else
+        echo -e "  ${YELLOW}Could not source common.sh / load .operator.conf — using fallback${NC}"
+    fi
+fi
+
+if [ "$COMPOSE_DOWN_RAN" = false ]; then
+    # Fallback: best-effort base + dev overlay, no env file. We deliberately
+    # don't `cd` (we're already in $DOCKER_DIR) and don't pass --env-file
+    # since we can't trust .operator.conf in this path.
     docker compose -f docker-compose.yml down $DOWN_FLAGS 2>/dev/null || true
     if [ -f "$DOCKER_DIR/docker-compose.dev.yml" ]; then
         docker compose -f docker-compose.yml -f docker-compose.dev.yml down $DOWN_FLAGS 2>/dev/null || true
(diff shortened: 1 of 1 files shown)


aaronsb:
## Review

The diagnosis is solid and the fix shape is right (use the overlay set the platform actually started with; second-pass sweep for SHA-named anonymous volumes that compose can't reach). The concerns below are about *new* edge cases the rewrite opens up rather than the central approach.

---

### 1. `set -e` interaction with `source common.sh` — real, but bounded

`teardown.sh:2` runs `set -e`. After sourcing `common.sh`, every helper call (`load_operator_config`, `get_compose_cmd`) inherits it.

- `load_operator_config` (common.sh:24) does `if [ -f "$CONFIG_FILE" ]; then source "$CONFIG_FILE"; fi`. `.operator.conf` is shell. If a stale install has a malformed line in it (e.g. a partially-written value from an interrupted `guided-init`), sourcing aborts under `set -e` and teardown dies before doing anything. The pre-PR code never read `.operator.conf` at teardown time, so this is new exposure.
- `image-tag.sh` is sourced via `source "$_COMMON_DIR/image-tag.sh"` (common.sh:44) only when `KG_API_IMAGE_TAG` is unset. On this dev's box `.operator.conf` already sets `KG_API_IMAGE_TAG=rocm72-host` so the branch is skipped, but on a host where it isn't set, any failure inside `derive_kg_api_image_tag` propagates and kills teardown.

The PR description's whole point is "teardown was leaving stuff orphaned." A teardown that *aborts early on a stale config* is arguably worse than one that runs the wrong overlay — the user came here specifically to clean up broken state.

**Suggested mitigation**: wrap the sourcing/loading block so a partial failure falls through to the existing fallback path rather than killing the script. Something like:
```bash
if [ -f "$SCRIPT_DIR/common.sh" ] && \
   ( set +e; source "$SCRIPT_DIR/common.sh" && load_operator_config && COMPOSE_CMD=\$(get_compose_cmd) ); then
    \$COMPOSE_CMD down \$DOWN_FLAGS 2>/dev/null || true
else
    # existing fallback
fi
```
or simply `set +e` before the source block and `set -e` after.

---

### 2. Quoting on `\$COMPOSE_CMD` (teardown.sh:216) — fine here, fragile shape

`get_compose_cmd` returns a string like `docker compose -f /home/aaron/.../docker-compose.yml -f ... --env-file /home/aaron/.../.env`. `$COMPOSE_CMD down $DOWN_FLAGS` (unquoted) relies on word-splitting on whitespace. **It will break the moment `PROJECT_ROOT` contains a space.** The `cd "\$PROJECT_ROOT"`-style installs in `/Users/Some Name/...` (macOS) or `/c/Program Files/...` (mingw) would silently produce a corrupted compose invocation here.

The existing call sites in `start-infra.sh:47`, `stop.sh:79`, `upgrade.sh:234` use the same pattern (or `run_compose` which has the same hazard at `\$compose_cmd "\$@"`), so this is a pre-existing project convention, not a PR regression. Still worth a one-line comment at teardown.sh:215 noting the assumption — or switch to `run_compose down \$DOWN_FLAGS` for consistency with the rest of the codebase, which is what new code should be doing per the established pattern.

---

### 3. Dangling-detection timing — correct sequence, one subtle gap

The order is: compose down (line 216) → force-remove leftover containers (lines 227-242) → list dangling volumes (line 280). By the time line 280 runs, our containers are gone, so volumes that were attached to them are now dangling. That part is right.

**The gap**: `INCLUDE_OPERATOR=false` (any teardown short of `--full`) deliberately preserves `kg-operator` (line 232 greps it out of the kill list). The operator container has a bind mount of the project root and may have anonymous volumes for `/root` or `/tmp` mounts. Those will *not* appear as dangling because the operator is still running, so the existing logic is fine.

**But**: `--full` sets `INCLUDE_OPERATOR=true` (line 56). The operator container is force-removed at lines 234-238 *before* line 280 runs, so by the time we list dangling volumes, the operator's anonymous volumes are also in the list. That's by design (you want them gone in `--full`) but it means the dangling list on `--full` is strictly larger than just "kg leftovers."

Not a bug, just worth noting in the comment block at teardown.sh:271-279 so the next person reading it doesn't wonder why operator volumes are showing up under "kg leftovers."

---

### 4. Blast radius under `-y/--yes` (teardown.sh:298-300) — the real risk

This is the one I'd push back on hardest. With `--yes`:

```bash
DANGLING_VOLS=\$(docker volume ls -qf dangling=true 2>/dev/null || true)
# ... no project filtering ...
docker volume rm "\$vol"
```

On a shared dev box / CI runner / workstation with other compose projects (the PR description literally names `sable_sable-pgdata` and `whisper-service_whisper-temp`), `operator.sh teardown --full -y` will silently delete unrelated projects' data. The interactive path is safe (user sees the list). The `-y` path is not.

**The fix isn't to remove `-y` support** — CI needs it. The fix is to filter by ownership we're confident about, and only auto-yes within that filter:

- Anonymous volumes (no compose project label) with mountpoint under `/var/lib/docker/volumes/<sha>/_data/18/docker/` or whose mount contains evidence of our services — keep.
- Anonymous volumes with no PG/garage/hf evidence — they could be ours from a long-dead container, or someone else's. **Don't auto-delete under `-y`**, only delete interactively.
- Volumes labeled `com.docker.compose.project=<our project>` — safe to auto-delete.
- Volumes labeled with someone else's project — never auto-delete, even interactively.

A practical compromise: under `-y`, restrict to volumes whose compose project label is in our known set (`knowledge-graph`, `docker`, `kg`, the configured `CONTAINER_PREFIX`), plus anonymous volumes whose mountpoint inspection mentions postgres/garage/hf. Anything else gets listed but not auto-deleted, with a warning. The current code's blast radius is "every dangling volume on the host" and that's a footgun in CI.

The help text at teardown.sh:74 says `-y, --yes  Skip confirmation prompt` (singular). Now there are *two* prompts and `-y` answers yes to both, with very different consequences. At minimum, update the help to call this out explicitly. Better: make the dangling-volume prompt require a separate flag like `--prune-dangling` so `-y` alone doesn't widen the blast radius.

---

### 5. `--env-file` failure when `.env` is gone

`get_compose_cmd` unconditionally appends `--env-file \$ENV_FILE` (common.sh:178). If `.env` doesn't exist (a not-uncommon teardown precondition — partial init, prior teardown didn't restore it, etc.), `docker compose down` exits 1 with `couldn't find env file:`. The `|| true` at teardown.sh:216 absorbs the exit code, but the compose down never actually ran, so anonymous volumes from a running stack remain attached — exactly the bug this PR is trying to fix.

The pre-PR code at the original teardown.sh:200-205 explicitly checked `if [ -f "\$PROJECT_ROOT/.env" ]` before passing `--env-file`. That guard is gone now.

**Fix**: either (a) make `get_compose_cmd` skip `--env-file` when the file doesn't exist, or (b) in teardown specifically, fall through to the no-env fallback path when `.env` is missing. (a) is the better fix because every caller of `run_compose` has this latent issue, not just teardown.

---

### 6. Minor

- teardown.sh:194 changed the log message from `docker-compose down` to `docker compose down` (hyphen vs space). Cosmetic, but worth knowing it churns the visible output.
- teardown.sh:260 — `^kg_` in the grep matches *Docker volume names* which use underscores (`kg_postgres_data`), good. But `CONTAINER_PREFIX` is configurable in `.operator.conf` (common.sh:31), so a deployment that overrides it bypasses this pass entirely. Probably acceptable since pass 2 catches the rest, but worth a comment.
- teardown.sh:280 — `2>/dev/null || true` on `docker volume ls` is defensive but I can't think of a scenario where that command fails on a working docker daemon. If it does fail, you want to know, not silently get an empty list and skip the dangling sweep. Consider dropping the `|| true`.
- The new code introduces `REMOVE_DANGLING` as a local-ish variable but it's set in branches and read across them (lines 300/304/306, then 312). Initialize it to `false` at the top of the `if [ -n "\$DANGLING_VOLS" ]` block for clarity.

---

### Assessment

The architectural change (use the same compose stack the platform was started with) is correct and overdue — the duplicated `docker-compose ... down` calls in the pre-PR code were already a smell, and the GPU overlay miss was inevitable. Reusing `common.sh` is the right move.

The two things I'd block on:
1. **`-y` blast radius on the dangling sweep** (#4) — real risk of deleting unrelated projects' data on shared hosts/CI.
2. **`--env-file` with missing `.env`** (#5) — silently reintroduces the exact bug the PR fixes.

Everything else is suggestion-grade.

aaronsb:
Addressed in a4d3f8c:

- **Blocking #1 (`-y/--yes` blast radius)**: dangling-volume sweep now classifies each volume as kg-owned or unrelated before removal. kg-owned = compose project label matches `docker`/`kg`/basename, OR anonymous with mountpoint containing recognized schemas (PG18 cluster `18/docker/`, garage `meta/data.db`, HF `hub/`). Unrelated volumes are NEVER auto-deleted under `-y` — interactive confirmation only, even with `--yes` set. Preserves the nuclear-option semantics for our own data without widening blast radius on shared workstations.
- **Blocking #2 (`--env-file` failure when `.env` is gone)**: when `.env` is missing, strip the `--env-file` flag from the composed command rather than fail. Restores the pre-PR guard semantics without giving up the `run_compose` benefit.
- **Suggestion #3 (`set -e` + stale `.operator.conf`)**: source `common.sh` and call `load_operator_config` under `set +e` with explicit return-code checks; fall back to base+dev overlay path on failure and log it. A teardown that aborts on stale config is worse than one that runs the wrong overlay.
- **Suggestion #5 (help text)**: `--help` now calls out that `-y` only auto-removes kg-owned dangling volumes; unrelated ones still prompt.

Skipped:
- `$COMPOSE_CMD` unquoted expansion (issue #4): matches the rest of the codebase (`start-infra.sh`, `stop.sh`, `upgrade.sh`); not a regression. Worth a separate sweep PR if/when the project standardizes on quoted arrays.
- Other minors (log churn `docker-compose` → `docker compose`, `REMOVE_DANGLING` init position, `2>/dev/null || true` on volume listing): kept as-is for diff-scope discipline.

PR #428, opened 2026-05-28
fix(teardown): route operator.sh teardown through lib/teardown.sh

aaronsb:
## Summary

Follow-up to #427. The dangling-volume sweep added there lived in `operator/lib/teardown.sh`, but `./operator.sh teardown --full` — the user-facing entry — went through a separate, simplified `cmd_teardown` in the top-level `operator.sh` that just did `run_compose down -v`. So the fix was real but invisible from the path users actually invoke.

Confirmed empirically: after #427 merged, `./operator.sh teardown --full` still left four orphaned SHA-named anonymous volumes (two PG18 clusters, two web node_modules) — the exact failure mode #427 was meant to close.

## Change

`cmd_teardown` now `exec`s `operator/lib/teardown.sh` and forwards `\"$@\"`. All the lib script's flags become passable from the top-level entry:

| Flag | Effect |
|---|---|
| `--full` | Containers + volumes + images + operator |
| `--remove-volumes` | Volumes only (incl. dangling-volume sweep from #427) |
| `--remove-images` | Force image rebuild on next start |
| `--include-operator` | Also remove the operator container |
| `--keep-env` | Keep `.env` (default deletes secrets) |
| `-y, --yes` | Skip confirmation prompts |

Fallback inline path retained in case `operator/lib/teardown.sh` is missing (partial-install edge case).

Help text in `operator.sh` updated to document the now-passable flags.

## Test plan

- [ ] `./operator.sh teardown --full` on a host with running platform + leftover anonymous volumes: confirm overlay-aware compose down fires AND dangling-volume sweep prompts for kg-owned volumes (per #427's classification)
- [ ] `./operator.sh teardown --keep-env --full`: confirms flag passthrough — `.env` survives
- [ ] `./operator.sh teardown -y --full` on a shared host: confirm unrelated dangling volumes (other projects) are NOT auto-removed
- [ ] `./operator.sh teardown` (no args): basic stop, operator preserved (matches lib script default)
- [ ] `./operator.sh teardown --help`: should now show the lib script's full help (since `exec` replaces the process)

### DIFF
--- a/operator.sh
+++ b/operator.sh
@@ -442,31 +442,30 @@ cmd_upgrade() {
 cmd_teardown() {
     check_env
     load_config
-    cd "$DOCKER_DIR"
 
-    # Teardown is simple enough for host
-    local remove_volumes=false
-    local auto_yes=false
-
-    while [[ $# -gt 0 ]]; do
-        case $1 in
-            --full) remove_volumes=true; shift ;;
-            -y|--yes) auto_yes=true; shift ;;
-            *) shift ;;
-        esac
-    done
-
-    echo -e "${RED}${BOLD}TEARDOWN${NC}"
-    if [ "$auto_yes" = false ]; then
-        read -p "Type 'yes' to confirm: " -r
-        [[ "$REPLY" != "yes" ]] && echo "Cancelled" && exit 0
+    # Delegate to operator/lib/teardown.sh, which has the full
+    # implementation: overlay-aware compose down (honors GPU_MODE etc.),
+    # --remove-orphans, kg-owned vs. unrelated dangling-volume
+    # classification (PR #427), plus --keep-env / --remove-images /
+    # --include-operator flags that this entry point used to silently
+    # ignore.
+    #
+    # Pre-fix history: this function inlined a simplified `run_compose
+    # down -v` that left SHA-named anonymous volumes (e.g. the PG18
+    # cluster at /var/lib/postgresql/18/docker) orphaned after teardown
+    # --full. The lib script's dangling sweep was the actual fix; we
+    # were just bypassing it from the user-facing entry point.
+    local lib_teardown="$SCRIPT_DIR/operator/lib/teardown.sh"
+    if [ ! -x "$lib_teardown" ]; then
+        echo -e "${RED}✗ Cannot find $lib_teardown${NC}"
+        echo -e "${YELLOW}  Falling back to inline teardown${NC}"
+        cd "$DOCKER_DIR"
+        run_compose down -v --remove-orphans
+        echo -e "${GREEN}✓ Teardown complete (fallback path)${NC}"
+        return
     fi
 
-    local flags=""
-    [ "$remove_volumes" = true ] && flags="-v"
-
-    run_compose down $flags
-    echo -e "${GREEN}✓ Teardown complete${NC}"
+    exec "$lib_teardown" "$@"
 }
 
 # ============================================================================
@@ -780,7 +779,13 @@ ${BOLD}Lifecycle:${NC}
   restart <service|all>  Restart a service (or all)
   upgrade            Pull, migrate, restart
   update             Pull images only
-  teardown [--full]  Remove containers (--full: +volumes)
+  teardown [opts]    Remove platform (delegates to operator/lib/teardown.sh)
+                     --full              Containers + volumes + images + operator
+                     --remove-volumes    Volumes only (incl. dangling-volume sweep)
+                     --remove-images     Force image rebuild on next start
+                     --include-operator  Also remove the operator container
+                     --keep-env          Keep .env (default deletes secrets)
+                     -y, --yes           Skip confirmation prompts
 
 ${BOLD}Management:${NC}
   status             Show container status
--- a/operator.sh
+++ b/operator.sh
@@ -440,32 +440,70 @@ cmd_upgrade() {
 }
 
 cmd_teardown() {
-    check_env
-    load_config
-
-    # Delegate to operator/lib/teardown.sh, which has the full
-    # implementation: overlay-aware compose down (honors GPU_MODE etc.),
-    # --remove-orphans, kg-owned vs. unrelated dangling-volume
-    # classification (PR #427), plus --keep-env / --remove-images /
-    # --include-operator flags that this entry point used to silently
-    # ignore.
+    # Delegate to operator/lib/teardown.sh when present (dev / git
+    # checkout install). That script has the full implementation:
+    # overlay-aware compose down (honors GPU_MODE etc.), --remove-
+    # orphans, kg-owned vs. unrelated dangling-volume classification
+    # (PR #427), plus --keep-env / --remove-images / --include-operator
+    # flags that the inline fallback below doesn't support.
     #
     # Pre-fix history: this function inlined a simplified `run_compose
     # down -v` that left SHA-named anonymous volumes (e.g. the PG18
     # cluster at /var/lib/postgresql/18/docker) orphaned after teardown
     # --full. The lib script's dangling sweep was the actual fix; we
     # were just bypassing it from the user-facing entry point.
+    #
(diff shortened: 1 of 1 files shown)


aaronsb:
## Review

Took a careful pass through `cmd_teardown` (operator.sh:442-468), `operator/lib/teardown.sh`, `operator/lib/common.sh`, and the dispatch path. The exec routing is correct and the architectural move is the right call — but there are two real regressions in the fallback path and one in `check_env`. Findings below in order of severity.

### Issue 1 (blocking): fallback path silently destroys data

**Location**: operator.sh:460-465

**Problem**: The fallback path that runs when `operator/lib/teardown.sh` is missing executes `run_compose down -v --remove-orphans` **unconditionally and without confirmation**. Pre-PR, `cmd_teardown` parsed `--full` and only added `-v` when `--full` was set, *and* prompted \"yes\" unless `-y` was present.

**Why this matters specifically**: `install.sh` (lines 1696-1748) **does not install `operator/lib/`** — only `operator.sh` and the compose files land in the standalone install dir. So this fallback is the **default code path for every curl-installer deployment**, not a partial-install corner case. A standalone-install user running `./operator.sh teardown` (no args, expecting just a stop) will silently lose their PostgreSQL data, Garage uploads, and HF cache. The PR description calls this a \"partial-install edge case\" — empirically it's the production path.

**Suggestion**: Either (a) make the fallback honor the same arg shape as the lib script (parse `--full`/`--remove-volumes`/`-y` minimally, prompt on destructive ops), or (b) refuse to proceed in fallback and tell the user to reinstall. (a) is more user-friendly; (b) is safer and a 3-line fix.

```bash
# Suggested (a) — minimal arg parsing in fallback
local remove_volumes=false auto_yes=false
for arg in \"$@\"; do
    case \"\$arg\" in
        --full|--remove-volumes) remove_volumes=true ;;
        -y|--yes) auto_yes=true ;;
    esac
done
if [ \"\$auto_yes\" = false ]; then
    read -p \"Type 'yes' to confirm: \" -r
    [[ \"\$REPLY\" != \"yes\" ]] && echo \"Cancelled\" && exit 0
fi
cd \"\$DOCKER_DIR\"
local flags=\"--remove-orphans\"
[ \"\$remove_volumes\" = true ] && flags=\"\$flags -v\"
run_compose down \$flags
```

### Issue 2 (medium): `check_env` blocks recovery teardown

**Location**: operator.sh:443

**Problem**: `check_env` exits 1 if `.env` is missing. The lib script (operator/lib/teardown.sh:215-260) deliberately handles missing `.env` gracefully — that's one of the things #427 fixed. But because `check_env` runs before `exec`, that careful handling is unreachable from `./operator.sh teardown`. A user who ran a prior teardown without `--keep-env` and now wants to clean up leftover volumes can't, without manually creating a stub `.env`.

This regression predates this PR (pre-PR cmd_teardown also called `check_env`), but this PR is the right moment to fix it since you're rerouting the entry point to a script that explicitly designed around missing-env tolerance.

**Suggestion**: Drop `check_env` from `cmd_teardown` entirely. The lib script doesn't need it. Keep `load_config` only if you want `KG_API_IMAGE_TAG` available before exec (the lib script re-derives it anyway, so this is also droppable).

### Issue 3 (low): `--help` UX

**Location**: operator.sh:782-788 + operator/lib/teardown.sh:64-93

**Problem**: `./operator.sh teardown --help` execs the lib script, which prints `Usage: /path/to/operator/lib/teardown.sh [OPTIONS]` (line 66) using `$0`. A user who typed `./operator.sh teardown --help` and sees the lib script's internal path in the usage line will reasonably wonder if they did something wrong.

You flagged this in the PR and asked which direction feels right — I lean toward fixing the lib script's `$0` rather than intercepting `--help` in `cmd_teardown`. The lib script should print the user-facing invocation, not its own path. One-line fix:

```bash
# In operator/lib/teardown.sh line 66
echo \"Usage: \${KG_INVOKER:-\$0} [OPTIONS]\"
```

And in operator.sh:467, set it before exec:
```bash
KG_INVOKER=\"./operator.sh teardown\" exec \"\$lib_teardown\" \"\$@\"
```

Optional; status quo is defensible if you'd rather not couple them.

### Issue 4 (low): stricter unknown-arg handling is a behavioral change

**Location**: operator/lib/teardown.sh:95-97

**Problem**: Pre-PR cmd_teardown silently ignored unknown args (`*) shift ;;` on the original line 460). The lib script errors and exits 1 on any unknown option. `./operator.sh teardown foo` used to succeed silently and now fails. Stricter is generally better and I'd keep the lib script's behavior, but worth noting in the PR description so users hitting CI breakage know what changed.

### Confirming the things you asked about

- **`load_operator_config` covers `load_config`'s exports** (common.sh:24-48 vs operator.sh:55-79): both derive `KG_API_IMAGE_TAG` via the same `operator/lib/image-tag.sh` helper. `load_operator_config` additionally sets `CONTAINER_PREFIX`, `CONTAINER_SUFFIX`, `COMPOSE_FILE`, `IMAGE_SOURCE` — strictly a superset. The lib script's `get_compose_cmd` (common.sh:116-180) builds the full overlay set from these, including the GHCR / standalone / SSL / GPU layers that operator.sh:82-107 also handles. No drift in overlay selection that would route the exec'd shell to a different compose stack than the parent.
- **No env vars from the parent leak into the lib script's logic**: every var the lib script depends on (`SCRIPT_DIR`, `PROJECT_ROOT`, `DOCKER_DIR`, `CONFIG_FILE`, `ENV_FILE`) is recomputed in common.sh or teardown.sh from `BASH_SOURCE`. `exec` replacing the parent process is safe.
- **Flag overlap with the dispatcher**: none. operator.sh:822 is a pure subcommand dispatch with no global flags. `shift` at line 840 consumes `teardown` and the remaining argv flows cleanly into the lib script.
- **Exit code propagation through `exec`**: correct. `set -e` at teardown.sh:2 plus explicit `exit 0/1` at unambiguous decision points means `./operator.sh teardown --full && echo done` will work.
- **`DOCKER_DIR` layout in the lib script**: teardown.sh:26 hardcodes `PROJECT_ROOT/docker`, while operator.sh:41-44 detects standalone-vs-repo. As noted in Issue 1, this is moot because standalone installs don't ship the lib script — but if someone ever does install the lib script into a standalone layout (e.g. via `self-update` extending coverage, or a custom deploy), the layouts diverge. Worth a comment in teardown.sh acknowledging this, or detecting the same way common.sh:13-19 does.

### Architectural note (acknowledgment)

The core move — collapsing two parallel teardown implementations into one — is the right call and is a model fix for SRP at the script level. The pre-PR drift between cmd_teardown and the lib script was exactly the kind of duplication ADR-101 (image-tag helper) was extracted to prevent in the compose-overlay code. Same pattern, same fix.

---

Bottom line: Issue 1 is a data-loss regression for the curl-installer install mode, which is the standalone-deployment path. The fix is small. Issues 2-4 are improvements, not blockers.

aaronsb:
Addressed in 6ccc3e6:

- **Blocking #1 (data loss in fallback path)**: you're right — `install.sh` doesn't ship `operator/lib/`, so curl-installer users land in the fallback by default, not as edge case. Restored the original `cmd_teardown` behavior in the fallback: parse `--full`/`--remove-volumes`/`-y`/`--yes`, prompt for confirm, only pass `-v` on explicit volume-removal intent. Added `--remove-orphans` (safe-by-default flag-free addition) and a brief `--help` that points users to the git checkout for the full flag set.
- **Medium #2 (`check_env` defeats `.env`-tolerant lib design)**: dropped from `cmd_teardown`. The lib script and the fallback's `run_compose` both handle missing `.env` already.

Skipped:
- Help-text UX (low): proper fix needs a `KG_INVOKER` env-var pattern in `lib/teardown.sh` to swap `$0` in the `Usage:` line. Worth doing but out of scope for this PR's blast radius.
- Unknown-arg strictness change (low): noted, leaving as-is.

Issue #313, opened 2026-02-10
bug: Image ingestion fails with NomicVisionModel AttributeError


aaronsb:
## Bug

Image ingestion via `POST /ingest/image` fails with:

```
Visual embedding generation failed: 'NomicVisionModel' object has no attribute 'all_tied_weights_keys'
```

## Steps to Reproduce

1. Submit any PNG image for ingestion (tested via FUSE `cp image.png ~/Knowledge/ontology/fuse-test/ingest/`)
2. Job submits successfully, gets `pending (analyzing)` status
3. Job fails after ~3 seconds with the above error

## Context

- Job `job_24477ca9c5d2` on local dev instance
- The FUSE driver correctly routes PNG files to `/ingest/image` — the failure is in the API's vision embedding pipeline
- Likely a library version mismatch with the Nomic vision model

PR #429, opened 2026-05-28
fix(gpu): rename dtype kwarg + remove ROCm 6.x wheel path

aaronsb:
## Summary

Two cleanup themes triggered by warnings observed after PR #426:

**[1] Deprecation kwargs**: transformers 4.42+ renamed `torch_dtype` → `dtype` on `AutoModel.from_pretrained`; PyTorch 2.5+ unified the allocator env vars under `PYTORCH_ALLOC_CONF` (supersedes `PYTORCH_HIP_ALLOC_CONF` and `PYTORCH_CUDA_ALLOC_CONF`). Both old names still work but emit warnings on every model load / process start.

**[2] Drop ROCm 6.x wheel variants**: ADR-101 defined `rocm60` / `rocm61` wheel-based variants but they were never published. The init wizard advertised an "AMD GPU (ROCm 6.x — preview)" option that resolved to `kg-api:latest` (the CUDA-bundled wheel), then quietly fell back to CPU on AMD hosts via ADR-043. That's a silent-broken state by every name in the book — the original ADR-101 was supposed to *close* exactly this kind of gap.

| Area | Change |
|---|---|
| `api/Dockerfile` | Removed `PYTORCH_VARIANT` arg + ROCm wheel conditional; file is now CPU/NVIDIA only |
| `docker/docker-compose.gpu-amd.yml` | **Deleted** |
| `docker/docker-compose.gpu-amd-host.yml` | `PYTORCH_HIP_ALLOC_CONF` → `PYTORCH_ALLOC_CONF` |
| `api/app/lib/visual_embeddings.py` | `torch_dtype` → `dtype` |
| `operator/lib/image-tag.sh` | `amd` becomes deprecated alias for `amd-host` (warns to stderr) |
| `operator/lib/common.sh`, `operator.sh`, `install.sh` | Compose-cmd builders route both `amd` and `amd-host` to the host-ROCm overlay |
| `operator/lib/guided-init.sh` | Wizard options renumbered: `[3]` AMD GPU (single), `[4]` CPU |
| `operator/lib/start-app.sh` | Prints deprecation notice when `GPU_MODE=amd` |
| `scripts/publish.sh` | Removed `rocm60`/`rocm61` from publish targets, help text, case |
| `docs/architecture/infrastructure/ADR-101-*.md` | Appended "Update — 2026-05-28" section documenting the removal + rationale |
| `docker/docker-compose.ghcr.yml` | Updated comment to reflect new tag scheme |

### Backward compatibility

`GPU_MODE=amd` in existing `.operator.conf` files keeps working — treated as a deprecated alias for `amd-host`. Users get one line of stderr from the image-tag helper and a yellow warning at platform start. Re-running `./operator.sh init` (and picking option [3]) writes the new mode.

## Verification

Ran the rebuild on my AMD-host box after the change:

```
torch=2.9.1+rocm7.2.3
hip=7.2.53211
cuda_avail=True
device_count=2
device_0_name=AMD Radeon RX 7900 XTX
```

Neither the `torch_dtype is deprecated` nor the `PYTORCH_HIP_ALLOC_CONF is deprecated` warnings fire anymore.

## Test plan

- [ ] Fresh `./operator.sh init` on AMD host: wizard shows 4 options, picking [3] writes `GPU_MODE=amd-host`
- [ ] Existing `.operator.conf` with `GPU_MODE=amd`: platform starts, deprecation warning visible in `./operator.sh start` output and (separately) in image-tag stderr during compose-cmd assembly, otherwise behaves identically to `amd-host`
- [ ] `./operator.sh start` on CPU-only host: builds from `api/Dockerfile` without `PYTORCH_VARIANT`, no ROCm wheel index hit
- [ ] `./operator.sh start` on AMD host: rebuilt image is `torch=*+rocm*`, no deprecation warnings in startup log
- [ ] `./publish.sh images-rocm` builds and pushes `rocm72-host` only (no rocm60/rocm61 attempt)
- [ ] curl-installer (`install.sh`) with `GPU_MODE=amd-host` downloads the gpu-amd-host overlay and routes to it

## Not included

GPU enumeration + post-start device selection (`./operator.sh gpu list/select`) — separate design conversation, queued.


aaronsb:
## Review

Focused walkthrough against the seven correctness/risk questions in the review brief. Branch reviewed at `c45c2bfb`.

---

### 1. Compose-cmd routing (no orphan `gpu-amd.yml` refs)

Traced all three builders. Verdict: **no live code references the deleted file**, but two **docs/config files still mention it** and should be cleaned up in this PR or a fast-follow:

- `docs/operating/configuration.md:141` — table row says ``| `GPU_MODE=amd` | + `docker-compose.gpu-amd.yml` |``. That's now wrong on two counts: the file doesn't exist, and `amd` resolves to `docker-compose.gpu-amd-host.yml`. The row below it (`GPU_MODE=amd-host`) is accurate and the `amd` row should be either removed or annotated as "deprecated alias for amd-host".
- `.env.example:268-271` — the `ROCM_VERSION` block still says "ROCm wheel version for docker-compose.gpu-amd.yml" and "Only used with GPU_MODE=amd (ROCm wheels, not host ROCm)". With the wheel path gone, this whole block is dead config that will confuse anyone reading the env template.

Compose-cmd builders themselves are clean:
- `operator.sh:101` — `amd|amd-host) → docker-compose.gpu-amd-host.yml` ✓
- `operator/lib/common.sh:161-167` — same routing ✓
- `install.sh:2304-2306` — same routing ✓

### 2. `derive_kg_api_image_tag` callsites + dead config

All three callers still pass the legacy 2nd arg (now ignored) — works correctly because the helper just drops it:
- `operator.sh:76` — `derive_kg_api_image_tag "$GPU_MODE" "$ROCM_VERSION"` ✓
- `operator/lib/common.sh:45` — same pattern ✓
- `operator/lib/guided-init.sh:307` — same pattern ✓

**Dead config to clean up**:
- `docs/operating/configuration.md:98` — ``| `ROCM_VERSION` | ROCm wheel version (`rocm60`, `rocm61`) |`` is documenting a variable that no longer drives behavior.
- `.env.example:268-271` (already noted in #1) — same dead `ROCM_VERSION` block.

The ROCM_VERSION variable is read from environment but never written to `.operator.conf` by guided-init. So the deprecation path is graceful (env-only); no operator.conf rewrite needed. Recommend either: (a) silently ignore in this PR with the docs fixes, or (b) add a one-line stderr warning if `ROCM_VERSION` is set, mirroring the `GPU_MODE=amd` warning pattern.

### 3. Backward compat trace for `GPU_MODE=amd`

Walked the full path. All branches handle the alias correctly:
- `image-tag.sh:51-56` — stderr warning fires, returns `rocm72-host` ✓
- `start-app.sh:150-153` — yellow alias notice prints ✓
- `guided-init.sh:324` — multi-GPU note correctly covers both modes ✓
- `guided-init.sh:367-372` — embedding device map: `amd|amd-host → cuda` ✓ (PyTorch ROCm presenting as CUDA is correct for both)

One nit: the deprecation warning in `image-tag.sh` fires from `derive_kg_api_image_tag`, which is called *during config load* (`operator.sh:73-77`, `common.sh:42-46`). That means the warning fires once on **every** invocation of any operator subcommand (`status`, `logs`, `shell`, etc.), not just `start`. Probably fine — it's stderr, and the goal is to nag — but if you want quieter behavior, consider gating it on `[ "$KG_API_IMAGE_TAG" = "" ]` so it only fires when the operator is computing the tag fresh, not when `.operator.conf` already has it pinned.

### 4. transformers `dtype` kwarg floor — **THIS IS WRONG**

This is the only blocking issue I found.

The PR description says "Floor is 4.41 per requirements.txt — kwarg `dtype` is supported there too" and the inline comment in `visual_embeddings.py:102-105` claims transformers 4.42+ renamed the kwarg. Both are incorrect.

I verified against the actual transformers source on GitHub. The `dtype` kwarg was added in **transformers 4.56.0**, not 4.42:

```
v4.45.0: dtype = kwargs.pop("dtype" ...) count=0
v4.50.0: count=0
v4.55.0: count=0
v4.56.0: count=2   ← introduced here
v4.57.0: count=2
```

`requirements.txt:7` pins `transformers>=4.41.0,<5.0`. A fresh install with pip resolving the floor gets a version where `dtype` is silently passed through `**kwargs` and **never reaches the dtype-resolution code path** — meaning `torch.float16` will not be applied, and the model loads in fp32 on GPU. That's a real correctness regression for anyone on the floor, doubling VRAM usage on accelerator paths.

The signature inspection in your PR description shows `params: ['model_args', 'kwargs']` — that's because `from_pretrained` always accepted `**kwargs`. The signature inspection doesn't tell you whether the kwarg is *consumed*; only reading the source does. The verification was done against the currently-installed 4.57.6 only.

**Fix options**:
- (a) Bump the floor: `transformers>=4.56.0,<5.0`. Lowest-effort. Need to confirm Nomic Vision's `trust_remote_code` still works on 4.56+; the existing pin comment cites #313 as the upper-bound rationale, not a lower-bound issue.
- (b) Conditional kwarg: detect transformers version and pick `dtype` vs `torch_dtype` at runtime. More resilient but uglier.
- (c) Keep `torch_dtype` and just suppress the warning. Counterproductive given the goal of this PR is removing the warning.

I'd lean (a) — the floor `>=4.41.0` is already 18 months old; the only reason for that floor is the Nomic compat note, and that constraint is about the upper bound, not the lower.

### 5. Orphaned `PYTORCH_VARIANT` references

Checked `scripts/publish.sh:818-821` — `rocm72-host` case explicitly sets `build_arg=""` and the subsequent `[ -n "$build_arg" ] && build_arg_flag=(--build-arg "$build_arg")` (`publish.sh:858-859`) means no `--build-arg PYTORCH_VARIANT=...` is ever passed. Clean.

Repo-wide grep for `PYTORCH_VARIANT` returns only the ADR-101 historical references (lines 26, 45, 81-83, 86, 147) and the new addendum mentioning it as removed. No live code paths.

### 6. Image tag consistency between `IMAGE_SOURCE=local` and `ghcr`

Confirmed both paths route consistently for `GPU_MODE=amd`:
- GHCR: `KG_API_IMAGE_TAG=rocm72-host` → `kg-api:rocm72-host` (correct image)
- Local build: compose merges base `docker-compose.yml` (`image: kg-api:latest`) with `docker-compose.gpu-amd-host.yml` (`build: dockerfile: api/Dockerfile.rocm-host`). Docker compose's behavior here is: `build:` is used to construct the image, and the resulting image is tagged with the base `image:` value (`kg-api:latest`). So the local-build `amd-host` image is *tagged* `kg-api:latest` but *built* from the ROCm host Dockerfile. This is consistent with PR #427/#428 history as you noted.

No edge case. The warning we emit ("treating as amd-host") is honest because the variant selection — whether we end up with rocm/pytorch base or PyPI torch wheel — is fully determined by which compose overlay loads, and both `amd` and `amd-host` load the same overlay.

### 7. ADR amendment

The addendum at `docs/architecture/infrastructure/ADR-101-...md:253-290` is well-written and accurate. Specifically:

- The summary of what was deleted matches the diff exactly ✓
- The "Backward compatibility" bullet correctly enumerates the three back-compat surfaces (image-tag stderr, compose-cmd routing, start-app warning) ✓
- The rationale section (lines 284-290) is candid about the original ADR's mistaken premise — that's the right tone for an addendum, not a supersession ✓

**My take**: inline addendum is the right call here, not a superseding ADR. The core architectural decision (install-time variant selection, kg-api:<variant> tag scheme, multi-arch publish flow) is unchanged. What's removed is one branch of the variant tree (wheel-based ROCm). A new ADR would be the right vehicle if the variant *mechanism* were changing, but it isn't — only the membership of the variant set is.

One minor suggestion: the addendum at line 268-270 says option `[4]` becomes `[3]`. The current diff shows `[3]` is now "Linux with AMD GPU (ROCm)" — consider noting that the option text changed too, not just the number, so a future reader cross-referencing wizard screenshots can follow the rename. Trivial.

### 8. Other findings

- **`install.sh:1502-1509` validation gap**: `case "$GPU_MODE" in auto|nvidia|amd|cpu)` rejects `amd-host` as an invalid value. A user passing `--gpu amd-host` to the curl installer (or pre-setting `GPU_MODE=amd-host` in env) gets `add_validation_error "Invalid GPU mode: $GPU_MODE (must be: auto, nvidia, amd, cpu)"` and the install aborts. Meanwhile the compose-cmd builder at `install.sh:2304` explicitly accepts `amd-host`. Inconsistent — the validator should add `amd-host` to the allowed set.

- **`install.sh:1324-1328` interactive prompt** still offers "amd (AMD GPU)" as the user-facing choice with no UI hint that this is now the deprecated alias. After install, the user gets the stderr deprecation warning from image-tag.sh on every operator command. Suggest either (a) update the prompt text to "amd-host (AMD GPU, ROCm)" and write `GPU_MODE=amd-host` to the install, or (b) keep `amd` for the v1 standalone install but acknowledge the gap.

- **`PYTORCH_ALLOC_CONF` requires PyTorch 2.5+**: the `Dockerfile.rocm-host` base image ships PyTorch 2.9.1, so the published `rocm72-host` image is fine. But a developer doing a local build of the `amd-host` path on a host with older PyTorch would get an unrecognized env var. Practically a non-issue because amd-host requires the rocm/pytorch base which is always 2.5+. Worth a sentence in the inline comment in `docker-compose.gpu-amd-host.yml:48-51` clarifying the requirement is satisfied by the base image.

---

### Assessment

The cleanup is correct in shape and the architectural reasoning (especially the ADR addendum) is honest about the prior gap. **The blocker** is the transformers floor mismatch in #4 — that needs a fix (or a floor bump) before merge, because installs on the pinned floor will silently lose fp16 on GPU paths.

Secondary items (docs/.env.example dead refs, install.sh validation gap, install.sh interactive prompt text) are mergeable-with-fast-follow but should not block.

---
*AI-assisted review via Claude (Opus 4.7)*

aaronsb:
Addressed in 5c008d2:

- **Blocker (transformers floor)**: bumped `requirements.txt:7` to `>=4.56.0,<5.0`. You're right — I verified against the installed 4.57.6 and missed that the floor was still 4.41. With the old floor, the `dtype` kwarg would have silently no-op'd through `**kwargs` and loaded fp32 on GPU, defeating the entire VRAM-reduction goal of #426 + this PR. Lower-bound comment cites the dtype kwarg specifically; upper-bound retains the Nomic Vision / transformers 5.x note.
- **Stale doc references (configuration.md + .env.example)**: collapsed the `amd` / `amd-host` rows in the compose-files table, removed the `ROCM_VERSION` entry from the AMD GPU env table, replaced the `.env.example` block with a removal-pointer to the ADR-101 update.
- **install.sh inconsistencies**: validator at `:1502-1509` now accepts `amd-host` (kept `amd` as alias); interactive installer at `:1324-1328` now offers `amd-host (AMD GPU via host ROCm)` to match operator.sh's wizard option [3].

Skipped (with rationale in commit message):
- Warning-gate suggestion on image-tag.sh: the warning already gates on `[ -z "$KG_API_IMAGE_TAG" ]` at both callsites (`operator.sh:73`, `common.sh:42`), so it only fires for truly stale `.operator.conf` — which is the right place to nudge re-init.

Issue #442, opened 2026-05-28
Tracking: internet-hardening — endpoint auth/RBAC remediation before public exposure


aaronsb:
Tracking issue for the **internet-hardening** cluster: closing the auth/authorization gaps found in the 2026-05-28 endpoint security audit before the platform is exposed on the open internet.

**Common context (read first):** `docs/security/endpoint-security-audit-2026-05-28.md` (PR #430)
**Operative auth model (authoritative):** ADR-400 — Operative RBAC and Endpoint Security Baseline

## Verdict: NOT READY for exposure
Two critical (verified), three high blockers, plus a systemic authenticated-but-unauthorized pattern. The audit reconstructed the model from the SQL migrations + enforcement code (not the drifted ADRs) and measured every endpoint against the seeded grants.

## Phase 0 — Blockers: stop the bleeding
- [x] #431 — 🔴 `/auth/register` accepts arbitrary role (platform_admin escalation)
- [x] #432 — 🔴 `/admin/models/catalog` router fully unauthenticated

## Phase 1 — Systemic authZ gaps (high)
- [x] #433 — database router authenticated-only (arbitrary Cypher read exfiltration)
- [x] #434 — enforce `ingest:create` (denial-of-wallet)
- [x] #435 — graph query surface lacks `graph:read` + ungated compute jobs
- [x] #436 — raw content exfiltration via `documents/*` and `sources/*`
- [x] #437 — anonymous `GET /vocabulary/category-flows` + unauthenticated image health

## Phase 2 — Consistency & hardening (medium)
- [x] #438 — projection/ontology/vocabulary read gates
- [x] #439 — `get_current_user`→`get_current_active_user` + NULL-owner artifact bypass
- [x] #440 — CLI coherence (dead paths, RBAC role guidance, pre-flight auth)

## Phase 3 — Low-severity
- [x] #441 — `admin_workers` cancel vs `jobs:cancel` scoping; `oauth_clients:write` admin gap

## Phase 4 — ADR re-baseline
- [x] ADR-400 authored + accepted; ADR-017/027/028/060/074 superseded (PR #430)
- [ ] Migration to clean orphaned/vestigial grants (`jobs:write`/`approve`, old `roles`/`resources` resources)

## Related existing issues
- #419 — web jobs-view 401 / inconsistent session invalidation (add global 401 interceptor + token-lifetime criteria here)
- #417 — per-provider concurrency semaphore (partial mitigation for #434)
[...]

Issue #437, opened 2026-05-28
SECURITY (high): close anonymous GET /vocabulary/category-flows and unauthenticated GET /ingest/image/health


aaronsb:
## Summary
Two endpoints are reachable with zero credentials and are not on the public-by-design list:
- GET /vocabulary/category-flows (vocabulary.py:1380) — empty parameter list, runs two full-graph Concept scans and returns total_concepts/total_edges/category flow matrix to anyone.
- GET /ingest/image/health (ingest_image.py:311-344) — empty parameter list, returns GPU/CPU, VRAM stats, model info, vision providers.

## Acceptance Criteria
- [ ] /vocabulary/category-flows gains `current_user: CurrentUser` + Depends(require_permission('vocabulary','read')).
- [ ] /ingest/image/health gains at minimum `current_user: CurrentUser`, preferably Depends(require_permission('admin','status')).
- [ ] A test enumerates all routers and asserts no handler (except the documented public-by-design set) is reachable without authentication.

---
*Part of the **internet-hardening** cluster. Common context: `docs/security/endpoint-security-audit-2026-05-28.md` (PR #430). Operative auth model: ADR-400.*

aaronsb:
Fixed in PR #444 (merged). Gated against the seeded-grant model (ADR-400); regression tests assert anonymous rejection + real authZ denial.

Issue #433, opened 2026-05-28
SECURITY (high): database router is authenticated-only — gate raw Cypher and metrics endpoints


aaronsb:
## Summary
Every handler in api/app/routes/database.py depends only on CurrentUser and calls no require_permission, despite docstrings claiming database:read/execute/write. Arbitrary read Cypher (POST /database/query) is reachable by any authenticated user incl. self-registered read_only — full-graph read exfiltration (check_cypher_safety blocks writes but not reads). refresh_graph_counters mutates+commits (docstring falsely claims unseeded 'database:write'). info/stats/counters leak Postgres version, host:port, db user, graph scale.

Note: a duplicate arbitrary-Cypher executor exists at POST /query/cypher (queries.py) — fix both.

## Acceptance Criteria
- [ ] POST /database/query and POST /database/counters/refresh gated on require_permission('database','execute') (platform_admin). Do NOT use 'database','write' (unseeded -> permanent 403).
- [ ] GET /database/stats, /database/info, /database/counters gated on require_permission('database','read') (admin/platform_admin).
- [ ] POST /query/cypher gated on require_permission('graph','execute') (platform_admin).
- [ ] /database/epoch either gated on database:read or explicitly documented as authenticated-by-design (it is on the main.py:149 polling allowlist).
- [ ] Docstrings corrected to match enforced permissions.

---
*Part of the **internet-hardening** cluster. Common context: `docs/security/endpoint-security-audit-2026-05-28.md` (PR #430). Operative auth model: ADR-400.*

aaronsb:
Fixed in PR #444 (merged). Gated against the seeded-grant model (ADR-400); regression tests assert anonymous rejection + real authZ denial.

Issue #432, opened 2026-05-28
SECURITY (critical): entire /admin/models/catalog router is unauthenticated — add auth + extraction_config gating


aaronsb:
## Summary
api/app/routes/models.py has no auth on any handler (Depends is not even imported; router created with no dependencies=; main.py:533 adds none). All six endpoints are reachable anonymously:
- GET /admin/models/catalog — leaks provider catalog + pricing
- POST /admin/models/catalog/refresh — outbound provider API amplification + DB upsert
- PUT .../{id}/enable, /disable, /default — mutate model availability (extraction/embedding DoS)
- PUT .../{id}/price — overwrite per-token pricing feeding ingest cost-estimation/approval

## Acceptance Criteria
- [ ] Import Depends + require_permission in models.py.
- [ ] GET catalog gated on require_permission('extraction_config','read') (admin/platform_admin).
- [ ] All mutators (refresh/enable/disable/default/price) gated on require_permission('extraction_config','write') (platform_admin), OR a dedicated models:read/models:write resource is seeded via migration and used.
- [ ] Regression test asserts anonymous calls to every /admin/models/* route return 401/403.

---
*Part of the **internet-hardening** cluster. Common context: `docs/security/endpoint-security-audit-2026-05-28.md` (PR #430). Operative auth model: ADR-400.*

aaronsb:
Fixed in PR #443 (merged): the /admin/models/catalog router now gates reads on extraction_config:read and all mutations on extraction_config:write. Regression tests assert anonymous rejection + admin-denied writes.

PR #443, opened 2026-05-28
fix(auth): internet-hardening Phase 0 blockers — register escalation + models router (#431, #432)

aaronsb:
Closes the two **critical** blockers from the 2026-05-28 endpoint security audit (`docs/security/endpoint-security-audit-2026-05-28.md`), measured against the operative model in **ADR-400**. Tracking: #442.

## #431 — `/auth/register` privilege escalation *(critical)*
`POST /auth/register` honored a client-supplied `role` (the validator even permitted `platform_admin`), so any anonymous caller could mint a `platform_admin` and own the platform via role inheritance.
- New `RegisterRequest` model with **no role field**; the handler forces `primary_role='read_only'`. Elevated roles only via the `users:create`-gated admin path (`POST /users`), consistent with the operator-seeded root admin being the account-creation root.
- Migration **071** seeds a `registration_enabled` platform-config flag (default `true`; operators set `false` for internet deployments → endpoint returns 403). Missing flag = enabled (back-compatible). Flag checked before bcrypt.

## #432 — `/admin/models/catalog` router fully unauthenticated *(critical)*
The whole router had no auth dependency — anonymous catalog mutation, pricing corruption, outbound API amplification.
- Reads require `extraction_config:read` (admin + platform_admin); all mutations require `extraction_config:write` (platform_admin only). Both pairs seeded in migration 028 — no dead-check.

## Tests
- `tests/api/test_auth.py`: self-registration yields `read_only`; attempted `platform_admin`/`admin`/`curator` is ignored, not honored.
- `tests/api/test_models_catalog_auth.py`: anonymous rejected on every endpoint; contributor denied read; admin passes read but denied mutations (real RBAC vs seeded grants).
- Full API suite: **434 passed, 15 skipped**.

## Deferred decision (not in this PR)
Auto-setting `registration_enabled=false` in the operator **prod** init path. No operator→platform_config write pattern exists today, and it needs live testing — flagged for your call. Operators can set it now via SQL/configure.py.

Remaining blockers #433–#437 (high) and Phase 2–3 follow in subsequent PRs per the audit's phased plan.

### DIFF
--- a/api/app/models/auth.py
+++ b/api/app/models/auth.py
@@ -43,6 +43,29 @@ class UserCreate(UserBase):
     }
 
 
+class RegisterRequest(BaseModel):
+    """
+    Schema for open self-registration via POST /auth/register.
+
+    Deliberately has NO role field: self-registered accounts are always created
+    with the least-privilege 'read_only' role server-side (ADR-400,
+    internet-hardening #431). Elevated roles are assigned only through the
+    users:create-gated admin path (POST /users -> UserCreate), never by the
+    self-registering client.
+    """
+    username: str = Field(..., min_length=3, max_length=100, description="Unique username")
+    password: str = Field(..., min_length=8, description="Password (will be hashed)")
+
+    model_config = {
+        "json_schema_extra": {
+            "examples": [{
+                "username": "alice",
+                "password": "SecurePass123!"
+            }]
+        }
+    }
+
+
 class UserRead(UserBase):
     """Schema for reading user data (excludes password)"""
     id: int = Field(..., description="User ID")
--- a/api/app/routes/auth.py
+++ b/api/app/routes/auth.py
@@ -119,7 +119,9 @@ async def register_user(user: RegisterRequest):
             # that have not applied migration 071 keep working.
             cur.execute("SELECT kg_api.get_platform_config('registration_enabled')")
             flag_row = cur.fetchone()
-            if flag_row and str(flag_row[0]).strip().lower() == "false":
+            # Treat any common falsy spelling as disabled (operators may set
+            # false/0/no/off). Missing / anything else = enabled (default).
+            if flag_row and str(flag_row[0]).strip().lower() in ("false", "0", "no", "off"):
                 raise HTTPException(
                     status_code=status.HTTP_403_FORBIDDEN,
                     detail="Open self-registration is disabled. Contact an administrator to create an account."
--- a/api/app/routes/auth.py
+++ b/api/app/routes/auth.py
@@ -47,6 +47,7 @@ from api.app.lib.auth import (
 )
 from api.app.models.auth import (
     UserCreate,
+    RegisterRequest,
     UserRead,
     UserUpdate,
     UserInDB,
@@ -78,9 +79,20 @@ admin_router = APIRouter(prefix="/users", tags=["admin"])
 # =============================================================================
 
 @router.post("/register", response_model=UserRead, status_code=status.HTTP_201_CREATED)
-async def register_user(user: UserCreate):
+async def register_user(user: RegisterRequest):
     """
-    Register a new user account.
+    Open self-registration for a new user account.
+
+    Self-registered accounts are ALWAYS created with the least-privilege
+    'read_only' role (ADR-400, internet-hardening #431); the client cannot
+    choose a role. Elevated roles are assigned only through the
+    users:create-gated admin path (POST /users), which — together with the
+    operator-seeded root admin and, for bespoke deployments, custom SQL
+    migrations — is the canonical way to provision accounts.
+
+    Open registration can be disabled per-deployment via the
+    'registration_enabled' platform-config flag (operators set it false for
+    internet-facing deployments). When disabled this endpoint returns 403.
 
     Password requirements:
     - Minimum 8 characters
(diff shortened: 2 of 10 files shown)


aaronsb:
## Code Review — internet-hardening Phase 0 (#431, #432)

Reviewed against ADR-400 and `endpoint-security-audit-2026-05-28.md` (F1/F2). **The two server-side security fixes are correct, fail-closed where it matters, and backed by non-vacuous tests.** There is one real coherence regression in the CLI that should land with (or immediately after) this PR, plus a few low/nit items.

### Assessment of the core fixes

**#431 register escalation — closed.**
- `RegisterRequest` (`api/app/models/auth.py:46`) is a plain `BaseModel` → Pydantic v2 default `extra="ignore"`, so a client-supplied `role` is silently dropped (not honored). The handler additionally hardcodes `SELF_REGISTRATION_ROLE = "read_only"` and the INSERT binds that literal (`api/app/routes/auth.py:139,149`). Two independent controls; escalation is closed even if one regresses.
- No other unauthenticated path writes `primary_role` from client input. The only other role-bearing insert is the gated `POST /users` (`auth.py:461`, `require_permission("users","create")`).

**#432 catalog router — closed.**
- All 6 `@admin_router` endpoints now carry `Depends(require_permission("extraction_config", read|write))`. `require_permission` → `get_current_active_user` enforces authenticated + non-disabled *before* the permission query, so the anonymous hole is shut and disabled accounts are rejected. Single mount point (`main.py:533`), OAuth-only (`get_current_user` → `validate_oauth_access_token`), no API-key bypass.
- The seeded pairs are live, not dead-checks: migration 028 grants `extraction_config:read` to `admin` + `platform_admin` (028:105,144) and `extraction_config:write` to `platform_admin` only (028:145).

**Tests prove the invariant (not vacuous).**
- `test_register_user_cannot_self_assign_privileged_role` sends `role` in the body for platform_admin/admin/curator/superadmin and asserts the result is `read_only`. Exercises the exact escalation vector.
- `test_models_catalog_auth.py` runs **real RBAC** (no `bypass_permission_check`). `mock_oauth_validation` returns a `UserInDB` whose `id` is parsed from the token; `PermissionChecker.can_user` then resolves the role from the DB row seeded by `ensure_test_users_in_db` (user 101 = `admin`). So `test_catalog_mutation_denied_for_admin` genuinely proves write is platform_admin-only, and the read allow/deny split is real.

### Findings

**HIGH — CLI `createUser` still posts to `/auth/register`; this PR silently breaks `kg admin user create`.**
`cli/src/lib/auth/auth-client.ts:411-419` (and the comment at :400-404 "Admins can create users with any role") posts `{username, password, role}` + Bearer token to `/auth/register`. After this PR:
- the `role` is ignored (extra-field) → every admin-created user becomes `read_only` regardless of `--role`;
- on the prod path *this PR itself* adds (`registration_enabled=false`), the call returns **403**, so `kg admin user create` is fully broken in production;
- the Bearer token was never checked by `/auth/register` anyway, so the "admin only" guarantee was always illusory (audit F15).
**Fix:** repoint `createUser` at the gated `POST /users` (admin_router, `UserCreate`, `require_permission("users","create")`) — exactly what the audit's F15 prescribes. Without this, the PR makes a documented admin workflow silently misbehave. This is arguably in-scope for "Phase 0 blockers" since the prod-hardening flag is what triggers the 403.

**LOW — `platform-config set` accepts arbitrary keys with no validation (silent typo → fail-open).**
`operator/configure.py` `cmd_platform_config` passes any `key` straight to `set_platform_config`, which upserts unconditionally. An operator typo (`registratoin_enabled false`) succeeds with a green checkmark while leaving registration enabled. The init scripts use a hardcoded literal so the automated path is safe, but a hand-run hardening command can fail open silently. Consider validating against a known-keys set (or at least warning on unknown keys).

**LOW — only the exact string `false` disables registration.**
`auth.py:120` treats anything other than case-insensitive `"false"` as enabled (deliberate, documented). But it means `platform-config set registration_enabled 0` / `no` / `disabled` will *not* disable registration, again failing open. Given this is the one knob standing between an internet deployment and open registration, consider also treating `0`/`no`/`off` as disabled, or normalizing the value on write.

**Note (not a defect) — fail-open on missing flag is the right call here.** `get_platform_config` returns NULL when migration 071 hasn't run → `str(None).lower() != "false"` → enabled. This preserves existing dev/standalone behavior and the prod init path explicitly sets `false`, so the fail-open default never reaches an internet deployment that went through `headless-init`/`guided-init` random-password mode. Reasonable.

**NIT — bcrypt ordering comment vs. duplicate-username timing.** The handler now does flag-check → duplicate-check → bcrypt, which correctly avoids bcrypt on the disabled/duplicate paths. Good. One subtlety: the username-existence check still happens before hashing, so registration remains a (cheap) username-enumeration oracle via 409 vs 201 — pre-existing, unchanged, just flagging for the threat model.

**NIT — removed `test_register_user_invalid_role` (was asserting 422).** Correctly removed since invalid roles are now ignored rather than rejected. No action; noting the intentional behavior change (invalid `role` in a register body is now a silent no-op, not a 422).

### Migration 071
Correct and idempotent: `INSERT ... ON CONFLICT (key) DO NOTHING` for the flag, `ON CONFLICT (version) DO NOTHING` for the migration row, wrapped in BEGIN/COMMIT. Version 71 is sequential (70 is the prior max), no gap. Uses the 031 helpers as documented.

### Bottom line
Server-side: ship it — the escalation and anonymous-access holes are genuinely closed and the tests would catch a regression. Please land the CLI `createUser` repoint (HIGH) alongside this so the admin user-creation path keeps working under the new prod flag, and consider the two fail-open LOWs on the operator knob.

Issue #441, opened 2026-05-28
RBAC consistency: reconcile admin_workers cancel vs jobs:cancel system scoping, and oauth_clients:write admin gap


aaronsb:
## Summary
Two authorization inconsistencies in the seeded grant model:
1. POST /admin/workers/jobs/{id}/cancel (admin_workers.py:189-231) checks only workers:manage and cancels any running job with no is_system check, while DELETE /jobs/{id} routes through JobPermissionChecker where cancelling a system job requires jobs:cancel {is_system:true} (platform_admin only). An admin can interrupt system-lane housekeeping through the workers path it is denied on the jobs path.
2. Migration 028 grants admin oauth_clients create/delete but NOT write, so PATCH /oauth/clients/{id}, rotate-secret, and DELETE /tokens/{hash} (which check oauth_clients:write) are platform_admin-only by accident — admin is locked out of mutating clients it can create/delete.

## Acceptance Criteria
- [ ] Decide and document whether workers:manage intentionally supersedes jobs:cancel system scoping (ADR-100); if not, cross-check is_system via JobPermissionChecker so system-lane cancel requires platform_admin.
- [ ] Decide whether admin should manage OAuth clients; if yes, seed oauth_clients:write for admin via migration; if no, document the platform_admin-only intent.

---
*Part of the **internet-hardening** cluster. Common context: `docs/security/endpoint-security-audit-2026-05-28.md` (PR #430). Operative auth model: ADR-400.*

aaronsb:
Done (commit 88cf0e06 on main): migration 072 seeds admin oauth_clients:write (resolving the accidental lockout from 028); the worker-plane cancel now routes through JobPermissionChecker.can_access_job, aligning it with DELETE /jobs. Note: the audit's F18 premise was corrected — admin can cancel system jobs on both paths by design via the seeded global jobs:cancel grant (041); that is the operative model per ADR-400, not a gap.

Issue #438, opened 2026-05-28
SECURITY (medium): projection + ontology + vocabulary read endpoints lack their documented permission gates


aaronsb:
## Summary
A cluster of read endpoints take only CurrentUser, with docstrings claiming a permission that is never enforced. Sibling mutators in the same files DO gate correctly, confirming drift.

**projection.py:** get_projection (graph:read), get_available_algorithms (graph:read), invalidate_projection (graph:delete), regenerate_projection (graph:create/write — this one also enqueues auto-approved jobs and is high severity).
**ontology.py:** list_ontologies, list_proposals, get_proposal, get_ontology_info, get_ontology_node, get_ontology_files, get_ontology_scores, get_ontology_candidates, get_ontology_affinity, get_ontology_edges (all ontologies:read).
**vocabulary.py:** status, types, category-scores/{type}, similar/{type}, analyze/{type}, epistemic-status, epistemic-status/{type} (all vocabulary:read).

## Acceptance Criteria
- [ ] require_permission('graph','read') on get_projection + get_available_algorithms; require_permission('graph','delete') on invalidate_projection; require_permission('graph','create') on regenerate_projection (plus clamp work and gate job creation).
- [ ] require_permission('ontologies','read') on all ten ontology read endpoints.
- [ ] require_permission('vocabulary','read') on all seven vocabulary read endpoints.
- [ ] Docstrings reconciled with enforcement; test asserts read_only is denied where the grant excludes it.

---
*Part of the **internet-hardening** cluster. Common context: `docs/security/endpoint-security-audit-2026-05-28.md` (PR #430). Operative auth model: ADR-400.*

aaronsb:
Fixed in PR #445 (merged): projection/ontology/vocabulary read endpoints gated (graph:read / ontologies:read / vocabulary:read; projection regenerate→graph:create, invalidate→graph:delete). Regression tests assert the read_only grant split.

Issue #436, opened 2026-05-28
SECURITY (high): raw content exfiltration via documents/* and sources/* — enforce sources:read / graph:read


aaronsb:
## Summary
api/app/routes/documents.py and api/app/routes/sources.py handlers depend only on get_current_active_user with no permission/ownership check, contradicting docstrings. Most sensitive: GET /documents/{id}/content (full original text + base64 image bytes), GET /sources/{id}/document (original pre-chunking document bytes from Garage), GET /sources/{id}/image (raw image binary), GET /sources/{id} and GET /sources (full chunk text + internal storage_key/garage_key). The sources router does not even import require_permission.

## Acceptance Criteria
- [ ] require_permission('sources','read') added to /documents/{id}/content, /sources, /sources/{id}, /sources/{id}/document, /sources/{id}/image.
- [ ] require_permission('graph','read') added to /documents, /documents/{id}/concepts, /documents/concepts/bulk, /query/documents/search, /query/documents/by-concepts.
- [ ] Internal storage_key/garage_key fields restricted to admin/platform_admin responses.
- [ ] Follow-up ticket opened for ADR-028 ontology/ownership scoping so non-owners cannot fetch arbitrary documents even with sources:read.
- [ ] Test asserts read_only is denied and the docstrings match enforcement.

---
*Part of the **internet-hardening** cluster. Common context: `docs/security/endpoint-security-audit-2026-05-28.md` (PR #430). Operative auth model: ADR-400.*

aaronsb:
Fixed in PR #444 (merged). Gated against the seeded-grant model (ADR-400); regression tests assert anonymous rejection + real authZ denial.

Issue #435, opened 2026-05-28
SECURITY (high): graph query surface (queries.py) lacks graph:read and gates no compute jobs


aaronsb:
## Summary
Every handler in api/app/routes/queries.py takes only CurrentUser with zero require_permission/check_permission, despite docstrings claiming graph:read. Exposes concept labels/descriptions, evidence quotes, full_text, source provenance (filename/path/hostname), and ingested_by user IDs to any authenticated user incl. read_only. POST /query/polarity-axis and /query/polarity-axis/jobs enqueue + auto-approve worker jobs (up to 8 workers) with attacker-controlled compute parameters.

## Acceptance Criteria
- [ ] Depends(require_permission('graph','read')) added to search, sources/search, concept/{id}, related, connect, connect-by-search, polarity-axis, polarity-axis/jobs.
- [ ] POST /query/cypher gated on require_permission('graph','execute') (tracked with the database issue).
- [ ] polarity-axis job submission additionally enforces a job-creation gate; max_workers and timeout_seconds clamped server-side.
- [ ] Test asserts read_only receives 403 on graph query endpoints; contributor+ succeeds.

---
*Part of the **internet-hardening** cluster. Common context: `docs/security/endpoint-security-audit-2026-05-28.md` (PR #430). Operative auth model: ADR-400.*

aaronsb:
Fixed in PR #444 (merged). Gated against the seeded-grant model (ADR-400); regression tests assert anonymous rejection + real authZ denial.

PR #444, opened 2026-05-28
fix(auth): internet-hardening Phase 1 — close the systemic authZ gaps (#433–#437)

aaronsb:
Closes the five **high** blockers from the 2026-05-28 endpoint security audit, measured against the operative model in **ADR-400**. Tracking: #442. Follows the Phase 0 criticals (#430, #443).

These all share one root cause: handlers that authenticated but never authorized, despite docstrings claiming a permission. Each `(resource, action)` pair below is confirmed seeded in the migrations (no dead-checks).

| # | Surface | Gate added |
|---|---------|-----------|
| #433 | `/database/*` | reads → `database:read`; `POST /query` + `counters/refresh` → `database:execute`. `/database/epoch` documented authenticated-by-design (polling). |
| #434 | `/ingest`, `/ingest/text`, `/ingest/image` | `ingest:create` (closes denial-of-wallet) |
| #435 | `queries.py` (9 endpoints) | reads/search/connect/polarity → `graph:read`; `POST /query/cypher` → `graph:execute` |
| #436 | `documents.py` + `sources.py` | content/source reads → `sources:read`; metadata/search/linkage → `graph:read` |
| #437 | `GET /vocabulary/category-flows`, `GET /ingest/image/health` | the two fully-anonymous endpoints → `vocabulary:read` / `admin:status` |

## Notes
- Endpoints with non-defaulted request-body params use route-level `dependencies=[...]` rather than a signature param (avoids parameter-ordering issues).
- Shared test fixture: added a `read_only` user (id=102) + `auth_headers_readonly` to conftest; used across the new auth tests.
- Reconciled two pre-existing `test_database_counters` handler tests that assumed the old ungated behavior (bypass RBAC; authZ now covered in `test_database_auth.py`).

## Tests
New: `test_database_auth.py`, `test_ingestion_auth.py`, `test_queries_auth.py`, `test_documents_sources_auth.py`, `test_anonymous_endpoints_auth.py` — each asserts anonymous rejection + a real authZ denial (read_only/admin) against seeded grants. **Full API suite: 466 passed, 15 skipped.**

Remaining: Phase 2 (medium — #438/#439/#440) and Phase 3 (low — #441) per the audit's phased plan.

### DIFF
--- a/api/app/routes/database.py
+++ b/api/app/routes/database.py
@@ -182,7 +182,9 @@ def get_database_info(
 
 
 @router.get("/health", response_model=DatabaseHealthResponse)
-def check_database_health():
+def check_database_health(
+    current_user: CurrentUser,
+):
     """
     Check database health and connectivity.
 
@@ -190,6 +192,13 @@ def check_database_health():
     - Basic connectivity (ping test)
     - AGE extension availability
     - Graph schema existence
+
+    **Authorization:** Authenticated-by-design (ADR-400, #444 review). This is a
+    deeper diagnostic than the public liveness probe GET /health: it discloses
+    AGE-extension and graph-schema existence. It is NOT anonymous (closing the
+    pre-exposure hole) but is intentionally not permission-gated — any
+    authenticated user / the CLI may check DB health. The detailed /database/stats
+    and /database/info remain database:read-gated.
 
     Returns:
         DatabaseHealthResponse with overall status and check results
--- a/api/app/routes/database.py
+++ b/api/app/routes/database.py
@@ -7,11 +7,11 @@ Provides REST API access to:
 - Health checks
 """
 
-from fastapi import APIRouter, HTTPException
+from fastapi import APIRouter, Depends, HTTPException
 import logging
 import os
 
-from ..dependencies.auth import CurrentUser
+from ..dependencies.auth import CurrentUser, require_permission
 from ..models.database import (
     DatabaseStatsResponse,
     DatabaseInfoResponse,
@@ -35,7 +35,8 @@ def get_age_client() -> AGEClient:
 
 @router.get("/stats", response_model=DatabaseStatsResponse)
 def get_database_stats(
-    current_user: CurrentUser
+    current_user: CurrentUser,
+    _: None = Depends(require_permission("database", "read")),
 ):
     """
     Get database statistics including node and relationship counts (ADR-060).
@@ -125,7 +126,8 @@ def get_database_stats(
 
 @router.get("/info", response_model=DatabaseInfoResponse)
 def get_database_info(
-    current_user: CurrentUser
+    current_user: CurrentUser,
+    _: None = Depends(require_permission("database", "read")),
 ):
     """
     Get database connection information (ADR-060).
@@ -277,6 +279,12 @@ def get_graph_epoch(
     Designed for polling by clients that need to detect graph changes
     without the overhead of full counter queries.
 
+    **Authorization:** Authenticated-by-design (ADR-400). Unlike the other
+    /database/* endpoints, this is intentionally NOT gated on database:read:
+    it is a cache-invalidation poll used by all authenticated clients (any role,
+    incl. read_only) and exposes only a single opaque monotonic integer. Gating
+    it admin-only would break non-admin pollers. Keep it CurrentUser-only.
+
     Returns:
         {"epoch": <integer>}
     """
@@ -299,7 +307,8 @@ def get_graph_epoch(
 
 @router.get("/counters")
 def get_graph_counters(
-    current_user: CurrentUser
+    current_user: CurrentUser,
+    _: None = Depends(require_permission("database", "read")),
 ):
     """
     Get all graph metrics counters with categorization (ADR-079).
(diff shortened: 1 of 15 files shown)


aaronsb:
## Review: Phase 1 internet-hardening authZ (#433–#437)

**Verdict: ship-able.** Every gate added is reachable (each `(resource,action)` is seeded), the gate choices match the audit's per-finding fixes and ADR-400 intent, both gating styles work, `/database/epoch` left ungated is justified, and the new tests prove the invariant non-vacuously against real RBAC. Findings below are advisories, not blockers — except one completeness gap worth a conscious decision (`/database/health`).

### Seeded-grant verification (no dead-checks)
All gate pairs are seeded in migration 028 / 040 / baseline:

| gate | seeded for | reachable? |
|---|---|---|
| `database:read` | admin, platform_admin (028) | ✅ |
| `database:execute` | platform_admin only (028) | ✅ (admin correctly 403s) |
| `graph:read` | contributor→admin→platform_admin (028+040) | ✅ (read_only excluded by design) |
| `graph:execute` | platform_admin only (028) | ✅ |
| `ingest:create` | contributor+ (028) | ✅ (read_only excluded) |
| `sources:read` | contributor+ (028) | ✅ (read_only excluded) |
| `vocabulary:read` | read_only+ (baseline) | ✅ |
| `admin:status` | admin, platform_admin (028) | ✅ |

Inheritance chain `platform_admin→admin→curator→contributor` confirmed in `PermissionChecker._check_inherited_permissions`, so admin reaching `sources:read`/`graph:read` via contributor inheritance is real, not assumed. `read_only` is standalone — the deliberate `graph:read`/`sources:read`/`ingest:create` lockouts are intended (ADR-400 roles table, audit F4/F5/F6).

### Completeness — every route in the 5 files accounted for
- **database.py**: stats/info/counters→`database:read`; counters/refresh + query→`database:execute`; epoch→authenticated-only (documented); **`/database/health` (line 184) is fully unauthenticated** — no `current_user`, no dependency. Not in scope of #433's stated fix and pre-existing, but the F3 header said "`database/query`, etc." and `/database/health` leaks AGE/schema check results. **Decide consciously:** either add it to ADR-400's public-by-design list or gate it. Currently it is neither.
- **ingest.py / ingest_image.py**: all 3 ingest endpoints + image/health gated. ✅
- **queries.py**: all 9 endpoints gated (8×`graph:read`, cypher×`graph:execute`). ✅
- **documents.py**: search/by-concepts/list/concepts/bulk→`graph:read`; content→`sources:read`; pre-existing `DELETE /{id}`→`sources:delete` (seeded 054). ✅
- **sources.py**: all 4 reads→`sources:read`. ✅
- **vocabulary.py**: only `/category-flows` was in scope (the lone zero-dependency route) and is now gated. Note: vocabulary.py still has other `CurrentUser`-only reads (`/status`, `/types`, `/similar`, `/analyze`, `/category-scores`) — out of scope here, presumably later phases.

### `/database/epoch` ungated — justified
Authenticated-by-design is correct: cache-invalidation poll for all roles, returns a single opaque integer. Documented in the handler and tested (`test_epoch_allowed_for_contributor`). Not a hole.

### Gating styles — both correct
Signature `_: None = Depends(...)` (database/ingest) and route-level `dependencies=[...]` (queries/documents/sources/vocabulary/image-health) both execute. In the signature form every param after the `Depends` default is itself defaulted (`File()`/`Form()`), so no non-defaulted-after-default break. `Depends` imported in all touched files; all compile.

### Test quality — non-vacuous
The denial tests run **real RBAC** (no `bypass_permission_check`) with `ensure_test_users_in_db`, so they'd fail if a gate were removed:
- `test_raw_cypher_denied_for_admin` (database + queries) is the strongest: a plain admin holds `*:read` but not `*:execute`, so 403 specifically proves the execute-tier gate, not just "auth exists."
- `test_*_denied_for_read_only` exercise the standalone-role lockout that auth-only could never catch.
- admin-allowed tests assert `not in (401,403)` (right call — handler may legitimately 404/500 on fake ids; asserting 200 would be flaky).
- `test_database_counters.py` correctly switched to `bypass_permission_check` so handler-logic tests don't fail on the new gate; RBAC for those routes is covered in `test_database_auth.py`. Clean separation.

### Regression to legitimate callers (advisory)
- `web/src/components/home/HomeWorkspace.tsx:66` calls `getDatabaseStats()` on the home workspace for **any** authenticated user. Post-gate, `contributor`/`read_only` get 403 → `.catch` degrades to `null` (no crash) but the stats display silently disappears for non-admins. Intended hardening, but a user-visible change — confirm the home widget is acceptable-to-hide for non-admins or move it behind an admin check in the UI.
- `web/src/components/admin/SystemTab.tsx:171` `refreshDatabaseCounters()` now needs `database:execute` = **platform_admin only**; a plain `admin` will 403 on the refresh button. SystemTab uses `.catch(()=>null)` so no crash, but the button silently no-ops for admin. Worth a UI note or relaxing to `database:read` if admins are expected to refresh. (Audit intent supports execute-tier; flagging the UX mismatch only.)
- CLI `cli/src/api/client.ts` calls these too (`db query`, counters) — same role expectations; `kg db query` now requires platform_admin, consistent with F15's note that the "read-only" db group actually runs mutating Cypher.

### Nits
- `test_ingestion_auth.py` has no positive "contributor allowed" case (only anonymous + read_only denial). Denials are non-vacuous, but a contributor-allowed assertion would lock the positive side too.
- `category-flows` gated on `vocabulary:read` still lets `read_only` trigger two full-graph scans — but that matches audit F7 (goal was closing the *anonymous* hole), so correct as specified.

Issue #233, opened 2026-01-29
Improve kg artifact usability and clarify storage vs artifact boundaries


aaronsb:
## Context

PR #232 added `kg storage` as an admin-level diagnostic tool for S3-compatible object storage. During review, we identified overlap and usability gaps with the existing `kg artifact` command.

**Intended audience split:**
- `kg storage` — admin/platform_admin infrastructure diagnostics (raw S3 objects, integrity checks, retention policies)
- `kg artifact` — user-facing view into computed results (polarity analyses, projections, search results)

## Artifact usability issues

1. **Regenerate not exposed in CLI** — `POST /artifacts/{id}/regenerate` exists in the API but `kg artifact` has no `regenerate` subcommand. Users must use curl or the web UI to re-run stale analyses.

2. **Stale artifact guidance** — `kg artifact list` shows freshness with a symbol but doesn't explain what "stale" means or how to fix it (regenerate, re-run the analysis, etc.)

3. **Inline vs Garage distinction is visible but unhelpful** — The `show` command exposes storage location (inline/garage) which is an implementation detail users don't need. The important info is: type, name, freshness, and how to get the payload.

4. **No bulk operations** — Can't delete multiple stale artifacts, export results, or clean up old analyses.

5. **`kg artifact create` is test-only** — Real artifacts are created by async jobs (polarity, projection). The create command is only useful for developer testing, which could be confusing for end users.

6. **Overlap with `kg storage`** — Both can inspect artifact S3 objects:
   - `kg storage inspect artifacts/polarity_analysis/42.json` (raw S3 metadata)
   - `kg artifact show 42` (semantic metadata)
   
   This is fine given the audience split, but worth documenting.

## Suggested improvements

- [ ] Add `kg artifact regenerate <id>` subcommand
- [ ] Improve stale artifact messaging (explain what stale means, suggest regenerate)
- [ ] Hide storage implementation details from default `show` output (or move to `--verbose`)
- [ ] Consider `kg artifact cleanup` for bulk stale artifact removal
- [ ] Document the `kg storage` (admin) vs `kg artifact` (user) distinction in CLI help text

aaronsb:
## Recast under ADR-207 — freshness half done, CLI usability remains

[...]

PR #447, opened 2026-05-30
feat: ADR-501 Catalog Browse Facade (backend + CLI/MCP + web + FUSE)

aaronsb:
## Summary

Implements **ADR-501: Catalog Browse Facade** — a deterministic, hierarchical "what's actually in here?" view of the knowledge graph (`ontology → document → concept`), surfaced across all four clients from one shared API contract.

Browse complements (does not replace) semantic search: it's structural, cheap, and read-only. Membership is projected from the graph's **canonical edges** (`:SCOPED_BY` / `:HAS_SOURCE` / `:APPEARS`), never the denormalized `Source.document` string — so it stays correct under autonomous annealing (ADR-200). Closes the gap behind #233.

## What's included (6 commits, one per layer)

- **ADR-501** (Proposed) — design, with adjacency to ADR-700 (Ontology Explorer consumes this), ADR-069 (FUSE), ADR-500 (DSL seam), ADR-201 (why not graph_accel).
- **Backend** — `CatalogNode` DTO; `catalog_facade.py` (epoch-invalidated materialized index `catalog_node`+`catalog_edge`, DAG-correct projection, pg_trgm fragment filter, advisory-locked rebuild); `/catalog/children` + `/catalog/node/{id}` with `catalog:read` RBAC; migration 073. **16 pytest passing.**
- **CLI + MCP** (shared TS client per ADR-013) — `kg catalog ls/stat` (filesystem-like, `--json`); `catalog` MCP tool (ls/stat actions). **46 CLI tests passing.**
- **Web** — Catalog Explorer: lazy-expanding tree + per-kind detail panel, registered via the ExplorerPlugin pattern; substrate for ADR-700.
- **FUSE** — driver's deterministic half (`/ontology/`, `documents/`) reads from the catalog via a pure, dependency-free adapter. **12 adapter tests passing.**
- **Detail-panel enrichment** — reusing existing endpoints only (no new paths): concept → description/grounding/clickable search-term chips; document → content preview + copy/download (+ image download); ontology → stats. Copy-ID on every node.

## Freshness model

Every index row carries the `graph_change_counter` it was built at (same signal as artifact freshness, ADR-035/203). The facade rebuilds on epoch advance; counts may briefly lag after an annealing cycle and self-heal. Membership is always rebuilt live from canonical edges — bounded staleness on *counts*, never wrong membership.

## Verification

- **Python** 16 ✓ · **CLI** 46 ✓ · **FUSE adapter** 12 ✓ · **web** `tsc` build clean.
- Exercised end-to-end against a **real ingested document** (1 ontology → 1 doc → 32 concepts): CLI drill-down + fragment filter + stat; web tree + per-kind detail panel (verified in-browser, no console errors); migration 073 applied to dev DB; `/catalog` over HTTP returns 401/200/400/404 correctly.

## NOT verified in this session (see follow-up issue)

- **Live FUSE mount** — `pyfuse3` isn't importable in the dev environment, and the FUSE driver publishes separately (installed 0.11.0). Adapter logic is unit-tested; the mounted `ls /mnt/knowledge/ontology/.../documents/` path needs verification next session after publish.

## Reuse notes

Leans on existing infrastructure throughout — shared `apiClient`/`KnowledgeGraphClient`, `IconRailPanel`, `SidebarItem`, lucide icons, Tailwind tokens, and existing endpoints (`getConceptDetails`, `getDocumentContent`, `listOntologies`, `getSourceImageUrl`). No duplicate functions or new backend paths for the enrichment.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/api/app/lib/catalog_facade.py
+++ b/api/app/lib/catalog_facade.py
@@ -39,7 +39,7 @@ _CATALOG_REBUILD_LOCK = 0x0CA7A106  # "CATALOG" mnemonic
 _SORT_SQL = {
     "name": "n.name ASC",
     "child_count": "n.child_count DESC, n.name ASC",
-    "created": "(n.properties->>'creation_epoch') NULLS LAST, n.indexed_at ASC, n.name ASC",
+    "created": "(n.properties->>'creation_epoch')::bigint DESC NULLS LAST, n.indexed_at ASC, n.name ASC",
 }
 
 
@@ -85,47 +85,47 @@ class CatalogFacade:
         """
         conn = self.client.pool.getconn()
         try:
+            # Fast path: read epochs without the lock.
             with conn.cursor(cursor_factory=RealDictCursor) as cur:
                 current = self._current_epoch(cur)
                 indexed = self._index_epoch(cur)
-                conn.commit()
+            conn.rollback()  # close the read txn cleanly
 
             if indexed is not None and indexed >= current:
                 return (False, current)
 
-            # Stale (or empty). Try to claim the rebuild lock; if another worker
-            # holds it, serve what we have.
-            with conn.cursor() as cur:
+            # Stale (or empty). Claim a TRANSACTION-scoped advisory lock: it is
+            # released automatically when this transaction commits or rolls back
+            # (including on exception), so there is no orphaned-lock risk and no
+            # dependency on a manual unlock or on reusing one connection. The
+            # rebuild's TRUNCATE+INSERT runs in this same transaction, so the
+            # lock is held for exactly the rebuild's duration and the single
+            # commit both publishes the new index and frees the lock.
+            with conn.cursor(cursor_factory=RealDictCursor) as cur:
                 cur.execute(
-                    "SELECT pg_try_advisory_lock(%s)", (_CATALOG_REBUILD_LOCK,)
+                    "SELECT pg_try_advisory_xact_lock(%s) AS got", (_CATALOG_REBUILD_LOCK,)
                 )
-                got_lock = cur.fetchone()[0]
-                conn.commit()
+                got_lock = cur.fetchone()["got"]
+                if not got_lock:
+                    # Another worker is rebuilding — serve what we have.
+                    conn.rollback()
+                    return (indexed is not None, current)
 
-            if not got_lock:
-                return (indexed is not None, current)
-
-            try:
                 # Re-check under lock: another worker may have just rebuilt.
-                with conn.cursor(cursor_factory=RealDictCursor) as cur:
-                    indexed = self._index_epoch(cur)
-                    current = self._current_epoch(cur)
-                    conn.commit()
+                indexed = self._index_epoch(cur)
+                current = self._current_epoch(cur)
                 if indexed is not None and indexed >= current:
+                    conn.rollback()  # releases the lock
                     return (False, current)
 
-                self._rebuild(conn, current)
-                return (False, current)
-            finally:
-                with conn.cursor() as cur:
-                    cur.execute(
-                        "SELECT pg_advisory_unlock(%s)", (_CATALOG_REBUILD_LOCK,)
-                    )
-                    conn.commit()
+            # Same transaction (lock still held): rebuild, then commit once.
+            self._rebuild(conn, current)
+            conn.commit()  # publishes the index and releases the xact lock
+            return (False, current)
         except Exception as e:
             logger.error(f"catalog ensure_fresh failed: {e}", exc_info=True)
-            conn.rollback()
-            # Best-effort: report not-stale with epoch 0 so callers still serve
+            conn.rollback()  # releases the lock if held
+            # Best-effort: report stale with epoch 0 so callers still serve
             # whatever index exists rather than erroring the browse request.
             return (True, 0)
         finally:
@@ -160,7 +160,10 @@ class CatalogFacade:
                 f"absent from document drill-down"
             )
 
-        # Atomic swap: truncate + bulk insert in one transaction.
+        # Atomic swap: truncate + bulk insert. The caller (ensure_fresh) owns
+        # the transaction and the advisory lock — it commits once after this
+        # returns, which both publishes the index and releases the lock. We do
+        # NOT commit here, so a failure leaves the prior index intact on rollback.
         from psycopg2.extras import execute_values, Json
 
         with conn.cursor() as cur:
@@ -192,7 +195,6 @@ class CatalogFacade:
                     edges,
                     page_size=1000,
                 )
-            conn.commit()
 
         logger.info(
             f"catalog rebuild complete: {len(nodes)} nodes, {len(edges)} edges "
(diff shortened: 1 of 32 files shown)


aaronsb:
## Code Review — ADR-501 Catalog Browse Facade

Large but well-structured change. Architecture is sound: a relational projection of canonical graph edges, rebuilt on epoch advance, consumed by four thin client wrappers. The pure-projection split (`_project`) for testability is a genuinely good call, and the canonical-edge invariant is correctly honored. One real correctness bug, a few robustness fixes, some polish.

### BLOCKER
None. The `created`-sort bug and the lock-lifetime items below are the closest to blocking — please address before merge.

### SHOULD-FIX

**S1 — `created` sort orders a numeric epoch as text (`catalog_facade.py:42`)**
`creation_epoch` is an int (ontology.py:128, migration 044) but `properties->>'creation_epoch'` extracts text, so it sorts lexicographically (`10` before `9`, `100` before `20`). Leading key also has no explicit direction. Fix:
```python
"created": "(n.properties->>'creation_epoch')::int DESC NULLS LAST, n.indexed_at DESC, n.name ASC",
```
Stays fully whitelisted (no user input). Add a unit assertion that epochs 9 vs 10 order correctly.

**S2 — advisory lock lifetime is coupled to a shared `conn`, undocumented (`catalog_facade.py:98-118`)**
The TRUNCATE+INSERT is atomic (one txn, :167-195) — no half-truncated-index window, that part is correct. But the session-scoped `pg_advisory_lock` is acquired on a separately-committed cursor (:98-103) and survives only because `ensure_fresh` happens to hold one `conn` for the whole sequence. A future refactor of `_rebuild` to grab its own pooled connection would silently detach the lock. Recommend switching to `pg_advisory_xact_lock` inside the rebuild transaction, or a load-bearing comment.

**S3 — exception path can orphan the advisory lock (`catalog_facade.py:119-132`)**
If the manual `pg_advisory_unlock` itself throws (connection died mid-rebuild), the session-level lock is orphaned; since connections are pooled, that can wedge all future rebuilds (catalog silently stops refreshing). A transaction-scoped lock (per S2) eliminates this — releases on commit/rollback automatically. Treat S2+S3 as one "use `pg_advisory_xact_lock`" fix.

**S4 — `_fetch_documents` Cartesian fan-out (`catalog_facade.py:301-314`)**
Two independent `OPTIONAL MATCH` legs off `s` (`:SCOPED_BY→o` and `←APPEARS−c`) cross-product before aggregation. `count(DISTINCT c)` / `collect(DISTINCT o)` keep the *values* correct, but intermediate rows blow up (sources × ontologies × concepts) — a rebuild perf cliff on large docs. Split into separate aggregating subqueries; verify with `EXPLAIN ANALYZE`.

**S5 — FUSE name→id cache not invalidated on annealing rename (`directory.py:266-377`)**
On an autonomous ontology rename (ADR-200), `_list_documents(old_name)` misses the map, re-lists ontologies (now under the new name), still misses, and silently returns an empty document list (:375-377). Cold-path fallback handles *new* ontologies but not *renamed* ones. Acceptable-ish, but the silent-empty is confusing; ensure the stale inode is pruned and add a test.

### NICE-TO-HAVE

- **N1** `get_node` cross-kind ambiguity resolution (`catalog_facade.py:456-470`) picks ontology<document<concept on bare id; relies on id-prefix disjointness. Assert that in a test or always pass `kind`.
- **N2** `_current_epoch`/`_index_epoch` dual cursor-shape branching (`:60-75`) — the tuple branch is unreachable; mild YAGNI.
- **N3** `CatalogNode` TS type duplicated across `cli/` and `web/` — **acceptable.** Separate build graphs (ADR-013), 4-field mirror of the Pydantic source of truth, both annotated as mirrors. A shared package would add more coupling than it removes. If drift bites, generate from OpenAPI.
- **N4** `q` fragment LIKE doesn't escape `%`/`_` (`:348`) — value is correctly parameterized (no injection), but wildcards in user input leak into matching. Minor UX; escape for literal matching.
- **N5** `OntologyDetail` enriches by name not id (`CatalogDetailPanel.tsx:330`) — same rename-race as S5, best-effort with graceful null. Add a comment noting the name-keyed assumption.

### Verified correct (no need to re-check)
- **Canonical edges:** all three fetch passes traverse `:SCOPED_BY`/`:HAS_SOURCE`/`:APPEARS` only; `Source.document` never read for membership (`d.ontology` collected as display-only property, fine).
- **SQL safety:** ORDER BY whitelisted + route-validated; every WHERE/LIMIT/OFFSET value parameterized; no f-string injection of caller data.
- **DAG:** `_project` emits N edges for N parents; child_count semantics correct; tests pin the invariants incl. sourceless/orphan accounting.
- **RBAC:** migration 073 matches the 042 pattern (bare `ON CONFLICT DO NOTHING` resolves on the partial unique index); both endpoints gated; auth tests assert 401/403.
- **Index atomicity:** single-txn TRUNCATE+INSERT; failed rebuild leaves prior index intact; serve-stale fallback sound.
- **Web effects:** all fetches use cancelled-flag + cleanup; debounce clears timer and guards value change; store correctly not persisted. No stale-closure bugs.
- **Detail panel adds no new API paths:** reuses getCatalogNode / getConceptDetails / getDocumentContent / getSourceImageUrl / listOntologies.
- **FUSE image split preserved:** image→(raw + `.md`) two-entry path intact; adapter pure + unit-tested.

---
*AI-assisted review via Claude Code (code-reviewer). Reviewer only — no code modified.*

aaronsb:
## Code review addressed — all 5 findings fixed in 699caa8d

Thanks for the thorough pass. No blockers; all five resolved:

- **S1** (created sort as text): cast to `::bigint DESC NULLS LAST`.
- **S2/S3** (session advisory lock orphan risk): switched to `pg_try_advisory_xact_lock` — auto-released on commit/rollback. `_rebuild` no longer commits; `ensure_fresh` owns the single commit that publishes the index *and* frees the lock. TRUNCATE+INSERT stays atomic; failure rolls back to the prior index.
- **S4** (Cartesian fan-out in `_fetch_documents`): the ontology and concept legs are now aggregated in separate `WITH` stages, so cost is additive instead of `sources×ontologies×concepts`.
- **S5** (FUSE stale name→id on rename): the document-list miss path now refreshes the name→id map directly from the catalog root (bypassing the dir-cache early return) and logs a clear "renamed or removed" warning instead of silently listing empty.

Verified after the fixes: 16 API tests + 12 FUSE adapter tests green; live drill-down, rebuild under the xact lock, and `created` sort all correct against the real ingested document.

On the non-issues you confirmed (no `Source.document` membership leak; duplicate `CatalogNode` TS type acceptable across the separate `cli/` and `web/` build graphs) — agreed, left as-is.

aaronsb:
Deferred FUSE live-mount verification tracked in #448 (adapter is unit-tested; the mounted path needs verification after the next kg-fuse publish, since pyfuse3 isn't importable in the dev env).

Issue #459, opened 2026-05-31
Design: a primordial/default ontology as an ingestion fallback


aaronsb:
## Idea

There should always be a **primordial / default ontology** — an always-present collection that ingestion can fall back to when the target ontology is unexpectedly absent at execute time, instead of failing the job with a 'missing target' anomaly and nowhere to land.

## Context / motivation

Surfaced while validating #457 (vision facade collapse). The #402 Defect B1 guard in `workers/ingestion_worker.py` raises when an ontology "existed at submit but was missing at execute" — correctly refusing to silently recreate a dissolved/deleted ontology. That guard is right to refuse *silent recreation*, but it also means a job can dead-end with no safe landing zone.

A guaranteed default ontology would give such jobs a deterministic home (with an operator-visible signal that the original target was missing), rather than a hard failure. It parallels how many systems keep a 'default'/'uncategorized' bucket that can never be deleted.

## Open questions

- Is the default created at platform init and made non-deletable (lifecycle 'frozen'-against-delete)?
- Does fallback re-home the job to the default, or just annotate the failure? (Silent re-home risks hiding data-routing mistakes — the #402 guard exists precisely to avoid silent behavior.)
- Interaction with annealing/dissolve: should a dissolved ontology's future-queued jobs route to default?
- Naming/semantics: 'primordial' vs 'default' vs 'inbox'.

## Scope

Design discussion first (likely an ADR). Not blocking #457. Raised by @aaronsb.

PR #460, opened 2026-05-31
refactor(vision): collapse parallel vision provider hierarchy into AIProvider (#457)

### LINKED ISSUE #457: ADR-802 §4: collapse the parallel vision provider hierarchy into the AIProvider contract
## Context

ADR-802 §4 deferred unifying `vision_providers.py` into the `ai_providers.py` (`AIProvider`) contract, framing it as a "high-risk rewrite of the working extraction path for no immediate benefit." A scoping investigation (ADR-803 session) found that premise is **outdated**: the unified contract already exists.

## Key finding — it's de-duplication, not a rewrite

- **`AIProvider` already declares `describe_image()`** (`api/app/lib/ai_providers.py:421`), and OpenAI/Anthropic/OpenRouter/Mock already implement it. `routes/ingest.py` already calls `get_provider().describe_image()`.
- The whole `VisionProvider` surface (`describe_image`, `get_provider_name`, `validate_api_key`, `list_available_models`) has equivalents on `AIProvider`.
- `_load_api_key` is **100% duplicated** between the two modules; the catalog helpers are ~80% duplicated (the `ai_providers` version is a superset already handling `category="vision"`).
- **Blast radius is small:** only `workers/ingestion_worker.py` is tightly coupled (`get_vision_provider()`); the vision routes use only utility functions; `ingest.py` needs no change.

## The real risk — behavioral fidelity, not size

The two `describe_image` implementations are **not equivalent**, and the ingestion one is research-validated (ADR-057):

| | `vision_providers` (ingestion) | `ai_providers` (ingest.py) |
|---|---|---|
| Prompt | `LITERAL_DESCRIPTION_PROMPT` (transcribe) | `IMAGE_DESCRIPTION_PROMPT` (structure-aware) |
| OpenAI detail | `detail="high"` | absent |
| Temperature | 0.1 | 0.3 |
| Return | `{text, tokens, model, provider}` | `{text, tokens}` |
| Ollama | implemented | **not implemented (raises)** |

A naive factory swap would **silently change ingestion's vision quality** and **drop Ollama vision**.

## Plan (thin adapter + medium refactor, ~half a day)
1. Move `_load_api_key` + vision catalog helpers into `ai_providers` (shared); `vision_providers` re-exports during transition.
2. Reconcile `describe_image`: parameterize prompt/detail/temperature, unify the return shape, implement on `OllamaProvider`.
3. Migrate `ingestion_worker.py` to `get_provider().describe_image(...)` **preserving the literal-prompt / high-detail / low-temp settings**; golden before/after comparison on a real image.
4. Repoint `routes/vision.py` utilities; reduce `vision_providers.py` to `resolve_vision_selection` (ADR-802 §2 policy) or fold it in.
5. Update tests (`test_vision_provider_catalog.py`, `test_vision_routes.py`).

## Acceptance
- [ ] One provider abstraction; `vision_providers.py` parallel classes removed
- [ ] Ingestion vision output unchanged (before/after image comparison)
- [ ] Ollama vision implemented on the unified provider
- [ ] ADR-802 §4 updated (or a follow-up ADR) to reflect the corrected, lower-risk framing

Refs ADR-802 §4, ADR-803, ADR-057.

aaronsb:
Closes #457. Implements ADR-802 §4 (and its 2026-05-31 amendment): vision is no longer a parallel provider hierarchy — it speaks the one `AIProvider` contract.

## What & why

`vision_providers.py` was a parallel fork of `ai_providers.py` (its own `VisionProvider` ABC, per-provider classes, `get_vision_provider` factory, a 100%-duplicated `_load_api_key`). The unified contract already existed (`AIProvider.describe_image` + `routes/ingest.py` used it), so this is **de-duplication + one call-site migration, not a rewrite** — exactly as the ADR-802 §4 amendment predicted.

## Changes

- **Unify `AIProvider.describe_image`** across OpenAI / Anthropic / OpenRouter / Ollama / LocalEmbedding / Mock: parameterized `(model, detail, temperature)`, unified return shape `{text, tokens:{input,output,total}, model, provider}`. Defaults reproduce the prose route's historical behavior byte-for-byte.
  - `detail=None` omits the OpenAI `detail` key (provider "auto"); `model` threads an explicit vision model so the **vision slot resolves independently of extraction** (ADR-802 §2). New `_openai_token_usage()` normalizes token shape.
- **Migrate `workers/ingestion_worker.py`** off `get_vision_provider()` to `resolve_vision_selection() → get_provider(name) → describe_image(..., detail=None, temperature=0.1)`, preserving the ADR-057 literal path exactly.
- **Reduce `vision_providers.py` 799 → 230 lines**: only the selection policy remains (`resolve_vision_selection`, catalog helpers, `LITERAL_DESCRIPTION_PROMPT`).

## Bundled fix (separate commit)

`fix(ingest): align image route ontology handling with prose route` — the dedicated image route had drifted from the prose route: missing the frozen-ontology guard and never setting `ontology_existed_at_submit`, so the #402 anomaly guard tripped on **every image ingest into a new ontology**. Same parallel-fork-drift class; pre-existing; bundled because it blocked end-to-end validation.

## Verification

- **Golden fidelity** (`scripts/development/diagnostics/golden_describe_image.py`): request payloads **byte-identical old-vs-new across 10 images**; live Anthropic descriptions are correct literal transcriptions.
- **Live front-door run** (`kg ingest image` → worker → graph): real images ingested into a new ontology, concepts created. This is the integration coverage for the worker (which isn't unit-tested) and is how three bugs were caught (two migration, one pre-existing drift) — all now fixed.
- **Full suite green**: 1236 passed, 16 skipped. New `test_describe_image_contract.py` (7 tests) pins the parameterization/return-shape; `test_vision_provider_catalog.py` updated for the collapsed surface.

Refs ADR-802 §4, ADR-803, ADR-057. Follow-up design idea filed as #459 (primordial/default ontology fallback).

### DIFF
--- a/api/app/lib/ai_providers.py
+++ b/api/app/lib/ai_providers.py
@@ -214,6 +214,23 @@ def _load_api_key(provider: str, explicit_key: Optional[str] = None, env_var: Op
     return None
 
 
+def _openai_token_usage(response: Any) -> Dict[str, int]:
+    """Normalize an OpenAI/OpenRouter chat usage object to the describe_image
+    token shape ({input_tokens, output_tokens, total_tokens}). Zeros when the
+    SDK omits usage.  @verified ff971ab5"""
+    usage = getattr(response, "usage", None)
+    if not usage:
+        return {"input_tokens": 0, "output_tokens": 0, "total_tokens": 0}
+    prompt = getattr(usage, "prompt_tokens", 0) or 0
+    completion = getattr(usage, "completion_tokens", 0) or 0
+    total = getattr(usage, "total_tokens", 0) or (prompt + completion)
+    return {
+        "input_tokens": prompt,
+        "output_tokens": completion,
+        "total_tokens": total,
+    }
+
+
 def _anthropic_drops_sampling_params(model_id: str) -> bool:
     """True if the model rejects temperature/top_p/top_k (Opus 4.7 family).
 
@@ -418,20 +435,44 @@ class AIProvider(ABC):
         pass
 
     @abstractmethod
-    def describe_image(self, image_data: bytes, prompt: str) -> Dict[str, Any]:
+    def describe_image(
+        self,
+        image_data: bytes,
+        prompt: str,
+        *,
+        model: Optional[str] = None,
+        detail: Optional[str] = "high",
+        temperature: float = 0.3,
+    ) -> Dict[str, Any]:
         """
-        Generate detailed description of an image using multimodal AI.
+        Generate detailed description of an image using multimodal AI (ADR-057).
 
-        Used for ingesting visual content (slides, diagrams, charts) into the
-        knowledge graph by converting them to text descriptions that can be
-        processed by the normal concept extraction pipeline.
+        The single image->prose capability for the whole system — both the
+        synchronous ingest route (`routes/ingest.py`) and the async image
+        worker (`workers/ingestion_worker.py`, the ADR-057-validated path)
+        call this method. The defaults reproduce the route's historical
+        behaviour; the worker passes explicit args to reproduce the literal
+        transcription path (ADR-057): LITERAL prompt, no high-detail, temp 0.1.
 
         Args:
             image_data: Raw image bytes (PNG, JPEG, etc.)
-            prompt: Description prompt (e.g., "Describe this slide in detail")
+            prompt: Description prompt (e.g. IMAGE_DESCRIPTION_PROMPT or the
+                worker's LITERAL_DESCRIPTION_PROMPT). Required — no hidden
+                default, so each call site declares its intent.
+            model: Vision model id. None resolves it from the catalog (ADR-800
+                /801); an explicit id is used as-is. Threading an explicit id is
+                what lets the vision capability slot resolve independently of
+                the extraction model (ADR-802 §2) without re-slaving vision to
+                extraction.
+            detail: OpenAI/OpenRouter image ``detail`` hint. None omits the key
+                (provider "auto"); "high"/"low"/"auto" set it. Ignored by
+                providers without a detail concept (Anthropic, Ollama).
+            temperature: Sampling temperature. Anthropic models in the
+                no-sampling family (Opus 4.7) drop it automatically.
 
         Returns:
-            Dict with 'text' (description) and 'tokens' (usage info)
+            Dict with 'text' (description), 'tokens'
+            ({input_tokens, output_tokens, total_tokens}), 'model', 'provider'.
         """
         pass
 
@@ -789,11 +830,20 @@ class OpenAIProvider(AIProvider):
         except Exception as e:
             raise Exception(f"OpenAI code translation failed: {e}") from e
 
-    def describe_image(self, image_data: bytes, prompt: str) -> Dict[str, Any]:
+    def describe_image(
+        self,
+        image_data: bytes,
+        prompt: str,
+        *,
+        model: Optional[str] = None,
+        detail: Optional[str] = "high",
+        temperature: float = 0.3,
+    ) -> Dict[str, Any]:
         """
-        Describe an image using GPT-4o vision capabilities.
+        Describe an image using GPT-4o vision capabilities.  @verified ff971ab5
 
-        Returns dict with 'text' (description) and 'tokens' (usage info)
+        See AIProvider.describe_image for the full contract. ``detail=None``
+        omits the key (OpenAI "auto"); the worker's literal path passes None.
         """
         import base64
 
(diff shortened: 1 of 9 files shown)


aaronsb:
## Review: ADR-802 §4 vision facade collapse (#457)

Reviewed both commits against the ADR-057 fidelity and ADR-802 §2 independence invariants. The collapse is sound and the three live-run bugs are fixed correctly. No blocking issues. Details below.

### Invariants verified (passing)

**ADR-057 image→prose fidelity (worker path).** Diffed the new worker call against the deleted `OpenAIVisionProvider`/`AnthropicVisionProvider`:
- Old OpenAI vision: LITERAL prompt, `temperature=0.1`, `max_tokens=8192`, no `detail` key. New worker call (`describe_image(LITERAL, model=vision_model, detail=None, temperature=0.1)`) omits the `detail` key when `None` (`ai_providers.py:861`) and sets temp 0.1 — byte-identical request shape.
- Anthropic path likewise preserved, including the Opus-4.7 no-sampling guard (`_anthropic_drops_sampling_params`, `ai_providers.py:1501`).
- Route (prose) path unchanged: `routes/ingest.py:207` relies on the `detail="high"`, `temperature=0.3` defaults, matching prior route behaviour.

**ADR-802 §2 vision-slot independence (no re-slaving to extraction).** Confirmed `get_provider(name)` builds the *named* provider, not the active extraction one: in production mode `get_provider` calls `load_provider_config(provider_name)` when the requested provider differs from active (`ai_providers.py:2960`), so `get_provider("anthropic")` yields an Anthropic provider even with OpenAI active. `resolve_vision_selection()` resolves the name independently (explicit → active vision config → vision-capable extraction default → fail-loud), and the resolved `vision_model` is threaded as `model=` into `describe_image`, where `model or _resolve_provider_model_id(...)` uses it as-is without catalog override (`ai_providers.py:854`). Vision is not re-slaved.

**No dangling references to deleted symbols.** Grepped `api/app` for `VisionProvider`, `get_vision_provider`, the per-provider vision classes, and `get_model_name` on a provider object — none remain. All surviving `get_model_name()` calls are on embedding managers / visual-embedding generators, which is correct.

**Return-shape consumers aligned.** `tokens` is now a dict everywhere. `routes/ingest.py:217` reads `(... or {}).get("total_tokens", 0)`; worker reads `description_response.get("tokens", {})` then `.get("total_tokens", 0)` (`ingestion_worker.py:260`). `_openai_token_usage` zero-fills missing usage gracefully. The old worker already consumed `tokens` as a dict, so that consumer is unchanged.

**Bug fixes are sound, no siblings found.**
- (a) Removed `vision_provider.get_model_name()` (deleted method) → now reads `description_response.get("model")`. Equivalent value (resolved vision model).
- (b) Function-local `get_provider` shadow removed. Verified no *other* function has both a local `get_provider` import and a later module-level `get_provider()` call — only `process_ingestion_job` had both paths, and it now correctly uses the module-level import (worker:21). All other local imports live in single-path functions.
- (c) `ingest_image.py` frozen-check + `ontology_existed_at_submit` correctly mirror `ingest.py`.

24 unit tests pass (`test_describe_image_contract.py` + `test_vision_provider_catalog.py`).

### Should-fix (non-blocking)

**OpenAI `describe_image` hardcodes the data-URL MIME as `image/png`** (`ai_providers.py:858-860`), regardless of the actual bytes. The Anthropic and OpenRouter variants both do magic-byte detection (png/jpeg/gif/webp); OpenAI does not. The old `OpenAIVisionProvider` did detect png-vs-jpeg (else-assume-jpeg). OpenAI's endpoint sniffs the real bytes and ignores the declared MIME, so this is low real risk — but it's an inconsistency across the four providers and a silent behaviour change vs. the deleted code. Recommend mirroring the magic-byte detection already present in the Anthropic/OpenRouter variants for uniformity.

**Golden diagnostic covers only Anthropic.** `golden_describe_image.py` reconstructs and asserts request-payload equality for the Anthropic path only. Given the OpenAI mime observation above, an OpenAI arm in the same request-equality check would be the place that catches exactly that kind of drift. (The request-payload-equality approach itself is the right call given LLM nondeterminism — good design.)

### Nit

- New `describe_image` calls `choice.message.content.strip()` (`ai_providers.py:886`); the deleted OpenAI vision path did not `.strip()`. Whitespace-only, immaterial to downstream extraction — noting for completeness.
- `vision_metadata["provider"]` now stores the lowercase provider id (`"openai"`) vs. the old capitalized display name (`"OpenAI"`). Intentional and documented at worker:314-316; verified nothing reads this field by case-sensitive match. The lowercase form is the more consistent choice (matches `get_provider()`'s expected arg).

### Assessment

Clean facade collapse: `vision_providers.py` 799→230 lines, a genuine SOLID win (one image→prose contract instead of a parallel hierarchy). Fidelity invariants hold for the validated OpenAI + Anthropic paths; independence invariant holds. The two should-fix items are about OpenAI-path consistency/coverage, not correctness of the merged behaviour. Safe to merge; suggest folding the OpenAI MIME detection in either here or as a fast follow.

---
*AI-assisted review via Claude Code*

Issue #422, opened 2026-05-26
Grounding polarity-axis cache: bound warm-path staleness + close include_types cache-key gap


aaronsb:
## Context

PR #420 (Feb 2 cluster, merged) added a warm-cache short-circuit on the three grounding/confidence batch methods that skips `pool.getconn()` when all requested concepts are cached. PR #420's code-review surfaced two related concerns that didn't block merge but should be revisited.

Both touch the same code (`api/app/lib/age_client/grounding.py`, polarity-axis cache + per-concept cache). Filing as one issue because fixing either likely involves the other.

## Concern 1: Warm-path staleness is unbounded in wall-clock time

The Phase-0 short-circuit at `grounding.py:269-274` (per-concept) and `:444-460` (batch) returns from cache when `_grounding_cache_generation` is set and the entry exists. It does **not** re-read `graph_accel.generation` on the warm path. Mirrored at `confidence_analyzer.py:262-288`.

The docstring (now honest, after the post-merge tightening in #420) says: "staleness bounded by the next cold-path call in this process — not by wall-clock time." A workload that repeatedly queries the same hot concept set will serve pre-mutation values indefinitely until something else takes the cold path.

### What's questionable

The optimization landed on principle ("warm cache shouldn't pay a round-trip"), not from a measured performance problem. The DB cost we're avoiding is ~1-2ms against operations that already pay 50-200ms for graph traversal + hydration. We don't have data showing the round-trip is hurting anything.

### Options

**A. Remove the short-circuit.** Cost: ~1-2ms per warm call (one `get_graph_generation()` probe + one `pool.getconn()`). Benefit: linearizable freshness on grounding/confidence reads. Right call if grounding freshness ever causes a real bug.

**B. Time-bounded staleness (middle path).** Keep the short-circuit but force a generation probe every N seconds. Bounds the stale window by wall-clock, not by next cold-path call.

```python
# In _get_polarity_axis / the Phase-0 check:
if time.monotonic() - _last_generation_probe > 30.0:
    # Fall through to cold path to re-read generation
    pass
```

Cost: one probe per N seconds per process per workload. Adds state (`_last_generation_probe`) and one more invariant to test. Bounded staleness with ~zero amortized DB cost.

**C. Keep as-is until measured.** Mark this issue as "sunset condition": revisit when grounding freshness causes a real bug OR when metrics show the round-trip is meaningful for us. Default-keep but with an explicit revisit trigger.

### Recommendation

[...]

Issue #384, opened 2026-05-19
feat(db): add status column to graph_epochs (track completed vs failed)


aaronsb:
## Context

ADR-203 records an epoch event at job *start* so the assigned `event_id` is available to tag Instances during the run. When a job fails mid-extraction (today's example: Anthropic returned malformed JSON ingesting the Wikipedia cat article, see job `job_86112e92d587`), the epoch row remains in `kg_api.graph_epochs` as a forensic record of the attempt, but with zero attributable Instances.

That's honest as forensics — the system *did* attempt to mutate the graph at that moment — but it pollutes downstream analytics. \"Epochs with zero attributable Instances\" surfaces failed jobs as a signal, when really they should be excluded from anchor/hot/stale derivations.

## Proposal

Add `status TEXT` to `kg_api.graph_epochs`:

| value | meaning |
|---|---|
| `in_progress` | Default at insert; set by `record_graph_epoch` |
| `completed` | Set by the originating subsystem when its unit of work finishes successfully |
| `failed` | Set when the unit of work aborts (LLM JSON failure, validation error, etc.) |

Update `ingestion_worker.py` to flip the status at the terminal states. Other kinds (`annealing`, `reasoning`, `edit`) get the same hook when they're wired (see #__non_ingestion_kinds_issue__).

Update `EpochFacade.list_epochs` to optionally filter by status (default: any).

## Related

- ADR-203 §Decision §2 wording on \"one row per job *invocation*\" — would gain status nuance
- PR #382 code review item #4 raised the column as a likely future need
- ADR-203 §6 mentions a `status` column as a follow-up

aaronsb:
Closed by ADR-207 (PR #462, merged). The `graph_epochs.status` column landed (in_progress/completed/failed), and it's the foundation of the committed-prefix watermark `kg_api.get_committed_epoch()` — resolved on every job exit (success/cancel/exception). Verified live: epochs resolve correctly and the watermark tolerates out-of-order completion + sequence gaps.

Issue #386, opened 2026-05-19
feat: record graph_epochs events for annealing / reasoning / edit kinds


aaronsb:
## Context

ADR-203 defines four event kinds and `graph_epoch_kinds` carries rows for all of them (migration 064), but only `ingestion` actually records events. The other three kinds are reserved but unwired.

### \`annealing\` — ontology annealing
- **Hook**: `api/app/services/annealing_manager.py`, at the start of each annealing cycle.
- **Granularity**: one row per cycle (matching the per-job pattern for ingestion).
- **Actor**: `system:annealing` or similar.
- **Metadata**: ontology being annealed, proposals count, etc.

### \`reasoning\` — agent-driven graph mutations
- **Hook**: write paths in MCP `graph` / `program` tools — specifically the API endpoints they call (`POST /concepts`, `POST /edges`, `POST /graph/batch`).
- **Granularity**: per session, not per call. Open question: how to scope \"session\" — by JWT subject + time window? Worth a short design note.
- **Actor**: agent session ID.

### \`edit\` — manual concept/edge CRUD
- **Hook**: route handlers in `api/app/routes/concepts.py`, `edges.py`, `graph.py`.
- **Granularity**: per mutation. Simpler than reasoning — each call is its own edit event.
- **Actor**: user ID from `current_user`.

## Cross-cutting

- Tag created nodes with `created_at_event_id` for all kinds, not just ingestion. The current `IngestionMixin.create_instance_node` is the only writer that takes the tag; concept/edge creation methods would need the same parameter threaded through.
- For `reasoning`/`annealing` kinds the `semantic_wallclock=false` flag (from `graph_epoch_kinds`) is already set, so the formatter correctly flags occurred_at as forensic.

## Related

- ADR-203 §Decision §2 (per-kind granularity is the open design question)
- ADR-200 (ontology annealing — natural pairing for `annealing`)
- Migration 064 (`graph_epoch_kinds` rows already exist for these kinds)

aaronsb:
Closed by ADR-207 (PR #462, merged). Every graph mutation kind now records an epoch event, making the `event_id` tick universal. One `AGEClient.record_mutation()` helper records a completed epoch + invalidates graph_accel + refreshes the counter, wired into ingestion, annealing (proposal executor), all edit routes, and database restore. Verified live: an edge edit advanced the tick and the catalog rebuilt on next browse.

Issue #278, opened 2026-02-02
Extract grounding/caching cluster from query.py into dedicated mixin


aaronsb:
## Context

PR #276 code review (finding #6) noted `query.py` grew to 1126 lines. The grounding and caching code forms a natural cluster:

- `_get_polarity_axis()` — tier-1 polarity cache
- `_get_graph_generation()` — generation reading with SAVEPOINT
- `calculate_grounding_strength_semantic()` — per-concept grounding
- `calculate_grounding_strength_batch()` — batch grounding with chunking
- Module-level cache globals (`_polarity_cache`, `_grounding_cache`, locks)
- `_parse_embedding()`, `_parse_confidence()` helpers

This cluster has no dependency on the rest of QueryMixin (search, learned knowledge) and could be a `GroundingMixin` in `api/app/lib/age_client/grounding.py`.

## Also Includes

- **Warm cache optimization** (review finding #7): Skip pool connection when all concepts are cache hits. Currently acquires a connection just to read graph generation even on 100% warm.
- **Docstrings**: Document the two-tier cache invalidation model on key methods:
  - `calculate_grounding_strength_semantic()` — polarity axis caching, vocab generation check
  - `GraphFacade._execute_sql()` — pinned connection, generation-based ensure_fresh
  - `graph_facade.py` module docstring — full pipeline: topology (sub-ms) → cached hydration

## References

- PR #276 (ADR-201 Phase 5f)
- Phase 5e in `.claude/todo-adr-201-graph-accel.md`

PR #462, opened 2026-05-31
ADR-207: Uniform freshness contract for materialized graph derivations

aaronsb:
## What

Draft ADR (status: **Proposed**) naming and deciding a structural problem: the platform has grown three independent **materialized derivations of the graph** — the grounding/polarity cache, artifacts, and the catalog index — each re-implementing the same stamp → compare → detect → reconcile freshness lifecycle, against version signals that aren't even the same clock.

## Why it matters

- Two of the three read `graph_change_counter`, which **ADR-203 already proved non-injective** ("delete a concept and add one — the counter is unchanged"). So a stale derivation can read **false-FRESH**. This is a latent correctness bug, not a style issue.
- The grounding cache skips the compare on its warm path → **unbounded staleness** (#422).
- Artifact reconcile (regenerate/cleanup) exists in the API but isn't exposed; storage tier leaks into the user view (#233).
- The honest clock (the ADR-203 epoch event log) can't yet tell a committed mutation from a failed one (#384).

Structurally this is the same parallel-hierarchy problem ADR-802 just collapsed for providers.

## What it decides

- **D1** — canonical clock = the ADR-203 epoch event-log sequence (`graph_epochs.event_id`, monotonic + unique → injective both ways), resolved via a commit-gated `get_committed_epoch()`. `graph_change_counter` demoted to a one-directional dirty-hint. **#384 is a prerequisite.**
- **D2** — mandatory deferred/on-read detection (no warm-path skip), explicit per-derivation staleness budgets, full-recompute baseline (IVM permitted, out of scope).
- **D3** — one `MaterializedDerivation` interface + registry + CI conformance test; uniform reconcile surface subsumes #233's items.

Execution is framed as **one PR / a coherent commit narrative**, not phased project work — the only load-bearing constraint is "trustworthy clock before any migration."

## Grounded in prior art

Named via the materialized-view-maintenance / bounded-staleness (PBS) / cache-coherence literature rather than invented — references in the ADR.

## Related issues
- #384 — epoch `status` (the trustworthy clock) — prerequisite
- #422 — grounding-cache warm-path staleness
- #233 — artifact lifecycle exposure + storage transparency (recast as part of this work, **not** closed by it)

_Review focus: is D1 (epoch sequence as canonical clock) the right call, and is the `graph_epoch`-column data migration (counter → committed sequence) safe for existing artifact + catalog rows?_

### DIFF
--- a/api/app/lib/age_client/grounding.py
+++ b/api/app/lib/age_client/grounding.py
@@ -42,7 +42,12 @@ import numpy as np
 from psycopg2 import extras
 
 from api.app.constants import BATCH_CHUNK_SIZE
-from .graph_generation import get_graph_generation
+from api.app.lib.freshness import (
+    Budget,
+    CollectionDerivation,
+    read_committed_epoch,
+    register_derivation,
+)
 
 logger = logging.getLogger(__name__)
 
@@ -269,25 +274,19 @@ class GroundingMixin:
         """
         global _grounding_cache, _grounding_cache_generation
 
-        # Warm-cache short-circuit (ADR-201 Phase 5f #278) — see the longer
-        # comment on calculate_grounding_strength_batch for the soundness
-        # rationale. Per-concept callers benefit even more from this:
-        # rendering a search result page typically calls this method
-        # serially per concept, so the bulk of repeat queries land here.
-        with _grounding_cache_lock:
-            cached_gen = _grounding_cache_generation
-            if cached_gen is not None:
-                cache_hit = _grounding_cache.get((concept_id, cached_gen))
-                if cache_hit is not None:
-                    return cache_hit
-
         conn = self.pool.getconn()
         try:
             with conn.cursor(cursor_factory=extras.RealDictCursor) as cur:
                 # ---- Tier 2 cache check: per-concept grounding ----
-                # Read graph generation from graph_accel if available,
-                # otherwise fall back to a simple counter.
-                graph_gen = get_graph_generation(cur)
+                # ADR-207 D2: read the canonical freshness clock (the committed-
+                # epoch tick) on EVERY call. The previous connection-free warm-
+                # path short-circuit skipped this re-read, leaving staleness
+                # unbounded in wall-clock time (#422). The cache still saves the
+                # expensive edge fetch + projection — it just no longer skips the
+                # clock read. graph_gen here is the universal tick, not
+                # graph_accel.generation (which co-advances with it as a
+                # sub-counter, ADR-207 D1).
+                graph_gen = read_committed_epoch(cur)
 
                 # Evict entire cache if graph generation changed
                 with _grounding_cache_lock:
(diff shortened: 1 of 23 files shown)


aaronsb:
## Code Review — ADR-207 Freshness Contract (reviewer pass before merge)

Reviewed the full diff (~1850 lines) against `main`: migration 076, `freshness.py`, the `record_epoch`/`complete_epoch`/`record_mutation` helpers, both workers, all six converted routes, the catalog facade, grounding + confidence derivations, and `artifact_freshness.py`. The contract design is clean and the SOLID split (Collection vs Instance over a shared base) is well-judged — the registry + conformance test is exactly the right CI gate. Findings below, ranked.

---

### BLOCKER

**B1 — Full database restore never advances the tick (`api/app/workers/restore_worker.py:164-178`)**

The restore worker is the single largest possible graph mutation (it `_clear_database()` + re-imports the entire graph), yet it only calls `refresh_graph_metrics()`. It never calls `record_mutation` / `record_epoch`, so `kg_api.get_committed_epoch()` does **not** advance across a restore. Every derivation in this PR keys on that watermark:

- catalog index: `version_stamp()` reads `MAX(graph_epoch)` from `catalog_node`. After a restore the watermark is unchanged, `indexed >= current` stays true, and `ensure_fresh()` is a no-op — the browse index reflects the *pre-restore* graph until something else ticks.
- grounding / confidence caches: `_grounding_cache_generation == graph_gen` stays true → stale cached values served against a completely different graph.
- artifacts: `is_fresh()` reports fresh against the wrong graph.

This is the exact "derivation goes stale-undetected" failure the contract was built to prevent, and it is reachable through the normal restore path. The whole point of #386 ("make the tick universal across all mutation kinds") is undercut if restore — the most destructive mutation — is the one path that doesn't announce.

**Fix:** after a successful restore, call `metrics_client.record_mutation("edit", actor=...)` (or a dedicated kind if you want restore distinguishable in analytics — note the `kind` CHECK constraint in migration 063 only allows `ingestion|reasoning|annealing|edit`, so a new kind needs a constraint widen). It must run on the same `metrics_client` you already construct at line 167, before `close()`.

---

### SHOULD-FIX

**S1 — `record_epoch` can leak a pool connection on the error path (`api/app/lib/age_client/ingestion.py:63-79`)**

```python
conn = self.pool.getconn()
try:
    with conn.cursor() as cur:
        cur.execute(...)        # raises → enters finally
        return row[0] if row else None
finally:
    conn.commit()               # commit on an aborted txn
    self.pool.putconn(conn)     # skipped if commit() raises
```

Two issues: (a) `conn.commit()` runs unconditionally in `finally` even when the cursor block raised — on an aborted transaction this is at best a disguised rollback, and (b) if `commit()` itself raises, `putconn` is skipped and the connection leaks from the pool. A steady-state `record_epoch` failure rate (which the docstring explicitly anticipates and logs at ERROR) would then slowly exhaust the pool. `complete_epoch` (lines 108-118) has the identical structure. Prefer explicit `conn.rollback()` on exception and put `putconn` in its own guaranteed-reached block (e.g. nested try/finally, or `conn.commit()` inside the `with` on success only).

**S2 — `record_mutation` is non-atomic across two transactions, transiently regressing the watermark (`ingestion.py:152-165`)**

`record_mutation` does `record_epoch` (INSERT `in_progress`, txn 1, commit) → `complete_epoch` (UPDATE `completed`, txn 2, commit). Between the two commits the row is `in_progress`, so `get_committed_epoch()` returns `MIN(in_flight)-1` — i.e. the watermark momentarily **drops below** where it was before this synchronous edit. A concurrent grounding/catalog read landing in that window reads a *lower* watermark than the stamp it already has, which (with the monotonic-stamp assumption baked into `Budget.tolerates`, where `current - stamp <= versions` and a negative delta reads "fresh") can mask the very mutation just made until the second commit lands. For a helper documented as "an already-committed atomic graph mutation," recording it as in_progress-then-completed is a semantic mismatch. Consider recording synchronous mutations as `completed` in a single statement (a `record_completed_graph_epoch()` SQL helper that inserts with `status='completed'` directly), reserving the in_progress/complete pair for genuinely long-running jobs.

**S3 — Watermark depression makes synchronous edits invisible while any ingestion is in flight**

Because `get_committed_epoch()` returns `MIN(in_flight)-1` whenever *any* job is in flight, a synchronous edit's `record_mutation` (S2) advances `MAX(event_id)` but the watermark is still pinned below the long-running ingestion's id. So an interactive edit made during a 20-minute ingestion is **not detectable** by any derivation until that ingestion completes — caches keep serving pre-edit values for the duration. This is a direct consequence of the chosen watermark semantics and is arguably correct-by-design (per-chunk-commit reasoning in the migration comment), but it deserves an explicit callout in ADR-207 D1/D2: the staleness window for an edit is bounded by *the longest concurrent in-flight job*, not by wall-clock. Worth confirming this is the intended contract, since the PR's framing ("freshness detected on read") could be read as implying tighter bounds.

---

### NICE-TO-HAVE

**N1 — `get_committed_epoch()` marked `STABLE` (`076_freshness_clock.sql:113`)** — correct within a statement, but note it reads `graph_epochs` which can change between statements in a longer txn; the catalog's `ensure_fresh` reads it twice (fast path + under lock) and relies on it possibly differing, which is fine because those are separate statements. No change needed, just flagging the `STABLE` label is doing real work here and shouldn't be "optimized" to `IMMUTABLE`.

**N2 — Migration back-fill predicate `occurred_at < NOW()` (`076:37-40`)** is always true for existing rows (they were inserted in the past), so the `AND occurred_at < NOW()` clause is a no-op guard. Harmless, but the intent ("mark all pre-existing in_progress rows completed") is clearer as just `WHERE status = 'in_progress'`. If the guard was meant to avoid racing a row inserted by a concurrently-deploying app instance, that's not what `< NOW()` achieves.

**N3 — Hot-path cost of removing the warm short-circuit** — every grounding/confidence call now does `pool.getconn()` + `get_committed_epoch()` even on a 100%-warm cache. The partial index `idx_graph_epochs_in_flight` keeps the watermark query cheap, and the trade-off (bounded staleness over an unbounded window) is the right call per #422. But the connection acquisition itself is now on every call; under high read concurrency this competes for pool slots with ingestion. Worth a benchmark note in the ADR, and consider whether a very-short-TTL (sub-second) memoization of the watermark read — explicitly bounded, unlike the old #278 short-circuit — would recover most of the cost without reintroducing unbounded staleness. Not blocking.

**N4 — Artifact false-stale during in-flight ingestion** — `_get_current_epoch` (`routes/artifacts.py`) stamps a newly-created artifact at the depressed watermark; when the concurrent ingestion completes the watermark jumps and the just-built artifact reads stale. This errs in the safe direction (false-stale, never false-fresh) and triggers at most one benign regeneration, consistent with the migration's stated philosophy. Documenting only.

---

### Assessment

The core design is sound and the conformance test will catch future derivations that forget the contract. The catalog facade's serve-stale-under-advisory-lock pattern is a genuinely good reference implementation. **B1 is a merge blocker** — restore is a graph-mutating path that silently bypasses the universal tick, which is precisely the class of bug this ADR exists to eliminate. S1/S2 are correctness/robustness fixes that should land before merge. S3 is most likely working-as-designed but needs an explicit ADR statement so the staleness bound isn't misread.

*Reviewer pass — analysis only, no files modified.*

PR #469, opened 2026-06-01
feat: operator intake controls (lanes, tombstones) + catalog fix

aaronsb:
## What

Operator controls for steering document intake, plus the catalog fix that was the root cause of an ingest 404 storm. All surfaced while debugging a large ADR ingestion that \"wasn't doing much.\"

### Worker lanes (ADR-100)
- Configurable per-lane slots: `kg admin workers lanes set <lane> --max-slots/--enable/--disable`, capped by `MAX_LANE_SLOTS` (16) defined once in `lane_manager`.
- Shared executor sized to `lanes × cap` so raising slots actually increases concurrency (was pinned at `MAX_CONCURRENT_JOBS`=4).
- WebUI: per-lane Freeze/Unfreeze toggle + `Active`/`Frozen` status pill.

### Ontology tombstones
- `GET/DELETE /ontology/tombstones` + `kg ontology tombstones list/flush/clear`.
- `kg ontology delete` now clears its own tombstone by default — a *complete* delete, name is immediately re-ingestable — with `--keep-tombstone` to deliberately block. Fixes the silent "deleted but still blocked re-ingestion" confusion.
- WebUI: Ontologies list with per-ontology delete + Tombstones panel.
- Tombstone deletes carry a `lock_timeout` so they can't hang on active work.

### Catalog staleness (the 404 root cause)
- Migration 059 seeded retired dated Anthropic snapshots as `enabled`. `claude-3-haiku-20240307` (retired 2026-04-19) was the cheapest enabled row, so `prefer_cheapest` resolution (markdown code-block prose translation) silently selected a dead model → every translation call 404'd, while main extraction (`sonnet-4-6`) succeeded. Disabled the two retired snapshots in the seed.
- Sketched the durable fix (self-healing reconcile on catalog Refresh) at `models.refresh_catalog`.

## Why

The platform lacked the controls to steer intake — recovering from a stuck/misconfigured run meant raw SQL (freeze lanes, flush tombstones, disable dead models). This adds the missing instrumentation.

## Deferred
- **Fail-fast on permanent provider errors** (the original branch intent): we learned 404s here can be *transient*, so it needs a transient-vs-permanent distinction, not a blanket fatal.
- **Operator actions as jobs** for eventual consistency.

### DIFF
--- a/.claude/ways/kg/web/way.md
+++ b/.claude/ways/kg/web/way.md
@@ -59,3 +59,13 @@ package.json, tsconfig*.json) — those aren't watched by HMR.
 Stores save to API first, fall back to localStorage:
 - Requires OAuth login for database persistence
 - localStorage serves as cache and offline fallback
+
+## Control UI: Show State, Don't Imply It
+
+<!-- epistemic: heuristic -->
+
+An action-labeled button (e.g. "Freeze") is ambiguous about *current* state —
+does "Freeze" mean it's frozen, or that clicking freezes it? For admin/control
+toggles, render an explicit status indicator (an Active/Frozen pill) next to
+the bare action verb, so the user reads the current state directly instead of
+inferring it from the button label.
--- a/api/app/models/ontology.py
+++ b/api/app/models/ontology.py
@@ -182,6 +182,7 @@ class OntologyDeleteResponse(BaseModel):
     deleted: bool
     sources_deleted: int
     orphaned_concepts_deleted: int
+    tombstone_cleared: bool = False  # True if the deletion tombstone was flushed (name re-ingestable)
     error: Optional[str] = None
 
 
--- a/api/app/routes/admin_workers.py
+++ b/api/app/routes/admin_workers.py
@@ -16,12 +16,18 @@ from psycopg2.extras import RealDictCursor
 
 from ..dependencies.auth import CurrentUser
 from ..services.job_queue import get_job_queue
+from ..services.lane_manager import MAX_LANE_SLOTS
 from ..lib.job_permissions import JobPermissionContext
 
 
 class LaneUpdate(BaseModel):
     """Request body for updating a worker lane."""
-    max_slots: Optional[int] = Field(None, ge=0, le=10, description="Max concurrent jobs")
+    # Cap is MAX_LANE_SLOTS (lane_manager) — the same ceiling the worker pool is
+    # sized against, so a configured value here is always backed by threads.
+    max_slots: Optional[int] = Field(
+        None, ge=0, le=MAX_LANE_SLOTS,
+        description=f"Max concurrent jobs in the lane (0–{MAX_LANE_SLOTS})",
+    )
     poll_interval_ms: Optional[int] = Field(None, ge=500, le=120000, description="Poll interval in ms")
     stale_timeout_minutes: Optional[int] = Field(None, ge=5, le=1440, description="Stale job timeout in minutes")
     enabled: Optional[bool] = Field(None, description="Enable/disable the lane")
--- a/api/app/routes/models.py
+++ b/api/app/routes/models.py
@@ -96,6 +96,39 @@ async def refresh_catalog(
         conn = client.pool.getconn()
         try:
             count = upsert_catalog_entries(conn, entries)
+
+            # --- SKETCH: self-healing catalog reconciliation (not yet implemented) ---
+            # Problem this would solve: upsert is additive. When a provider retires
+            # a dated snapshot, the model vanishes from `entries` but its catalog row
+            # stays enabled=TRUE — so prefer_cheapest/default resolution can still
+            # select a dead model that 404s (this is exactly how claude-3-haiku-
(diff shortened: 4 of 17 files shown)


aaronsb:
## Code Review — operator intake controls (lanes, tombstones) + catalog fix

Thorough pass against \`origin/main\` (5 commits, ~970 additions). No blocking issues. Findings below.

### Verified correct (focus areas)
- **Route ordering**: All three \`/tombstones*\` routes (ontology.py L189–261) are declared before any \`/{ontology_name}\` route (first at L1085). \"tombstones\" cannot be captured as an ontology name. ✓
- **Executor lifecycle**: \`_executor\` is only dereferenced in \`_lane_loop\` (L163), which is only spawned inside \`start()\` *after* the pool is assigned (L86). No None-deref, no race. \`stop()\` now guards \`if self._executor is not None\`. ✓
- **\`SET LOCAL lock_timeout\`**: Pooled connections are \`autocommit=False\` (no override in age_client/base.py), so an implicit transaction is open and \`SET LOCAL\` is honored, not a no-op. Issued inside the txn before the DELETE. ✓
- **Permission gating**: New tombstone routes all require \`ontologies:delete\`; lane PATCH requires \`workers:manage\`. Consistent on the server. ✓
- **Migration 059**: Only two literal \`TRUE\`→\`FALSE\` flips in VALUES; SQL still valid. ✓

---

### Should-fix

**1. The catalog fix does not reach existing databases.** \`schema/migrations/059\` uses \`ON CONFLICT ... DO NOTHING\`, so editing the seed only affects *fresh* installs. Existing deployments (e.g. cube) still have \`claude-3-haiku-20240307\` / \`claude-3-5-sonnet-20241022\` at \`enabled=TRUE\` and will keep 404ing on \`prefer_cheapest\` resolution — the exact bug this PR targets. No companion migration (latest is 076) applies the \`UPDATE\`. Suggest either a new forward migration:
\`\`\`sql
UPDATE kg_api.provider_model_catalog SET enabled = FALSE
 WHERE provider='anthropic' AND model_id IN ('claude-3-haiku-20240307','claude-3-5-sonnet-20241022');
\`\`\`
or confirm the intended remediation is a manual operator disable via the admin UI this PR adds. The in-place 059 edit alone leaves the live bug unfixed on every existing install.

**2. Worker pool sizing vs. DB connection ceiling.** \`pool_size = len(lanes) × MAX_LANE_SLOTS\` = 3 × 16 = 48 threads by default. Each job's worker instantiates its own \`AGEClient()\` with \`ThreadedConnectionPool(1, 20)\`, so this isn't a single-pool exhaustion bug — but it is the connection-proliferation pattern ADR-100 itself calls out. Real concurrency is still bounded by per-lane \`max_slots\` (defaults sum to 4), so the 48-thread ceiling is latent headroom, not active load. Worth a one-line note in the \`MAX_LANE_SLOTS\` docstring that raising lane slots toward the cap on all 3 lanes could push Postgres connection count high (48 workers × per-worker pools). Not blocking at current defaults.

---

### Nits

**3. CLI \`--max-slots\` accepts non-numeric input silently.** workers.ts: \`parseInt(v,10)\` on \`--max-slots abc\` yields \`NaN\`; \`NaN !== undefined\` is true, so \`body.max_slots = NaN\` → serialized as \`null\` → server treats it as \"no change\" rather than rejecting. Consider validating the parsed value (\`Number.isNaN\` guard) and erroring on bad input. Same applies to \`--poll-interval\` / \`--stale-timeout\`.

**4. Hardcoded \`(0–16)\` in CLI help.** workers.ts option help strings embed the cap literally; if \`MAX_LANE_SLOTS\` changes server-side, the help drifts. Low priority — the server is the source of truth and rejects over-cap values with a clear validation error (which the CLI now renders nicely).

**5. SystemTab freeze/unfreeze button has no client-side permission gate.** Unlike OntologyTab (which gates on \`canDelete\`), SystemTab.tsx has no \`hasPermission('workers','manage')\` check — a user lacking the permission sees the button and only discovers the block on click (server 403). The tab is admin-scoped so impact is low, but gating it would match the pattern this PR establishes one file over.

**6. \`list_ontology_tombstones\` leaves a read txn uncommitted before \`putconn\`.** ontology.py L204–213: the SELECT runs under the implicit txn and isn't rolled back/committed before \`putconn\`. Harmless here because \`client.close()\` → \`pool.closeall()\` destroys the connection immediately after, and it matches the existing read-route convention in this file. Mentioning only for completeness.

---

### Assessment
Clean, well-documented change. The lazy-executor refactor, route ordering, lock_timeout usage, and React loading/error/permission handling are all correct. The one thing that genuinely needs a decision before merge is **finding #1** — confirm how existing installs get the retired models disabled, since the 059 edit doesn't do it. Everything else is advisory.

The models.py reconciliation \"sketch\" is comment-only and inert (verified), and the staged design constraints in it are sound — good to capture the durable fix's hazards (fetch completeness, category scoping, never auto-disabling the active/default model) before someone implements it.

---
*AI-assisted review via Claude Code*

PR #478, opened 2026-06-01
ADR-102: Portable backup/restore foundations (spec, engine, validator, DR bugfixes)

### LINKED ISSUE #397: restore-database.sh reports success when pg_restore fails
## Problem

`operator/database/restore-database.sh` runs `pg_restore` inside a pipeline:

```sh
if docker exec -i $CONTAINER pg_restore ... -1 \
    < "$BACKUP_FILE" 2>&1 | grep -v "^WARNING:" | grep -v "already exists"; then
```

The pipeline's exit status is `grep`'s, not `pg_restore`'s. When `pg_restore`
aborts (e.g. a `-1` single-transaction rollback), `grep` still exits 0, so the
script prints `✓ Database restored successfully!` and continues — silent data
loss for anyone relying on the restore.

## How it surfaced

Found during the ADR-205 PostgreSQL 18 migration spike. A `pg_restore --clean`
of an AGE database aborted mid-transaction; the script reported success while
the target database stayed empty (0 concepts).

## Fix direction

Capture `pg_restore`'s real exit code — e.g. `set -o pipefail`, or
`PIPESTATUS[0]`, or write `pg_restore` output to a file and grep that
separately. Verify post-restore (the concept/source counts the script already
prints would have caught this had they not also failed silently).

## Scope

Pre-existing bug, independent of PostgreSQL 18. See ADR-205.

### LINKED ISSUE #398: pg_restore --clean fails on AGE label tables (DROP TABLE rejected)
## Problem

`pg_restore --clean` (used by `operator/database/restore-database.sh`) emits
plain `DROP TABLE IF EXISTS` statements for every table — including Apache AGE
label tables. AGE rejects a direct drop of a label table:

```
pg_restore: error: could not execute query: ERROR:  table "VocabType" is for label "VocabType"
Command was: DROP TABLE IF EXISTS knowledge_graph."VocabType";
```

So a `--clean` restore into any database that already holds an AGE graph
aborts. Restore only succeeds into a database with no pre-existing graph.

## How it surfaced

Found during the ADR-205 PostgreSQL 18 migration spike. Restoring into a
freshly-initialised database (which the container init populates with a graph
via the baseline migrations) fails on the first label table.

## Fix direction

Restore into a genuinely empty database — e.g. `DROP DATABASE` / `CREATE
DATABASE` then `pg_restore` without `--clean` — or drop the graph through AGE
(`SELECT drop_graph(...)`) before a `--clean` restore. The spike confirmed an
empty-database restore completes cleanly.

Note: this is distinct from the separate, deeper finding that AGE graph
identity (OID-coupled) does not survive logical dump/restore across clusters —
see ADR-205. This issue is only about the `--clean` / label-table abort.

## Scope

Pre-existing bug, independent of PostgreSQL 18. See ADR-205.

aaronsb:
## Summary

Foundations for **ADR-102 — Portable Backup and Restore with Clone/Merge Semantics**. Lands the accepted ADR, the format spec, the integration-mode engine, an offline validator, and two restore-script bugfixes. **No change to live restore behavior** — the sequential implementation spine follows on a separate branch.

## Changes
- **ADR-102** (Accepted) — clone/merge by target-emptiness; three restore modes (idempotent / adjacent / integration); phased epoch reconciliation; primary-inputs-in / derived-products-out; self-describing versioned format; post-restore rehydration; engine-agnostic storage; restore-time worker quiescence; single supported path.
- **Backup Object Spec** — `docs/reference/BACKUP_OBJECT_SPEC.md` (`kg-backup/2`): declarative header, dictionary/interning + cascade, record shapes, the `learned_id` remap landmine, version negotiation.
- **ConceptMatcher** ported to `api/app/lib/concept_matcher.py` — the integration-mode engine (cosine attach, no LLM). Not yet wired into the worker.
- **Offline validator** — `scripts/development/lint/lint_backup.py`: pure, importable (`validate_backup`), CI artifact-lint. Reused by future round-trip tests and the post-restore validate endpoint.
- **Fix #397 / #398** — capture `pg_restore`'s real exit code + gate success on graph counts; restore into a freshly recreated DB (avoids AGE label-table `--clean` abort). ⚠️ DR path now DROP/CREATEs the target DB (documented, confirmation-gated).

## Not in this PR (follows on a new branch)
Export + versioned header, restore-core + ID-remap layer, merge modes, epoch reconciliation, verify + purge of superseded code. The Phase 0 audit's reference-class checklist + landmines (`learned_id`, `Source.document` cache, image-key re-derivation) and two latent bugs (silent watermark stall, un-automated lane freeze) are captured for that work.

## Refs
ADR-102 · Closes #397 · Closes #398

### DIFF
--- a/api/app/lib/concept_matcher.py
+++ b/api/app/lib/concept_matcher.py
@@ -1,16 +1,20 @@
 """
 Concept matching engine for restore integration mode (ADR-102 §2).
 
-Ported from the legacy ``api/lib/restitching.py`` ``ConceptMatcher`` and adapted
-to ``api/app`` conventions. This module is the *matching engine only*: given an
-external concept (carrying an embedding, and optionally a label) it finds the
-best-matching existing concept in the target graph by cosine similarity, using
-the project's canonical two-tier thresholds. It performs no LLM calls, generates
-no embeddings, and is not wired into any worker, route, or service — that wiring
-is Phase 4 of ADR-102.
+Succeeds the legacy ``api/lib/restitching.py`` ``ConceptMatcher`` as the
+``api/app`` integration-mode engine. NOTE: this is not a line-for-line port —
+the legacy matcher used a single 0.85 threshold; this module implements the
+project's **canonical two-tier policy** (``api/app/lib/ingestion.py:432-461``),
+which is what ADR-102 §2 specifies for integration mode. It is the *matching
+engine only*: given an external concept (carrying an embedding, and optionally a
+label) it finds the best-matching existing concept in the target graph by cosine
+similarity. It performs no LLM calls, generates no embeddings, and is not wired
+into any worker, route, or service — that wiring is Phase 4 of ADR-102.
 
 The matching policy mirrors the canonical ingestion path
-(``api/app/lib/ingestion.py``):
+(``api/app/lib/ingestion.py``), with one deliberate refinement: ingestion
+inspects only the top candidate, whereas this engine walks candidates best-first
+so a label-boosted match just below a non-matching top hit is not missed:
 
 - **Strict tier (>= 0.85):** any candidate at or above this cosine similarity is
   an unconditional match.
--- a/api/app/lib/concept_matcher.py
+++ b/api/app/lib/concept_matcher.py
@@ -1,0 +1,224 @@
+"""
+Concept matching engine for restore integration mode (ADR-102 §2).
+
+Ported from the legacy ``api/lib/restitching.py`` ``ConceptMatcher`` and adapted
+to ``api/app`` conventions. This module is the *matching engine only*: given an
+external concept (carrying an embedding, and optionally a label) it finds the
+best-matching existing concept in the target graph by cosine similarity, using
+the project's canonical two-tier thresholds. It performs no LLM calls, generates
+no embeddings, and is not wired into any worker, route, or service — that wiring
+is Phase 4 of ADR-102.
+
+The matching policy mirrors the canonical ingestion path
+(``api/app/lib/ingestion.py``):
+
+- **Strict tier (>= 0.85):** any candidate at or above this cosine similarity is
+  an unconditional match.
+- **Label-boosted tier (>= 0.75):** a candidate in ``[0.75, 0.85)`` matches only
+  when its label is equal to, or a substring containment of, the incoming
+  concept's label (case-insensitive, whitespace-normalized).
+
+Differences from the legacy version (see "Adaptation notes" in the ADR-102
+Track C report):
+
+- Uses the ``api/app`` :class:`~api.app.lib.age_client.AGEClient` and its
+  :meth:`vector_search` rather than the legacy ``AGEConnection`` +
+  hand-rolled full-scan loop.
+- Does NOT generate embeddings from a label (the legacy code called
+  ``conn.generate_embedding``). A concept with no embedding cannot be matched by
+  similarity here; embedding rehydration is a separate ADR-102 §6 concern that
+  must run before integration matching.
+- Carries only the matching engine — the restitch-plan / execute-restitch /
+  console-printing surface of the legacy class is intentionally omitted.
+"""
+
+import logging
+from typing import Any, Dict, Optional
+
+from api.app.lib.age_client import AGEClient
+
+logger = logging.getLogger(__name__)
+
+# Canonical two-tier matching thresholds (mirrors api/app/lib/ingestion.py).
+STRICT_THRESHOLD = 0.85
+LABEL_BOOST_THRESHOLD = 0.75
+
+
+def _labels_match(incoming_label: Optional[str], candidate_label: Optional[str]) -> bool:
+    """Decide whether two labels are similar enough to boost a borderline match.
+
+    Normalizes both labels (lowercase, stripped) and returns True when they are
+    identical or one contains the other — the same rule used by the canonical
+    ingestion matcher's label-boost tier.
+
+    Args:
+        incoming_label: Label of the external/incoming concept (may be None).
+        candidate_label: Label of the existing candidate concept (may be None).
+
+    Returns:
+        True if the labels are considered a match for boosting purposes.
+
+    @verified 72ba3d3b
+    """
+    if not incoming_label or not candidate_label:
+        return False
+    norm_incoming = incoming_label.lower().strip()
+    norm_candidate = candidate_label.lower().strip()
+    if not norm_incoming or not norm_candidate:
+        return False
+    return (
+        norm_incoming == norm_candidate
+        or norm_incoming in norm_candidate
+        or norm_candidate in norm_incoming
+    )
+
+
+class ConceptMatcher:
+    """Match external concepts to existing target concepts by cosine similarity.
+
+    This is the ADR-102 integration-mode matching engine. It wraps an
+    :class:`~api.app.lib.age_client.AGEClient` and uses its ``vector_search``
+    (cosine similarity over concept embeddings) to find the best existing
+    concept for an incoming external concept, applying the project's canonical
+    two-tier thresholds (strict 0.85 / label-boosted 0.75).
+
+    The matcher is read-only with respect to the graph: it never writes,
+    creates placeholders, or calls an LLM/embedding service. Concepts that do
+    not meet the threshold return ``None`` (the caller's integration logic then
+    falls through to adjacent-mode behaviour — out of scope here).
+
+    Args:
+        client: The ``api/app`` AGEClient used for Cypher / vector search.
+        strict_threshold: Unconditional-match cosine threshold (default 0.85).
+        label_boost_threshold: Lower threshold below which a label match is
+            also required (default 0.75).
+
+    @verified 72ba3d3b
+    """
+
+    def __init__(
+        self,
+        client: AGEClient,
+        strict_threshold: float = STRICT_THRESHOLD,
+        label_boost_threshold: float = LABEL_BOOST_THRESHOLD,
+    ):
+        """Initialize the matcher with an AGEClient and matching thresholds.
+
+        Args:
+            client: AGEClient for graph access (vector search).
+            strict_threshold: Cosine similarity at/above which any candidate
+                matches unconditionally.
+            label_boost_threshold: Lower cosine similarity bound; candidates in
+                ``[label_boost_threshold, strict_threshold)`` match only if
+                their label matches the incoming label.
+
+        Raises:
+            ValueError: If thresholds are out of range or inverted.
+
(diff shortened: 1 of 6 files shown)


aaronsb:
## Code Review — ADR-102 portable backup/restore foundations

Reviewed all five commits against `origin/main`. This is well-structured foundations work: the spec is rigorous, the validator core/CLI split is clean (pure `validate_backup` + thin I/O boundary), and the intentional non-wiring of `concept_matcher.py` is correct for Phase 4. Findings below by severity. Nothing here is a merge blocker for *foundations* — the one blocker-class item is a correctness divergence to resolve before Phase 4 wires the matcher.

---

### restore-database.sh (#397 / #398)

**Both bug fixes are correct.** Verified the exit-code semantics carefully against `set -e` (active at line 9):

- **#397 pipefail** — `set -o pipefail` is scoped tightly around the `pg_restore | grep | grep` pipeline, and the pipeline sits in an `if` *condition*, which is exempt from `set -e`. So pipefail correctly surfaces pg_restore's exit status instead of the trailing grep's, and the `if/else` catches a real failure. `set +o pipefail` is restored on **both** branches (lines 234, 237). Correct.
- **#398 DROP/CREATE** — Sound. Terminating connections via `pg_terminate_backend` before `DROP DATABASE` is the right move (DROP fails if sessions remain), `pid <> pg_backend_pid()` correctly avoids self-termination, run against the `postgres` maintenance DB, `ON_ERROR_STOP=1` + `if !` wrapping gives real error handling, and the partial-drop hint in the failure path is a nice touch. The `--clean` removal genuinely fixes the AGE label-table abort.
- **Count gating** — The empty-graph guard correctly normalizes whitespace, defaults to 0, regex-validates numeric, and gates on `concepts==0 && sources==0`. Under `set -e` the `[ "$X" -eq 0 ]` tests are in a `||`/`&&` compound inside `if`, so a non-zero (false) test won't abort the script. Correct.

**should-fix (#397, minor):** The data-loss guard's premise — "a backup always contains concepts AND sources, so zero of both means failure" — is a reasonable heuristic but will false-positive on a legitimately empty/near-empty graph backup (e.g. a freshly-initialized instance, or a restore used in test fixtures). It exits 1 *after* the DROP/CREATE+restore has already run, so the operator is left with a correctly-restored-but-reported-failed empty DB. Consider gating on `concepts==0 && sources==0 && instances==0` plus an explicit note, or making the empty-graph case a loud WARNING rather than a hard `exit 1`. Not blocking, but worth a comment in the code about the known false-positive.

**nit:** The text-format (non-binary) restore branch (the `else` at line ~227 in the new file) did **not** get the count-gating or the DROP/CREATE treatment — only the binary pg_restore path did. That's defensible (the two known bugs are pg_restore-specific) but the success message is now only printed once at the bottom, gated by the binary-oriented guard. Confirm the plain-SQL `psql` restore path still reports correctly and isn't subject to the AGE concern. Worth one line in the commit/PR noting the text path was intentionally left as-is.

---

### concept_matcher.py — port faithfulness (blocker-class for Phase 4, not for this PR)

The engine itself is clean: threshold validation in `__init__`, empty-embedding short-circuit returning `None`, `vector_search` wrapped in try/except returning `None` on failure, best-first walk, correct two-tier accept logic. Boundary handling (`>=` on both tiers) matches `ingestion.py:452-460` exactly. `_labels_match` faithfully reproduces the ingestion normalize+containment rule and correctly guards empty/None.

**The docstring's faithfulness claim is contradictory — resolve before wiring (Phase 4):**

The module docstring says it is "Ported from the legacy `api/lib/restitching.py` `ConceptMatcher`" *and* that "the matching policy mirrors the canonical ingestion path." Those are **two different algorithms**, and the code implements the ingestion one, not the legacy one:

- **Legacy `restitching.py:130-156`** uses a **single** threshold (`best_score = self.threshold`, default 0.85), strict `similarity > best_score`, and **no label-boost tier at all**. It scans all concepts and returns the single global argmax above 0.85.
- **New `concept_matcher.py`** implements the **ingestion** two-tier policy (0.85 strict / 0.75 label-boosted) via `vector_search(threshold=0.75)`.

This is almost certainly the *right* choice (canonical ingestion behavior > legacy behavior), but the ADR §Context describes the legacy ConceptMatcher as "the engine for integration mode" with "0.85 strict / 0.75 label-boosted." That ADR line is actually inaccurate about the legacy code (legacy has no 0.75 tier), so the new code matches the ADR's *intent* but not the legacy *implementation*. **Recommendation:** Drop the "ported from restitching.py" framing or downgrade it to "supersedes," and state plainly that this is the ingestion-matcher policy reimplemented behind AGEClient. As written, a future reader diffing against `restitching.py` to verify the port will find a behavior change that the docstring obscures.

**should-fix — behavioral divergence from `ingestion.py` (best-first vs top-only):** `ingestion.py:445` only ever inspects `matches[0]` (the single top candidate) — if the top match is in [0.75, 0.85) and its label does NOT match, ingestion creates a new concept even if `matches[1]` would have label-matched. The new matcher **walks all candidates** and accepts the first that qualifies, so it can match `matches[1]` where ingestion would not. The docstring advertises "mirrors the canonical ingestion path," but this is a genuine policy difference. It's arguably an improvement, but it is *not* a mirror. Either match ingestion's top-only semantics or document this as a deliberate, justified divergence (with a forward note to reconcile ingestion too).

**nit:** `candidate["similarity"]` and `candidate["concept_id"]` use direct subscript while `candidate.get("label")` is defensive. `vector_search` (query.py:106-112) always populates all three keys, so this is safe today, but the asymmetry is odd — either trust the contract for all three or guard all three.

---

### lint_backup.py — spec conformance

Strong implementation. The check registry, JSON-path locators, severity model, and the pure-core/IO-boundary split all match the design notes in the header and the sibling lint-tool conventions. Negotiation logic (§7) is faithful: exact major → full; lower major / legacy flat → notice + subset; higher major / unknown family → ERROR+refuse. `_is_int_index` correctly excludes `bool`. Cascade resolution, range checks, dup detection, referential integrity, learned_id (§5.4.1), epoch-mode gating (faithful vs simple via `graph_epochs` presence), and §6 exclusions are all present and match the spec sections cited.

On the **concept→ontology join field** the author flagged: the validator's `concept.get("ontology") or concept.get("document")` assumption is **reasonable and I'd keep it, but it should be documented as an assumption the spec doesn't pin down.** Evidence: BACKUP_OBJECT_SPEC §5.1 (concepts) lists **no** ontology/document field on concepts at all, while §5.2 lists `document` ("Ontology/document name") on **sources**, and §3 declares `ontologies[].name`. So the join key the cascade needs (concept → ontology name) is genuinely unspecified in §5.1. The `ontology or document` fallback is the most plausible guess, but two failure modes exist:

1. **False E_NO_PROFILE_CASCADE** if a real backup names the concept's ontology under some *other* key (or doesn't carry it on concepts at all, relying on a join through instances→sources→document). In a uniform single-profile backup `header.default_embedding_profile` saves the cascade (step 3), so this only misfires on **mixed** backups where the backup default is absent/non-int and the per-concept ontology key differs from `ontology`/`document`. That's exactly the case the spec is silent on.
2. **Silent wrong-profile resolution** if a concept's `document` value collides with an `ontologies[].name` that is not actually its ontology — the cascade would resolve to the wrong profile index without erroring.

Recommendation: (a) add a code comment marking this as a spec-ambiguity workaround with a pointer to the open §5.1 question, and (b) tighten the spec — §5.1 should explicitly state the concept's ontology join field (or state that concepts inherit ontology via their sources). Right now spec and validator are silently making different-but-overlapping assumptions. **should-fix on the spec, nit on the validator** (the code's behavior is the best available given the spec gap).

**should-fix — `E_NO_PROFILE_CASCADE` is aggressive for legacy/partial backups:** It fires ERROR per-concept whenever the full cascade yields nothing. For a `kg-backup/2` that legitimately omits `default_embedding_profile` (e.g. an all-recompute backup where vectors will be regenerated on restore per §6/§6-cascade), every concept errors. Spec §4.1 frames the backup default as the cascade *top* but doesn't make `default_embedding_profile` strictly mandatory in all modes. Consider whether a missing-cascade-but-no-vectors case should be WARNING rather than ERROR.

**nit — `bulk.vocabulary` is unvalidated.** Spec §5.5 says bulk `vocabulary` rows mirror `header.relationship_vocabulary[]` and are the import payload. The validator checks neither presence nor consistency between the two. The self-test's valid sample carries `vocabulary: []` while `header.relationship_vocabulary` has one entry, and passes — a real mismatch class goes undetected. Low priority (foundations), but note it as a known gap so a future check code is reserved.

**nit — archive manifest filename.** `load_backup_object` looks for `manifest.json` in the `.tar.gz`, and the spec (§1) says the archive wraps "the JSON manifest." Good alignment. Confirm `backup_archive.py` actually names it `manifest.json` (couldn't verify the producer in this PR since the exporter is Phase 2). If the producer uses a different name, the `*manifest.json` suffix fallback covers most cases but not e.g. `backup.json`.

---

### Spec / ADR internal consistency (light pass)

- **§5.1 concept ontology field gap** — covered above; the most material doc gap.
- **ADR §Context vs reality** — the ADR says the legacy `restitching.py` ConceptMatcher does "0.85 strict / 0.75 label-boosted." The legacy code does **not** (single 0.85 threshold, no label boost). The *new* matcher does. The ADR describes the target design, not the legacy code it claims to describe. Minor, but it's the root of the concept_matcher docstring confusion above — worth a one-line correction in the ADR Context ("the legacy matcher is single-threshold; integration mode adopts the ingestion two-tier policy").
- **Validator header doc** references "P3/P6 pytest suites" and "a future live API endpoint" importing the core — consistent with ADR §5's "discoverable via API" note. Good traceability.
- Spec status is "Draft" and ADR-102 status is "Accepted" — fine for foundations; just flagging that the authoritative byte-spec is still Draft while the ADR referencing it is Accepted.

---

### SOLID / quality / traceability

Clean. Single-responsibility honored throughout (pure core vs IO; matching engine vs restitch surface deliberately dropped). Every artifact cites its ADR-102 section. `@verified` stamps present. No security concerns (validator is pure/offline; restore script's DROP is gated by the existing confirmation prompt and documented as irreversible). No tests added for `concept_matcher.py` — acceptable since it's unwired, but Phase 4 must add them, especially around the best-first-vs-top-only divergence above.

### Summary

Solid foundations. **One blocker-class item to settle before Phase 4 wiring:** the concept_matcher port is faithful to `ingestion.py`, not to legacy `restitching.py`, and diverges from ingestion in best-first-vs-top-only selection — pick a canonical semantics and document it. The spec's §5.1 concept→ontology join gap is the most worthwhile doc fix. Everything else is should-fix/nit and safe to land as foundations.

---
*AI-assisted review via Claude Code (reviewer role — no files modified).*

PR #506, opened 2026-06-12
feat: nomic-first default embeddings + offline model bake (appliance foundation)

aaronsb:
## What

Make **local nomic-768 the default embedding profile** and **bake the models into the image** so the platform boots fully offline — the foundation for an air-gapped / Raspberry-Pi appliance. Reasoning/extraction stays remote; embeddings run on-device.

## Why

Distribution is the biggest thing holding the system back from broader use. A nomic-first, offline-capable image means a fresh install needs a cloud key **only for extraction, never for embeddings**, and can run with no internet egress. Strategy recorded in ADR-103.

## Changes

**Nomic-first default** (`feat(embedding)`)
- Seed migrations 003/008 flip the active profile: OpenAI → inactive, Nomic v1.5 (768) → active
- mig-012 `validate_embedding` fallback `1536 → 768`; mock provider + two coupled tests aligned to 768
- operator-help + queries.py field description corrected

**Offline model bake** (`feat(appliance)`)
- `api/bake_embedding_models.py` + `api/docker-entrypoint.sh` — shared bake + first-boot cache seeder
- Invoked by **both** `api/Dockerfile` and `api/Dockerfile.rocm-host` (single source of truth; an inline bake in only the standard image previously skipped ROCm)
- Bakes text + vision nomic models into `/opt/hf-seed` (outside the cache volume) and seeds the mounted volume on first boot, avoiding the empty-volume-shadow trap

**ADR-103** (`docs(adr)`) — distribution strategy + nomic-first/reasoning-remote invariant.

## Validation (on real hardware, AMD ROCm)

- Fresh DB → active profile is `Nomic v1.5 / 768`
- Image rebuild → `/opt/hf-seed` carries all 3 nomic repos; entrypoint present
- Boot → `Loaded from local cache`, **zero** download lines
- Forced full re-embed of 4,375 entities (3994 concepts + 313 sources + 68 vocab) → 0 failures, **0 downloads**
- `pytest tests/test_mock_provider.py tests/test_vocabulary_manager_integration.py` → 32 passed

## Open (recorded in ADR, not blockers)

arm64 gates for the appliance path: AGE arm64 digest, Garage arm64, Pi 4GB RAM budget, QEMU arm64 bake.


aaronsb:
## Code Review — nomic-first defaults + offline model bake

Reviewed the migration chain, the bake/entrypoint mechanism, dimension coupling, and shell/quality. Validation on ROCm hardware covers the happy path well; my focus is the robustness gaps the live test wouldn't have exercised — chiefly the **standard `api`-user image** (not the root ROCm image that was tested live) and the **ROCm volume path**.

### Assessment: sound design, one real robustness gap on ROCm, a couple of latent couplings to document.

---

### Migration correctness — CORRECT

The 003 → 008 → 055 chain produces exactly one active profile, no partial-unique-index collision:

- **003**: OpenAI inserted `active=FALSE` → not covered by the partial index `(active) WHERE active=TRUE`, inserts cleanly.
- **008**: Nomic inserted `active=TRUE`. At that point no active row exists (OpenAI is inactive), so the partial index has no entry → succeeds as the sole active row.
- **055** (on main, not in this PR): both profile inserts copy `ec.active` from source rows, and Nomic (active) is inserted *before* OpenAI (inactive). The `embedding_profile(active) WHERE active=TRUE` index sees exactly one TRUE. Carry-forward is correct.

**Nice-to-have:** 003's `ON CONFLICT (active) WHERE active=TRUE DO NOTHING` → `ON CONFLICT DO NOTHING` weakens re-run idempotency, because there is no unique constraint on `(provider, model_name)` — a hand re-run of the seed would now insert a duplicate OpenAI row rather than being skipped. The migration-version guard in the runner prevents this in practice, so it's not a blocker, just a sharper edge than before.

---

### Dockerfile bake + entrypoint

**(a) Seed path outside the mount — CORRECT.** `/opt/hf-seed` is not under any mounted volume. The empty-volume-shadow trap is genuinely avoided.

**(b) HOME/cache resolution — CORRECT for standard, but a real gap on ROCm:**

- Standard image: `USER api`, `$HOME=/home/api`. Volume mounts at `/home/api/.cache` (parent of the HF cache `/home/api/.cache/huggingface`). Entrypoint `DEST=$HOME/.cache/huggingface` lands *inside* the volume. Runtime loader (no HF_HOME set) resolves the same path. **Match — seed persists.**
- **ROCm image (SHOULD-FIX):** no `USER` directive → runs as **root**, `$HOME=/root`. The ROCm override is a partial compose override and inherits the base `hf_cache:/home/api/.cache` mount. So at runtime root's HF cache is `/root/.cache/huggingface` while the persistent volume sits at `/home/api/.cache` — **root never touches the volume.** Entrypoint and runtime *agree* on `/root/.cache/huggingface`, which is why the live test passed and saw zero downloads (the seed is re-copied to the container's writable layer every boot, cheaply via `cp -Rn`). But the named volume is dead weight on ROCm, and the seeded cache does not survive `docker rm` / container recreation. Either mount `hf_cache:/root/.cache` in the ROCm override, or add `ENV HOME=/home/api` + a writable `/home/api/.cache` for root, so the cache actually persists.

**(c) `cp -Rn` — CORRECT.** `HF_HOME=/opt/hf-seed` at bake time yields `/opt/hf-seed/hub` + `/opt/hf-seed/modules`; `cp -Rn "$SEED/."` reproduces `hub/` + `modules/` under the live cache root, which is exactly what `local_files_only=True` expects (hub blobs + the trust_remote_code dynamic-module cache). Idempotent and non-clobbering.

**(d) No drift — CORRECT.** Both Dockerfiles invoke `HF_HOME=/opt/hf-seed python /app/api/bake_embedding_models.py` and copy the same entrypoint. Single source of truth achieved; the prior inline-bake-only-in-standard-image drift is resolved.

**(e) HF_HOME not leaked to runtime — CORRECT.** `HF_HOME` is inline on the bake `RUN` only, never an `ENV`. Runtime falls back to `$HOME/.cache/huggingface`, so the volume is honored. The entrypoint's `${HF_HOME:-...}` correctly resolves to the home-based path at runtime since HF_HOME is unset there.

---

### Dimension coupling (1536 → 768) — CORRECT, one legacy file to note

mock provider default, the two tests, and mig-012 `validate_embedding` default all move to 768 consistently. The live 4,375-entity re-embed at 768 with 0 failures confirms the end-to-end vector path handles 768.

**Nice-to-have (not in this PR):** `schema/init.cypher` still hardcodes `vector.dimensions: 1536` for the `concept-embeddings` and `ontology-embeddings` Neo4j `CREATE VECTOR INDEX` blocks. This appears to be dead Neo4j-era config (no references from operator/docker/scripts; the project runs on Apache AGE), so it's not load-bearing — but it's misleading for a nomic-first default and worth deleting or correcting in a follow-up. The remaining 1536 hits in `api/app/**` are docstrings/comments and OpenAI-specific paths — cosmetic.

---

### Shell safety & quality

- `set -e` with `cp ... 2>/dev/null || true` is correct — the `|| true` neutralizes cp failures (races/perms on first boot) so `set -e` won't abort. Variables are quoted throughout.
- `exec "$@"` correctly forwards the compose `command:` override (dev/prod/rocm all set one) or the Dockerfile CMD. ENTRYPOINT exec-form + CMD compose cleanly.
- Bake script docstrings are thorough and explain the load-not-download rationale and the mirror-the-seeds invariant.

**Latent coupling to document (nice-to-have):** the bake passes **no `revision`**, while the runtime loaders pass `model_revision` from the profile. Today migration 055 sets `text_revision`/`image_revision` to NULL, so both resolve to `main` and agree. But if a profile ever pins a revision, the bake would cache `main` while runtime requests the pinned commit → cache miss → offline-boot download attempt → failure on an air-gapped box. Worth a one-line note in the bake script (or pinning revisions in both places) so the appliance invariant doesn't silently break later.

---

### Summary

- **Blocker:** none.
- **Should-fix:** ROCm volume path — the persistent `hf_cache` volume is never used by the root-running ROCm container (cache lands on the ephemeral container layer). Works today, but defeats cross-recreation persistence on the one variant that was tested live.
- **Nice-to-have:** 003 re-run idempotency edge; `init.cypher` legacy 1536; bake-vs-runtime revision coupling (document or pin).

Standard `api`-user path (the untested one) is correct by construction — that was my main worry and it holds.

---

Issue #3, opened 2025-10-05
Build MCP Server with Semantic Tool Hints (ADR-003)


aaronsb:
Create MCP server with text adventure style tool guidance.

**Core MCP Tools:**
- `search_concepts` - Semantic search with vector similarity
- `create_concept` - Create new concept (requires search first)
- `create_relationship` - Link concepts
- `get_concept_details` - Full concept info with evidence
- `merge_concepts` - Librarian operation
- `flag_for_review` - Quality control
- `adjust_bias` - Curator operation

**Tool Hint Features:**
- [ ] Define tool prerequisites (e.g., search before create)
- [ ] Implement error hints with helpful messages
- [ ] Add suggested next actions
- [ ] Role-based tool visibility
- [ ] Audit logging for sensitive operations

**Reference:** docs/ARCHITECTURE_DECISIONS.md ADR-003

aaronsb:
Closing as superseded by current architecture:
- MCP server fully implemented with comprehensive read/write operations
- 20+ MCP tools including search, traversal, ingestion, and admin functions
- System uses Apache AGE (PostgreSQL), not Neo4j
- Current implementation in client/src/mcp-server/
- Documentation in docs/guides/MCP_SERVER.md

Issue #505, opened 2026-06-12
Restore doesn't rehydrate derived graph layers (ontology + document); clean install doesn't seed primordial


aaronsb:
## Summary

Two related defects found while restoring a backup into a fresh instance. The graph data restores correctly, but **derived graph-projection layers are not rebuilt**, so ontologies/documents appear empty in the CLI and catalog.

## Bug 1 — clean install doesn't seed the `primordial` ontology

The reserved `primordial` pool is created **lazily**, only when annealing first runs (`proposal_executor._ensure_primordial_pool`, called at `proposal_executor.py:375`). A fresh instance has no `primordial` `:Ontology` node until an annealing cycle fires. It should be seeded **eagerly at clean install**.

## Bug 2/3 — restore doesn't rehydrate derived graph-projection layers

Restoring a backup that predates a projection-layer serialization leaves that layer unbuilt. The restored `:Source` nodes retain all source-of-truth data (`document`, `garage_key`, `content_hash`, `source_id`), but restore creates **neither**:

- **`:Ontology` + `:SCOPED_BY`** (ontology layer) → `kg ontology list` shows "No ontologies found"
- **`:DocumentMeta` + `:HAS_SOURCE`** (document layer) → `kg catalog ls` shows `(0)` children; also the source of the `relation "knowledge_graph.DocumentMeta" does not exist` warning during restore.

Concept/source/vocab embeddings restore correctly (vectors carried as-is); only the derived node/edge projection layers are missing.

## Root cause

Restore has **no rehydration step for derived graph layers** (unlike the source-embedding rehydration it already enqueues). Backups predating ADR-200 ontology nodes (migration 044) / the DocumentMeta layer can't carry those nodes, and restore never reconstructs them from the Sources that remain the source of truth.

## Proven reconstruction (validated live against a restored graph)

- **Ontology layer** — re-ran migration 044 logic (distinct `s.document` → `:Ontology` + `:SCOPED_BY`): created **18 ontologies, 313 edges** → `kg ontology list` correct (concept counts populated).
- **Document layer** — grouped Sources by `garage_key` → one `:DocumentMeta` per file + `:HAS_SOURCE` to its chunks: created **120 DocumentMeta, 313 edges** → `kg catalog ls` document tier correct.

## Implementation notes / gotchas

[...]

PR #507, opened 2026-06-12
fix: restore the :Ontology + :DocumentMeta tiers + seed primordial (#505)

aaronsb:
Toward #505 — closes after the maintainer's end-to-end verification (wipe → rebuild → restore).

Restoring a backup left the catalog's **document tier** empty (`kg catalog ls`, document drill-down) and — for pre-serialization backups — the **ontology tier** too (`kg ontology list`). This PR restores both authoritatively, and seeds the reserved `primordial` pool on clean install.

## What's in here (4 commits)

1. **`feat(restore)` + `refactor(restore)` — ontology-layer rehydration.** `RehydrationMixin.rehydrate_projection_layers()` rebuilds `:Ontology` + `:SCOPED_BY` from distinct `s.document`. For new backups the importer already MERGEs these (serialization), so it's an idempotent no-op; for **old** backups it's the reconstruction. (The original document-half of this rehydration was removed — it produced a truncated 32-char `document_id` that broke drill-down + dedup; replaced by serialization below.)

2. **`feat(install)` — primordial seed.** Migration `078` seeds the `primordial` pool eagerly + idempotently (was only created lazily during annealing).

3. **`feat(serialization)` — serialize the `:DocumentMeta` layer + restore-time durability (Option B).**
   - **Exporter** emits `:DocumentMeta` nodes + `HAS_SOURCE` edges, carrying `document_id`/`content_hash` **verbatim** (canonical `sha256:`-prefixed digest) and `garage_key` (no prefix) — no recomputation.
   - **Importer** `_import_documents` MERGEs on the canonical `document_id` → restoring the same backup twice never duplicates a `:DocumentMeta` (**robustness req #1**).
   - **Restore-time durability**: pre-B backups (no `documents` stream) reconstruct the tier from the document bytes in Garage, hashing them to recover the **canonical full digest** (not the truncated `garage_key` prefix). Garage I/O injected; mixin stays DB-only; guarded on an empty tier (no-op for new backups).
   - **Determinism (req #2):** hashes/keys are carried, never recomputed at restore; the durability path hashes raw bytes and asserts the result matches the existing `garage_key` prefix.
   - **Validator**: `documents`/`has_source` referential + duplicate checks, mirroring the ontology streams.

## Why serialize instead of reconstruct

The canonical identity `document_id == content_hash == sha256(document bytes)` is **not recoverable** from `:Source` nodes — `garage_key` carries only `content_hash[:32]`, and `Source.content_hash` is the per-chunk hash. A Source-derived id is truncated and breaks `GET /documents/{id}` (reads `d.garage_key`) and re-ingest dedup. Confirmed against the live graph; full analysis in `.claude/todo-issue-505-restore-fidelity.md`.

## Validation

- **Non-destructive probe on the live graph:** 120/120 Garage objects hash back to their `garage_key` prefix (0 mismatch) — the durability path recovers canonical hashes exactly.
- **Tests (77 passing in the backup/restore set):** build→read round-trip preserving the full id; importer idempotency (double restore → one node); pre-B backup imports no tier; durability reconstructs the canonical hash + is idempotent; validator orphan/dup cases.
- **End-to-end** (wipe → rebuild → fresh restore) is the maintainer's acceptance test, run after review.

## Notes

- Data-quality observation (some `garage_key`s tagged to multiple `s.document` ontologies) tracked as a separate follow-up.
- A restore endpoint that imports *pre-extracted graph interchange* (vs. raw content) is a possible future feature — deliberately **not** in scope here; restore stays faithful rehydration, ingestion stays extraction.

### DIFF
--- a/api/app/lib/age_client/__init__.py
+++ b/api/app/lib/age_client/__init__.py
@@ -21,11 +21,13 @@ from .grounding import GroundingMixin
 from .ontology import OntologyMixin
 from .ontology_scoring import OntologyScoringMixin
 from .ontology_edges import OntologyEdgesMixin
+from .rehydration import RehydrationMixin
 from .vocabulary import VocabularyMixin
 
 
 class AGEClient(
     VocabularyMixin,
+    RehydrationMixin,
     OntologyEdgesMixin,
     OntologyScoringMixin,
     OntologyMixin,
@@ -44,6 +46,7 @@ class AGEClient(
     - OntologyMixin: Ontology node CRUD and lifecycle management
     - OntologyScoringMixin: Ontology scoring, analytics, and annealing mutations
     - OntologyEdgesMixin: Inter-ontology edges and proposal execution primitives
+    - RehydrationMixin: Rebuild derived :Ontology/:DocumentMeta layers from Sources (issue #505)
     - VocabularyMixin: Relationship type config, CRUD, sync, merge, embeddings
 
     Facades attached lazily via properties on QueryMixin:
--- a/api/app/lib/age_client/rehydration.py
+++ b/api/app/lib/age_client/rehydration.py
@@ -6,25 +6,31 @@ created before the :Ontology layer was serialized (pre-v0.15.1) does not carry
 that layer, and restore otherwise never rebuilds it — so the ontology list and
 the catalog read empty even though the underlying data is intact (issue #505).
 
-This mixin reconstructs the **ontology layer** idempotently from the Sources:
-``(:Ontology)`` + ``(:Source)-[:SCOPED_BY]->(:Ontology)``, keyed by distinct
-``s.document`` (mirrors migration 044). For new backups the importer already
-MERGEs these nodes/edges (serialization), so the pass is a harmless idempotent
-no-op; for old backups it is the reconstruction fallback. It also ensures the
-reserved ``primordial`` pool exists, since a clone-style restore can replace a
-freshly-seeded graph (see migration 078 / issue #505).
+This mixin provides two restore-time rebuilds:
 
-The :DocumentMeta layer is intentionally NOT reconstructed here: its canonical
-identity (``document_id == content_hash == sha256(document bytes)``) is not
-recoverable from Source nodes (the garage_key carries only a 32-char hash
-prefix), so a Source-derived DocumentMeta would have a truncated id that breaks
-drill-down and re-ingest dedup. That layer is handled authoritatively by
-serializing it into ``kg-backup/2`` (Option B, issue #505), with a faithful
-Garage-hashed reconstruction reserved for old backups.
+- :meth:`rehydrate_projection_layers` — the **ontology layer**:
+  ``(:Ontology)`` + ``(:Source)-[:SCOPED_BY]->(:Ontology)``, keyed by distinct
+  ``s.document`` (mirrors migration 044). For new backups the importer already
+  MERGEs these nodes/edges (serialization), so the pass is a harmless idempotent
+  no-op; for old backups it is the reconstruction. It also ensures the reserved
+  ``primordial`` pool exists, since a clone-style restore can replace a
+  freshly-seeded graph (see migration 078 / issue #505).
+
+- :meth:`rehydrate_document_layer` — **restore-time durability** for the
+  :DocumentMeta tier. New backups serialize :DocumentMeta into ``kg-backup/2``
+  (Option B) and the importer rebuilds it authoritatively. Old backups carry no
+  such stream, so this reconstructs it from the restored Sources — but
+  faithfully: the canonical identity (``document_id == content_hash ==
+  sha256(document bytes)``) is unrecoverable from Source nodes alone (the
+  ``garage_key`` carries only a 32-char hash prefix), so this hashes the actual
+  document bytes (fetched from Garage) to recover the full digest rather than
+  using the truncated prefix. Garage I/O is injected so the mixin stays DB-only.
 """
 
+import hashlib
+import os
 import logging
-from typing import Dict, Optional
+from typing import Callable, Dict, Optional
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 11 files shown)


aaronsb:
## Code Review — rehydration + primordial seed (#505)

Reviewed for correctness/robustness gaps the happy-path live run (which validated counts on `kg ontology list` / `kg catalog ls`) wouldn't surface. The epoch ordering and idempotency are sound; the gaps are all in the `DocumentMeta` property shape and the `document_id` derivation.

### Blockers

**1. Rehydrated `DocumentMeta` is missing `garage_key` → document-content drill-down silently returns empty** (`rehydration.py:108-120`)

`GET /documents/{doc_id}` (`documents.py:656-716`) reads `d.garage_key` to fetch the original document bytes from Garage:
```cypher
MATCH (d:DocumentMeta {document_id: $doc_id})
RETURN d.document_id, d.content_type, d.garage_key, ...
```
```python
garage_key = result.get('garage_key')   # None for rehydrated docs
...
if garage_key:                           # skipped → no 'document' in content
    doc_bytes = source_storage.get(garage_key)
```
The rehydration `SET`s only `filename, ontology, content_type, source_count` — never `garage_key`. So for every rehydrated document the content fetch is skipped and the drill-down returns chunks-only with no original document body. The `garage_key` is in hand (it's the grouping key for the very loop that creates the node) — set it: `SET d.garage_key = $garage_key`. The real ingestion path (`create_document_meta`) sets it too.

**2. Rehydrated `DocumentMeta` is missing `content_hash` → graph dedup breaks; re-ingesting a restored file creates a duplicate `DocumentMeta`** (`rehydration.py` vs `ingestion.py:704`)

Dedup keys on `{content_hash, ontology}`, not `document_id`:
```cypher
MATCH (d:DocumentMeta {content_hash: $hash, ontology: $ontology}) RETURN d
```
Rehydration never sets `content_hash`, so `get_document_meta()` returns None for every rehydrated doc → the "already ingested" short-circuit is bypassed and re-ingesting a file that was restored creates a second `DocumentMeta` for the same content. The `:Source` nodes carry `content_hash` (per the module docstring), so it's available — pull it into the projection query and `SET d.content_hash`. This also lets you derive a correct `document_id` (see #3).

### Should-fix

**3. `document_id` derivation is wrong for the dedup scheme, and the regex only handles `.md`** (`rehydration.py:104-106`)

```python
file_hash = re.sub(r"^.*/([^/]+)\.md$", r"\1", garage_key)
document_id = f"sha256:{file_hash}"
```
Two problems:
- **Scheme mismatch.** The garage_key basename is `content_hash[:32]` (`source_storage.py:127,151` — `HASH_PREFIX_LENGTH = 32`), the *truncated* hash. The real ingestion `document_id` is `sha256:<full 64-char hexdigest>` (`ingestion_worker.py:624,632`, `content_hasher.hash_content`). So this produces `sha256:<32 chars>`, which never equals the canonical id. Drill-down is internally self-consistent (list endpoints return this id and detail matches on it), so counts pass — but it diverges from the real scheme and won't line up with a re-ingested doc. If you `SET d.content_hash` from the Source (#2), use that as the document_id instead and the rehydrated id matches what ingestion would produce.
- **Regex is `.md`-only.** For any non-`.md` key (`.txt` is the storage default, plus `.pdf` etc.), the pattern doesn't match and `re.sub` returns the **whole string unchanged**, yielding `document_id = "sha256:sources/My_Docs/<hash>.txt"` (full path embedded). Still unique, so counts don't catch it, but it's clearly unintended. The live run presumably had all-`.md` fixtures.

**4. Migration 078 comment cites the wrong migration for the `:Ontology` label** (`078_seed_primordial_ontology.sql:25`)

The comment says "The :Ontology label is precreated by migration 058". The label is created by **migration 044** (`044_ontology_graph_nodes.sql` — `create_vlabel` + the distinct-`s.document` projection this rehydration mirrors). There is no `058` migration touching `:Ontology`. Ordering still holds (78 > 44), but fix the citation so the safety rationale points at the real dependency.

### Nice-to-have

**5. Best-effort try/except on rehydration — agree, but make a partial failure visible.** Swallowing to non-fatal is right (a completed restore shouldn't fail on a projection rebuild). But the layer rebuild is multi-statement and non-transactional: a mid-loop failure leaves a *partial* projection that still logs WARN and then completes the epoch — the catalog will look populated-but-incomplete with no easy signal. Consider logging the partial stats dict in the except branch (how far it got), or returning a `complete: bool` the worker can surface, so an operator can tell a clean no-op from a truncated rebuild.

**6. Provenance fields divergence (minor).** Rehydrated ontologies record `created_by="post_restore_rehydration"` and `creation_epoch` from the live metrics counter; migration 078 records `created_by='system'`/`creation_epoch:0`; lazy `_ensure_primordial_pool` records `annealing_worker` and a 12-char id (`ont_<uuid.hex[:12]>` vs `ont_<full-uuid>`). All three existence-guard on `name`, so no duplicate is created and the id-format difference is harmless — but the recorded provenance is first-writer-wins and therefore nondeterministic. Fine to leave; noting for the #505 data-quality follow-up.

**7. Multi-ontology-per-file handling is the right call.** One `DocumentMeta` per `garage_key` with parent resolved via `:SCOPED_BY` matches how the catalog already tolerates the 153-combos-vs-120-files skew, and the `seen` dedup is correct (first row per gk wins; the dropped rows only differ in `s.document`, which you don't key on). The one wrinkle: `d.ontology` is set to whichever ontology won the `seen` race for that gk — arbitrary among the file's ontologies. Since parent is resolved via `:SCOPED_BY` this is cosmetic, but worth a one-line note in the docstring that `d.ontology` on a multi-ontology file is non-authoritative.

### Assessment

Epoch ordering is correct and well-reasoned: rehydration runs inside the import window (derivations read stale), then `complete_epoch`/`_resolve_replayed_epochs` advances the committed watermark past the restored stamps and `refresh_epoch()` ticks freshness — the catalog rebuild is gated on the global watermark (whole-index re-projection, `catalog_facade._index_epoch`), not per-node `creation_epoch`, so the missing DocumentMeta epoch stamp doesn't block visibility. "No separate watermark bump needed" checks out. `_parse_agtype` scalar/None handling is correct and the `... or 0` guards are right. MERGE idempotency holds.

The structural data is sound — what's missing is the `DocumentMeta` *property completeness* (`garage_key`, `content_hash`), which is exactly what a count-based live check can't see. Blockers #1/#2 both resolve by carrying two more Source properties into the projection; fixing #2 also fixes #3's scheme mismatch. Recommend the follow-up integration test assert on a drill-down (content fetch + re-ingest dedup), not just layer counts.

---
*AI-assisted review via Claude Code (reviewer role — analysis only, no edits made)*

aaronsb:
Converting to **draft**. Review (above) found 2 blockers in the `:DocumentMeta` reconstruction half: the canonical `document_id == content_hash == sha256(document bytes)` isn't recoverable from restored Sources (only the 32-char prefix lives in `garage_key`), so a Source-derived reconstruction can't be authoritative — it breaks document drill-down and re-ingest dedup.

Decision: pursue **Option B** — serialize the `:DocumentMeta` layer into `kg-backup/2` so restore recreates it authoritatively, with the reconstruction kept only as a faithful fallback for pre-B backups. Design + task breakdown captured in `.claude/todo-issue-505-restore-fidelity.md`.

This PR will be reworked to merge only the unambiguously-correct parts (ontology-layer rehydration + primordial seed / migration 078); the document layer moves to the B work.

aaronsb (commented):
## Code Review — commit `e7df250ab` (serialize :DocumentMeta + restore-time durability)

Reviewed the focal commit against the two robustness requirements (idempotency + determinism), with the earlier 3 commits read for context. Verdict: **no blockers.** The two crux requirements hold. Findings below, grouped by severity.

### Verified sound (the crux)

**Requirement #1 (idempotency) — holds across all modes.**
- `_import_documents` MERGEs on `document_id` alone (`importer.py:533`), and HAS_SOURCE MERGEs the edge (`importer.py:540-543`). Both are idempotent; the round-trip test (`test_document_layer_round_trips_and_is_idempotent`) restores twice and asserts exactly one node + one edge.
- Source ids are **not** remapped in any mode (preserved 1:1 — `importer.py:80`, `:397`), so HAS_SOURCE `source_id` endpoints stay valid in clone / adjacent / integration. The maintainer's reasoning is sound: `document_id` is `sha256:` of bytes — embedding- and ontology-independent — so integration-mode embedding remap cannot perturb it. Confirmed.

**Requirement #2 (determinism) — holds, including the durability recompute path.**
- Serialized path carries `document_id` / `content_hash` / `garage_key` verbatim via `_parse_nullable_str` (no recompute). Good.
- The durability path's canonical form lines up with native ingestion **end to end**, which I traced specifically because it's the one recompute path:
  - Native dedup keys on `{content_hash, ontology}` (`get_document_meta`, `ingestion.py:705`), with `content_hash` passed in **prefixed** form (`content_hasher.py:42` returns `"sha256:"+hex`; `ingestion_worker.py:633` stores it prefixed; `create_document_meta` MERGEs on `document_id` = same prefixed string).
  - `rehydrate_document_layer` sets `content_hash = document_id = "sha256:"+full_hash` (`rehydration.py:166-168`) and `ontology = s.document` (`rehydration.py` — `Source.document` *is* the ontology grouping name per `ingestion.py:217`). So a re-ingest after a durability rebuild hits the dedup MATCH exactly. No drift.
  - `garage_key` is carried, never rebuilt, so `normalize_content_hash` is never re-invoked on the carried path — no prefix/no-prefix drift risk there.

**Epoch ordering (Q4) — correct.** Serialized doc import runs inside Stage-3 `import_backup` (before completion); the durability rebuild runs after ontology rehydration + Stage-4 Garage upload and **before** `complete_epoch` / `refresh_epoch` (`restore_worker.py:347-375` precede the epoch-resolution block at `:382+`). The first post-restore freshness tick therefore sees both. Confirmed.

**`head(collect(...))` grouping (Q3) — correct.** `MATCH (s:Source) ... RETURN s.garage_key AS gk, head(collect(s.document)) AS doc, count(s) AS cnt` groups implicitly by the lone non-aggregated key `s.garage_key` (AGE/openCypher implicit grouping). One DocumentMeta per garage_key, as intended.

### Should-fix

**1. Scoped (per-ontology) export drops the doc tier for multi-ontology documents — and the global durability guard won't recover it.**
`export_documents` / `export_has_source` filter on `DocumentMeta {ontology: $ontology}` (`exporter.py:478`, `:546`). A document whose bytes were ingested into ontology A and B exists as a **single** DocumentMeta node carrying only one ontology value (native `create_document_meta` MERGEs on `document_id` alone — `ingestion.py:838` — last-writer-wins on `ontology`/`garage_key`). A scoped export of the *other* ontology emits that ontology's Sources but **not** the doc node or its HAS_SOURCE edges. On restore, the durability fallback is gated on a **global** `count(DocumentMeta) == 0` (`restore_worker.py:351`), so if any other DocumentMeta survived, the orphaned-Source ontology never gets its tier rebuilt — silent partial gap.
This is downstream of the pre-existing multi-ontology data-quality issue (already tracked as task #9 / "separate issue"), and the PR does not regress full-database restore (the dominant path). Recommend: note this limitation in the `export_documents` docstring and the durability-guard comment so the gap is discoverable, and let task #9 own the real fix. Not a blocker for full-DB restore.

**2. Durability guard `count(DocumentMeta) == 0` is global, so a partial import is not self-healing.**
If a serialized restore writes *some* but not all DocumentMeta (e.g. an interrupted import, or the scoped case above), the guard sees a non-zero count and skips durability entirely (`restore_worker.py:351`). The rebuild method is itself idempotent (MERGE on `document_id`) and *could* safely run unconditionally to fill gaps, but it's gated off. The guard is the right discriminator for the clean new-backup vs. clean old-backup split the PR targets; it just doesn't cover partial states. Worth a one-line comment acknowledging the guard is all-or-nothing, or consider running durability whenever any Source lacks a HAS_SOURCE parent rather than on the zero-count.

### Nits

**3. `export_has_source` strips quotes inconsistently with `export_documents`.** `export_has_source` uses `str(r.get(...)).strip('\"')` (`exporter.py:556-558`) while `export_documents` (and the rest of the module) uses the `_parse_nullable_str` helper. The raw `.strip('\"')` won't normalize the literal agtype `"null"` to `None` — harmless here since both endpoints are non-null keys, but it diverges from the module's established convention. Prefer `_parse_nullable_str` for consistency.

**4. `created_by` parameter on `rehydrate_document_layer` is accepted but unused** (`rehydration.py`), and the call site passes `created_by=\"post_restore_rehydration\"` (`restore_worker.py:358`). Docstring already flags it as signature-symmetry-only, so this is intentional — just noting the dead param in case a future reader expects provenance stamping on the rebuilt nodes (the durability path does not set `ingested_by`, unlike the serialized path).

**5. Integrity-mismatch handling is log-and-proceed.** When `full_hash` doesn't start with the garage_key prefix (`rehydration.py:159`), the code warns but still writes the node from the recomputed hash. That's a defensible choice (recomputed hash is authoritative), and the docstring explains it — but a prefix mismatch means the stored bytes don't match the key, which is corruption-grade. Consider counting these into the returned stats (alongside `skipped`) so a restore surfaces \"N integrity mismatches\" rather than burying them in warnings.

### Assessment
Solid, well-tested change. The serialization-as-source-of-truth choice (Option B) is the right call given the truncated-prefix constraint, and the durability fallback is faithful — I specifically traced the recompute path against native dedup and it converges exactly. Test coverage is good: round-trip + verbatim-id, double-restore idempotency, pre-B no-tier, durability canonical-hash reconstruction + idempotency, and validator orphan/dup cases. The should-fix items are edge cases of an already-tracked multi-ontology data-quality issue, not regressions in the primary full-database restore path.

---
*AI-assisted review via Claude (code-reviewer). Scope: focal commit e7df250ab.*

Issue #502, opened 2026-06-09
Infra config asserts: prod port exposure, nginx security headers/HSTS, weak .env.example secrets + startup refusal


aaronsb:
**ADR-401 item 6** · Audit: `docs/security/security-consistency-audit-2026-06-09.md` (F5)

Verified findings:
- `docker/docker-compose.yml:38,92-93` publishes Postgres 5432 and Garage 3900/3903 to the host (dev); the Garage admin API has no auth. Prod compose publishes nothing — correct, but the invariant is unguarded.
- `docker/nginx.prod.conf:30` — HSTS commented out. Neither `nginx.prod.conf` nor `web/nginx.conf` sets `X-Frame-Options`, `X-Content-Type-Options`, or CSP.
- `.env.example:84,100,205` — `POSTGRES_PASSWORD=password` plus two `CHANGE_THIS_TO_A_RANDOM_SECRET_KEY` placeholders. Nothing refuses to boot on placeholder secrets.

**Acceptance criteria**
- [ ] Uncomment HSTS; add `X-Frame-Options`, `X-Content-Type-Options`, and a CSP to prod nginx (dev nginx: headers minus HSTS)
- [ ] CI assert script fails if: prod compose publishes 5432/3900/3903; HSTS/security headers absent or commented in `nginx.prod.conf`; `.env.example` contains a real-looking secret value
- [ ] API startup refuses to run outside `DEVELOPMENT_MODE` when `POSTGRES_PASSWORD == "password"` or any secret matches `CHANGE_THIS`
- [ ] Dev/prod port-exposure delta documented (one paragraph in the operator way or ops docs)

PR #517, opened 2026-06-15
feat(appliance): in-VM Traefik router — unified HTTP ingress (ADR-105 step 1)

aaronsb:
## What

Introduces **ADR-105** (scenario-driven TLS via an in-VM Traefik router) and implements its **step 1**: an opt-in Traefik (MIT) router giving the appliance a single front door over **plain HTTP** — no certs yet.

```
http://<host>/        -> web  (nginx SPA)
http://<host>/api/... -> api  (:8000, /api prefix stripped)
```

This is the consistent endpoint the CLI / FUSE / MCP need, and the foundation the cert path builds on. **Per the agreed PR split, certs (Traefik certResolvers, EXTERNAL_URL, recert) land in a separate follow-up PR.**

## Why

`operator.sh init` / the appliance path couldn't stand up TLS or a unified ingress; cert logic was stranded in `install.sh`. Rather than hand-build a cert factory, ADR-105 adopts Traefik (ACME + lego DNS providers + routing for free). Step 1 proves the routing before any TLS touches it.

## Changes

- **`docker/docker-compose.traefik.yml`** — traefik service + web/api routing labels; pinned `traefik:v3.3.1`; attaches to `knowledge-graph-network`; `VITE_API_URL=/api` (same-origin).
- **`ROUTER_MODE` gate** (default `none` = **zero behavior change**) wired into **all three** compose-assembly sites (`common.sh` `get_compose_cmd`, `operator.sh`, `start-platform.sh`) — mirrors the ssl-overlay pattern; deliberately avoids the #502 single-site-drift trap.
- **`operator init --headless --router=traefik|none`** — flag, validation, persisted to `.operator.conf`.
- **appliance-integration CI** — provisions with `--router=traefik` and asserts the unified ingress (`/` → web, `/api/health` → api) alongside the existing direct-port + #502 checks.

## Verification

- Local smoke: a throwaway Traefik against the running stack routed `/` → web (200, SPA html) and `/api/health` → api (200, healthy JSON, prefix stripped).
- `docker compose config` dry-run confirms the overlay merges cleanly (traefik on :80, correct routers/strip).
- `bash -n` clean on all four touched scripts; CI will exercise the real GHCR/prod topology.

## ADR

`docs/architecture/infrastructure/ADR-105-scenario-driven-tls-via-in-vm-traefik-router.md` — includes the five-scenario spine (dev/private/internet/proxied/public-nat), per-mode lifecycle, offload `EXTERNAL_URL` contract, trust posture, licensing posture, and topology/scenario/offload diagrams.


aaronsb:
## Code Review — ADR-105 step 1 (in-VM Traefik HTTP router)

**Verdict: needs-work.** The routing design is sound and the gate wiring is *almost* right, but there is one BLOCKER that makes the feature non-functional on its actual target (the appliance / GHCR path) — and the PR's own CI already caught it (run 27519793824 failed in 34s).

---

### BLOCKER 1 — Traefik overlay references `kg-network`, which is undefined on the appliance/prod compose assembly

**Symptom (from this branch's CI):**
```
service "traefik" refers to undefined network kg-network: invalid compose project
```
Failure occurs in `init` Step 4 → `start-infra.sh` → `run_compose` → `common.sh get_compose_cmd`.

**Root cause:** there are two `get_compose_cmd` implementations with *different base-file logic*:

- `operator.sh:91` hardcodes the base as `docker-compose.yml` (always present → `kg-network` defined).
- `operator/lib/common.sh:123` uses `local base_file=\"\${COMPOSE_FILE:-docker-compose.yml}\"`.

`headless-init.sh:385` sets `COMPOSE_FILE=docker-compose.prod.yml` for the appliance/GHCR path and persists it to `.operator.conf` (`:475`). So on the appliance path, `common.sh get_compose_cmd` uses **prod as the base file and never includes `docker-compose.yml`**. The `kg-network` (name `knowledge-graph-network`) network is defined **only** in `docker-compose.yml:291-293` — `docker-compose.prod.yml` defines only `kg-internal` + `lannet`. The traefik service (`docker-compose.traefik.yml:28-29`) and the `--providers.docker.network=knowledge-graph-network` pin then reference a network that doesn't exist in the merged project.

Reproduced locally:
```
# dev path (base = docker-compose.yml):   OK, valid
# appliance path (base = prod, no base):  service "traefik" refers to undefined network kg-network
```

Net effect: the overlay works in **dev** (the one place ADR-105 says Traefik is *absent*) and is broken in **every appliance scenario** (the entire point). The `--providers.docker.network` pin you added is correct *in intent* — the bug is that `kg-network` isn't in scope on the prod assembly.

Options (pick per your topology preference, do not implement from this review):
- Have the traefik overlay route over `kg-internal` (the network web/api actually share in prod) instead of `kg-network`; or
- Give the traefik overlay its own top-level `networks:` declaration so the alias is defined regardless of which base file is in play, and attach web/api to it; or
- Make `common.sh get_compose_cmd` always include `docker-compose.yml` as the true base and treat prod as an overlay (matches `operator.sh` and `start-platform.sh`) — but that's a broader change with its own blast radius.

Whatever the fix: the CI you added is the correct gate and is doing its job. It must go green on the appliance path, not just dev.

---

### SHOULD-FIX 2 — `nginx.prod.conf` on the web container fights the HTTP-only premise

On the appliance/prod path the web service still bind-mounts `docker/nginx.prod.conf` (`docker-compose.prod.yml:105`), which is cube-specific: `listen 80; server_name kg.broccoli.house; return 301 https://...` plus a `:443 ssl` block pointing at `/etc/nginx/certs/kg.broccoli.house.*` (a host path that won't exist on a generic appliance/CI runner). Once BLOCKER 1 is fixed and Traefik forwards `/` to web:80, nginx will answer with a 301→https redirect (and may fail to load the :443 server for missing certs). That defeats "HTTP-only step 1" and will make `curl -fsS http://localhost/` behave unexpectedly. ADR-105 §2 explicitly says the web image should "[go] back to just serving static assets" behind the router — that re-plumb (or at minimum a router-mode-appropriate nginx conf) is load-bearing for step 1 actually working end-to-end, not just compose-validating. Please confirm what the web container serves on :80 under `ROUTER_MODE=traefik` before calling this done.

---

### SHOULD-FIX 3 — CI race: `start` may report success before Traefik is reachable; relies on prod path that's currently broken

The verification step (`appliance-integration.yml:103-122`) is well-constructed (retry loop, strips `/api`, dumps traefik logs on failure). Two notes:
- The "Wait for platform health" step (`:84-101`) curls `http://localhost:3000/` directly, then the ingress step curls `http://localhost/`. Because health passes on the direct port, a totally-missing Traefik would only be caught by the second loop — which is fine, but the 120s window assumes Traefik is up shortly after `start`. There's no explicit wait for the traefik *container* to be running before the curl loop; the loop tolerates it, so this is minor. Just confirm 120s is enough headroom on a cold runner.
- `docker logs \"$(docker ps -aqf name=traefik)\"` (`:121`) — if multiple containers match `name=traefik` this passes multiple IDs to `docker logs` and errors; container name is `\${CONTAINER_PREFIX:-knowledge-graph}-traefik` so in CI it's `kg-traefik` and the filter is fine, but consider `... | head -1` for robustness.

---

### Gate-wiring consistency (the three sites) — mostly good, one note

The `ROUTER_MODE=traefik` gate is present at all three assembly sites and all three default-safe (empty/`none` ≠ `traefik`), so **the default path is preserved**:
- `operator.sh:104` — `load_config` (`:55`) does *not* default `ROUTER_MODE`; it's safe only because unset ≠ `traefik`. Works, but slightly inconsistent with common.sh which explicitly defaults+exports.
- `operator/lib/common.sh:152` + the explicit `ROUTER_MODE=\"\${ROUTER_MODE:-none}\"; export ROUTER_MODE` at `:35-36` — correct.
- `start-platform.sh:144` — sources `.operator.conf` (`:128`) so it sees a persisted value; unset → safe.

**NIT 4** — `--router` is only wired into `headless-init.sh`; `operator/lib/guided-init.sh` (interactive init) has zero router handling, so it can never persist `ROUTER_MODE`. Acceptable since the appliance uses headless, but worth a follow-up so the interactive path isn't a dead end.

**NIT 5** — `PathPrefix(\`/api\`)` + `stripprefix /api`: a request to exactly `/api` strips to empty string rather than `/`. `/api/health` → `/health` is correct (matches the API's route and your CI check). Just be aware bare `/api` may behave oddly; not blocking.

**NIT 6** — overlay comment (`docker-compose.traefik.yml:13`) and several PR/CI comments state web/api are pinned onto `knowledge-graph-network`; in the prod assembly the network they actually co-reside on is `kg-internal` (see BLOCKER 1). The comments describe the dev topology, not the appliance one — please reconcile the narrative with whichever network the fix lands on.

---

### What's good
- Image is pinned (`traefik:v3.3.1`), `exposedbydefault=false`, docker.sock mounted `:ro` — correct least-surprise defaults.
- Priority ordering is right: api `priority=10` (`/api`) beats the web `priority=1` catch-all (`/`). `VITE_API_URL: /api` same-origin override is the correct SPA wiring.
- Default `ROUTER_MODE=none` genuinely no-ops the overlay — existing non-traefik deployments are untouched.
- Adding the ingress assertion to CI (rather than trusting compose-validates) is exactly right, and it's what surfaced the blocker pre-merge.

Once BLOCKER 1 is fixed and the CI goes green on the GHCR/appliance path (and SHOULD-FIX 2 is resolved so :80 actually serves the SPA), this is a clean step-1 landing.

---

